Issue #136, opened 2013-10-12
Inacurate Hierarchy Results


malaverdiere:
`Hierarchy.getSubclassesOf` and `getSubclassesOfIncluding` for `java.lang.Object` are giving inaccurate results.

This issue is experienced when working on an unit test for a component that is dependent on the hierarchy.

Setup:

```
Options.v.set_prepend_classpath(true)
Scene.v.addBasicClass("java.io.IOException")
Scene.v.loadNecessaryClasses()
```

Test:

```
    val obj = Scene.v.getObjectType.getSootClass
    val found = Scene.v.hierarchy.getSubclassesOfIncluding(obj)
    val everything = Scene.v.classes
    assert(found == everything, "Descendants of Object is not the scope of all classes")
```

The test fails. `everything` has 128 classes, but `found` has 94 classes. The difference is a few interfaces, and a few exceptions that inherit `IOException`. Adding `IOException` to the basic classes has not changed anything.

PR #164, opened 2013-12-31
Improved Hierarchy

malaverdiere:
- Modernized implementation to use generics and for-each loop
- Loosened return types from List to Collection
- Added tests that use the old Hierarchy as an oracle for correctness and handles regressions
- Fix for #136

Note: I did not implement the unimplemented methods - is there really a need for it.


mbenz89:

linghuiluo:
This PR is too old. Please reopen it if it is needed.

PR #210, opened 2014-05-20
changerequest soot.util.HashChain

nignag:
the behavior is more strict and less "surprising".

tow examples:
follows(b,a) could raise a NoSuchElementException ; follows(a,b) does not.
iterator(a,b) works fine ; iterator(b,a) throws an NoSuchElementException

the behavior is changed in the following way:
- the iterator will always stop at the last element and will not throw an exception (simple example: a chain consists of "A","B","C","D" and you iterate over the end like iterator("C","A") the last element will be "D" ) (a big change, but i did'nt found a situation where this is a problem)
- if an element to insert is null a NullPointerException will be thrown ( see also http://docs.oracle.com/javase/7/docs/api/java/util/AbstractCollection.html#add(E) -> NullPointerException)
- if selecting an entry that is not part of the chain a NoSuchElementException will be thrown
- if you try to add an element that is already part of the chain a RuntimeException is thrown (maybe IllegalArgumentException would be better)

### DIFF
--- a/src/soot/util/HashChain.java
+++ b/src/soot/util/HashChain.java
@@ -41,7 +41,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
-import java.util.Objects;
 
 /**
  * Reference implementation of the Chain interface, using a HashMap as the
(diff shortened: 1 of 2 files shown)


mbenz89:

PR #338, opened 2015-01-21
Add constructor flag for <clinit> methods.

MarcMil:
Otherwise ART will provide a warning.

### DIFF
--- a/src/soot/SootMethod.java
+++ b/src/soot/SootMethod.java
@@ -487,7 +487,7 @@ public class SootMethod
 
 	/**
 	 * 
-	 * @return yes, if this function is a constructor.
+	 * @return yes, if this function is a constructor. Please not that <clinit> methods are not treated as constructors in this method.
 	 */
 	public boolean isConstructor()
 	{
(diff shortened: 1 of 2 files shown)


MarcMil:
See this code in ART:
https://android.googlesource.com/platform/art/+/241b5de%5E!/
Clinits may not have the kAccConstructor flag.

Bug: 11157540
Set the clinit access flag when we load the method and warn about badly
formed access flags.

+  … if (method_name[0] == '<') {
+    // Fix broken access flags for initializers. Bug 11157540.
+    bool is_init = (method_name == "<init>");
+    bool is_clinit = !is_init && (method_name == "<clinit>");
+    if (UNLIKELY(!is_init && !is_clinit)) {
+      LOG(WARNING) << "Unexpected '<' at start of method name " << method_name;
+    } else {
+      if (UNLIKELY((access_flags & kAccConstructor) == 0)) {
+        LOG(WARNING) << method_name << " didn't have expected constructor access flag in class "
+            << PrettyDescriptor(klass.get()) << " in dex file " << dex_file.GetLocation();
+        access_flags |= kAccConstructor;
+      }
+    }
+  }
So, in short: if the method name is <clinit> or <init>, it is expecting the Constructor flag. Otherwise, it warns and patches it on the fly anyway.

Regarding staticInitializerName: Makes sense; didn't know about this before. I will probably update it later this day :).

MarcMil:
I performed the requested changes :)

PR #401, opened 2015-04-30
Logging framework SLF4J for Soot

jpstotz:
The primary logging mechanism of Soot is currently G.v().out. 
Using a PrintStream for logging has multiple disadvantages compared to modern logging frameworks which supports multiple log levels, class/package based filters, allow to trace back which class printed a certain logging statement and log to multiple destinations and formats (file, console,...).

As Heros already includes the libraries of the SLF4J logging framework it was self-evident to use this framework also for Soot logging.

As a first step I redirected G.v().out to a wrapper that outputs each written line to a SLF4J logger. 
The conversion PrintStream -> OutputStream -> Logger bases on apache.commons.exec.LogOutputStream. 

Hence using G.v().out or a real SLF4J logger would end up in the SLF4J logging system. Only the widely used System.println(...) calls within Soot are not captured at the moment. 
They could be replaced step-by-step by real SLF4J loggers.


ericbodden:
Thanks for this. In the Eclipse plugin, Soot does not print to stdout but to the Eclipse console. Did you take this into account in your redesign, i.e., that the printstrm could be changed?

jpstotz:
AFAIR Eclipse provides an MessageConsoleStream for writing content to a console. This MessageConsoleStream is not a child class of PrintStream. Therefore I assume (have not checked the sources) that some PrintStream-to-MessageConsoleStream wrapper has been implemented for redirecting the output of Soot inside of Eclipse. 

At the moment this wrapper will still be functional as before as it is still possible to overwrite G.v().out. 
Only classes directly logging via SLF4J will not be captured.
However it is not a big deal to use SLF4J + logback and implement a custom Appender that "appends" everything logged via SLF4J to the Eclipse console.

wil3:
I finally got around to making a pull request for the same [issue](https://github.com/Sable/soot/pull/407), I didn't see this until afterwards. I took a different approach and just did a replace for all instances however I feel this is a more elegant approach. I think the only issue with forwarding everything from G.v() to SLF4J is that it will be all under a single logger and it will not have as fine grained control of having a separate logger for each class.

jpstotz:
I agree with @wil3 that my approach is not a perfect solution but it would be a first step. 

If someone from the group of Soot maintainers would promise to me that my pull request will be accepted I would take the time to adapt it to the current situation. 

Without such an promise I won't write another pull request for Soot as I don't want to waste my time for another pull request that is molding here in the tracker again for years until it is in a conflict state without any clear feedback what needs to be changed for being accepted.

mbenz89:
Hi Jan.

I would be happy to merge an updated version of this pr. Eventually though, I would like to see something similar in Soot to what @wil3 did, with some additional log levels, i.e., some messages should really be just debug messages and hidden by the default log level.

If you go throw the hassle of updating your pr, I'd like to see two additional things there. First, would you mind using logback as slf4j backend and, second, maybe you can use `logger.info` as default output and if the message is a warning (most of the time denoted by it starting with "[warning]), log it through `logger.warn`. 
We should then be able to successively put the verbose logging stuff into `logger.debug` calls.

I also consider making `G.v().out` deprecated, so that in the future developers are encouraged to use a per class logger.

Best,
Manuel

jpstotz:
AFAIR `logger.info` is the highest log level for non-error related messages. Therefore I would suggest to use `logger.debug` as default. Then there is one higher level (info) and one lower level (trace) for  differentiating the importance of log messages.

OK, let me conclude the list of tasks:
* mark G.v().out deprecated
* every class that contains calls to `G.v().out` gets a new static field 
   * `[org.slf4j.]Logger logger = [org.slf4j.]LoggerFactory.getLogger(<CurrentClassname>.class);`
* replace calls to G.v().out to 
  * if it is a logging output...
    * and it is a general message change it to `logger.debug(...)` 
    * and if it is a message starting with `[warning]` change it to `logger.warn(..)`
  * if it is console output (like in `soot.Main`) change it to `System.out.`
 * add slf4j and Logback to `pom.xml` as dependency. I would suggest using `1.7.5` as this is the version already used in [heros](https://github.com/Sable/heros/blob/develop/pom.xml). If we use the same version we won't get any version troubles.

Regarding logback: Do you really want to directly include Logback? AFAIR slf4j contains a minimum console logger therefore even without Logback the users who use Soot from command-line will get logging output.

On the other side people who use Soot as library don't need logback it even could come to conflicts (different version used or different slf4j-backend implementation). 

Therefore I would suggest not to include Logback into the dependencies.

mbenz89:
It sounds reasonable to use `debug` as default level.

Regarding logback, I am fine with staying to the default behavior of slf4j and leave picking a concrete backend to the user. As far as I remember, Heros is using slf4j-simple which has the same drawbacks as using logback: The user has to remove the dependency via Maven's exclude feature to register another backend.
I did not know you could use slf4j without any backend. If you manage to get this working, I'll remove the slf4j-simple dependency from Heros as well. Btw. it shouldn't be much of a hassle to update Heros' slf4j version, so feel free to use the newest release.

Regarding your task list, it sounds fine! And thanks for considering this contribution, this will definitely be a welcome advancement to Soot! 

Some notes:
- I think the warning prefix isn't consistent. It might be WARNING, warning, Warning, [Warning] and so on. 
- There are calls to G.v().out that are guarded by something like `if(Options.v.().verbose()`, it would be very cool to have these modeled as `logger.trace` and remove the verbose guard. Verbosity could then be defined by setting the logging level to "trace" and the verbose option be removed.
- While keeping `G.v().out` is essential to not break clients using it, we should still employ the PrintStream which uses a logger internally, as you proposed in this pr.

jpstotz:
Sorry I have to revise what I said on slf4j. Without implementation it falls-back to a (NOP) logger implementation. 
Looks that they do not consider that project can be a library and at the same time a program :(
They only consider well-engineered projects that have one library module-project and a separate module-project for the command-line client.

mbenz89:
Sad. I'm favoring logback here because of good experience. I would update Heros to use logback as well if we go for it. I'm open to other suggestions, though.

mbenz89:

jpstotz:
This pull request is replaced by https://github.com/Sable/soot/pull/910.

PR #402, opened 2015-05-01
Optimize a few things.

GUIpsp:
This turns a 15 minute run into an 8 minute run (!).
All these optimizations are safe, as far as I know.

### DIFF
--- a/src/soot/shimple/internal/PhiNodeManager.java
+++ b/src/soot/shimple/internal/PhiNodeManager.java
@@ -261,8 +261,8 @@ public class PhiNodeManager
             // termination, the challengers list never does.  This could
             // be optimised.
             Set<ValueUnitPair> pairsSet = valueToPairs.get(value);
-            List<ValueUnitPair> champs = new ArrayList<ValueUnitPair>(pairsSet);
-            List<ValueUnitPair> challengers = new ArrayList<ValueUnitPair>(pairsSet);
+            List<ValueUnitPair> champs = new LinkedList<ValueUnitPair>(pairsSet);
+            List<ValueUnitPair> challengers = new LinkedList<ValueUnitPair>(pairsSet);
             
             // champ is the currently assumed dominator
             ValueUnitPair champ = champs.remove(0);
@@ -277,23 +277,22 @@ public class PhiNodeManager
 
                 // go through each challenger and see if we dominate them
                 // if not, the challenger becomes the new champ
-                for(int i = 0; i < challengers.size(); i++){
-                    ValueUnitPair challenger = challengers.get(i);
-
-                    if(challenger.equals(champ))
+                for (Iterator<ValueUnitPair> iterator = challengers.iterator(); iterator.hasNext(); ) {
+                    ValueUnitPair challenger = iterator.next();
+                    if (challenger.equals(champ))
                         continue;
                     Unit challengerU = challenger.getUnit();
-                
-                    // kill the challenger
-                    if(dominates(champU, challengerU))
-                        phiExpr.removeArg(challenger);
 
+                    // kill the challenger
+                    if (dominates(champU, challengerU)) {
+                        phiExpr.removeArg(challenger);
+                        iterator.remove();
+                    }
                     // we die, find a new champ
-                    else if(dominates(challengerU, champU)){
+                    else if (dominates(challengerU, champU)) {
                         phiExpr.removeArg(champ);
                         champ = challenger;
                         champU = champ.getUnit();
-                        champs.remove(champ);
                     }
 
                     // neither wins, oops!  we'll have to try the next
--- a/src/soot/shimple/internal/PhiNodeManager.java
+++ b/src/soot/shimple/internal/PhiNodeManager.java
@@ -261,8 +261,8 @@ public class PhiNodeManager
             // termination, the challengers list never does.  This could
             // be optimised.
             Set<ValueUnitPair> pairsSet = valueToPairs.get(value);
-            List<ValueUnitPair> champs = new LinkedList<ValueUnitPair>(pairsSet);
-            List<ValueUnitPair> challengers = new LinkedList<ValueUnitPair>(pairsSet);
+            List<ValueUnitPair> champs = new ArrayList<ValueUnitPair>(pairsSet);
+            List<ValueUnitPair> challengers = new ArrayList<ValueUnitPair>(pairsSet);
             
             // champ is the currently assumed dominator
             ValueUnitPair champ = champs.remove(0);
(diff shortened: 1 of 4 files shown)


StevenArzt:
Thank you for your contribution to the Soot project. There are two issues I would like to discuss before merging your pull request:

1) You changed some lists from ArrayList to LinkedList. Does this really contribute to the performance gain? I guess that constructing an ArrayList with the proper initial size would do the job as well - and would make less assumptions about the code that uses Soot. If we return a LinkedList and people access indices inside there, things might get a low slower.

2) There is a new class that is under the Apache license. Soot is under LGPL v2.1. Can we get the new class under LGPL v2.1 as well?

GUIpsp:
> 1) You changed some lists from ArrayList to LinkedList. Does this really contribute to the performance gain? I guess that constructing an ArrayList with the proper initial size would do the job as well - and would make less assumptions about the code that uses Soot. If we return a LinkedList and people access indices inside there, things might get a low slower.

Well, yes but consider the following:
* We don't know the size of the list before creating, elements are added one at a time
* Why would someone index into that list? They'll probably be iterating over it.
* A LinkedList is the most natural representation. If they require other performance characteristics, they can shove the elements into their preferred colletion.

> 2) There is a new class that is under the Apache license. Soot is under LGPL v2.1. Can we get the new class under LGPL v2.1 as well?

I thought that the LGPL2 was compatible with the apache license. I was wrong. I'll fix this later.

I'm still in the process of profiling and discovering other optimizations too.

GUIpsp:
Oh I forgot to answer your question: 


> Does this really contribute to the performance gain?

Yes. My profiling showed over 50% time spend on the lists (yuck!).

flankerhqd:
Regard the clear() commit, I think it may add some memory pressure as clear() is often called in FlowAnalysis.

GUIpsp:
It's a tradeoff. It's very much worth it if you have big classes.

Also, I consider memory "cheaper" than cpu time.

StevenArzt:
I'm ok with the pull request as soon as the licensing issue is sorted out.

GUIpsp:
This latest optimization is one I'm not so sure about. If I understand the code correctly, it should be fine, however the code is a bit tangled.

Regarding the CompactHashMap, I haven't found an alternative implementation. However, HashMaps keep showing up in my profiles! I believe we should create our own implementation of a closed hash map. However, as I am "tainted" from copying the apache-licensed implementation, I don't think I can do it.

StevenArzt:
We already have some references to Google Guava inside Heros and Soot. If we can rust reference something from that, that would be a very simple option. I'm not sure whether they have anything helpful in this regard, though.

GUIpsp:
Guava does not contain a closed hash map.

GUIpsp:
Down to 5 minutes.

GUIpsp:
Not sure about this last optimization. Seems too obvious, but I couldn't find anywhere that wrote without updating the cache too, so I don't know.

nignag:
@GUIpsp as far as i see commit 640b0338c84f264af6eb67689bc4fc3e9810fe55 is the main reason for your improvment. In generell there are only some rare cases where an ArrayList is slower than a LinkedList (eg. used as working-queue with removefirst) but if it is used as in ArrayPackedSet the ArrayList is the better solution

GUIpsp:
@nignag While that may be true, all of these commits improve something identified as an hotspot by the profiler. This patchset yielded a 15minutes->3minutes improvement!

And I'm still looking to improve more, since even with this patchset, soot is still too slow for my usecase.

StevenArzt:
My main objection against LinkedList is the increased memory footprint. If we need to create a list, we might even be better off by first creating a LinkedList while we don't know how many elements we will have yet, and then convert it to an ArrayList once we're sure. That should cost a little performance, but would save us from the memory overhead of the LinkedList.

GUIpsp:
@StevenArzt Then you'll be doubling the runtime cost on a (very!) hot path. Does the memory cost matter much? How long are those lists even kept around?

GUIpsp:
This is what I am currently looking into optimizing: 
So much HashSet

nignag:
don't touch it :D im currently woring on that

GUIpsp:
Alright, I'll find some other part to optimize.

StevenArzt:
For the list issue, we have a lot of lists that are created unnecessarily anyway. People create a list and then create an iterator on it instead of just iterating over the base collection. This is the case since in the old days, you could not iterate over FlowSets. I changed that: d348314b9084f6de0c6c1ea7c97c6a291018775b This should remove a lot of pressure from FlowSet.toList() - regardless of how it is implemented.

nignag:
```
soot-devel.jar
      Building graphs: 07,5s (16,0%)
  Computing LocalDefs: 03,7s (07,8%)
  Computing LocalUses: 03,3s (07,0%)
     Cleaning up code: 00,0s (00,0%)
Computing LocalCopies: 00,0s (00,0%)
 Computing LiveLocals: 00,0s (00,0%)
Coading coffi structs: 00,0s (00,0%)

       Resolving classfiles: 00,0s (00,0%)
 Bytecode -> jimple (naive): 00,0s (00,0%)
        Splitting variables: 05,7s (12,1%)
            Assigning types: 03,6s (07,7%)
  Propagating copies & csts: 02,5s (05,3%)
      Eliminating dead code: 05,2s (10,9%)
                Aggregation: 05,8s (12,3%)
            Coloring locals: 00,0s (00,0%)
     Generating jasmin code: 00,0s (00,0%)
          .jasmin -> .class: 00,0s (00,0%)
totalTime:47,1s (100,0%)
stmtCount: 485887(10325.49 stmt/s)
totalFlowNodes: 1697069 totalFlowComputations: 2278621 avg: 1.34
Soot finished on Fri May 01 19:28:39 CEST 2015
Soot has run for 0 min. 47 sec.
```

```
soot-nightly-2015-05-22.jar
      Building graphs: 19,4s (20,7%)
  Computing LocalDefs: 30,5s (32,5%)
  Computing LocalUses: 02,2s (02,4%)
     Cleaning up code: 00,0s (00,0%)
Computing LocalCopies: 00,0s (00,0%)
 Computing LiveLocals: 09,3s (09,9%)
Coading coffi structs: 00,0s (00,0%)

       Resolving classfiles: 00,0s (00,0%)
 Bytecode -> jimple (naive): 00,0s (00,0%)
        Splitting variables: 15,9s (17,0%)
            Assigning types: 06,7s (07,2%)
  Propagating copies & csts: 14,2s (15,1%)
      Eliminating dead code: 09,3s (09,9%)
                Aggregation: 17,0s (18,2%)
            Coloring locals: 00,0s (00,0%)
     Generating jasmin code: 00,0s (00,0%)
          .jasmin -> .class: 00,0s (00,0%)
totalTime:93,6s (100,0%)
stmtCount: 487061(5201.64 stmt/s)
totalFlowNodes: 2414491 totalFlowComputations: 3254058 avg: 1.34
Soot finished on Fri May 01 21:11:39 CEST 2015
Soot has run for 1 min. 33 sec.
```

StevenArzt:
@nignag Which ones of today's changes and/or proposed changes did you take into account for the "current" measurement?

nignag:
neither nor, the upper one is an generell improvement of the Forward/Backward - FlowAnalysis and an improved DU/UD Chain (SmartLocalDefs, SimpleLocalDefs, SimpleLocalUses etc). I will commit them in the next one or two weeks i think

GUIpsp:
I still think we should just keep the LinkedList around. This project suffers from a plague of ArrayLists, and for most things you want to do in a parser-like projects, a LinkedList is the right choice. (Remove from top, remove from bottom, get from top, get from bottom, iterate, iterate and remove). Also, memory is cheap-ish, etc.

StevenArzt:
Well, as long as you do not keep the LinkedLists flying around in memory, that's fine. We however run analyses on tens of megabytes of bytecode which easy goes up to gigabytes with Soot's internal objects (ASTs, Callgraph, etc.) This is why I'm so concerned about memory. For a short time, it's not so much of an issue, but once things become persistent over the lifetime of the Soot instance, we're in trouble.

Let's do it like this: Double-check how these lists are used, also take my commit above into account. When we're sure that we're not creating a persistent bloat, we should be fine with the LinkedLists.

GUIpsp:
toList() only seems to be called for some analysis and seems to never be kept around.

StevenArzt:
Great, then I'll have a closer look in the next few days. Looks like the only thing we really need to sort out is the license.

GUIpsp:
I already removed the optimization that used the new HashSet.

By the way, how long do those tens of megabytes of bytecode take to parse?

StevenArzt:
Parsing as such is really a minor issue. However, the type assigner and the preparations for type assignment can take some time. A lot of that goes into the SmartLocalDefs etc. classes that @nignag is looking into, so this should also be sorted out soon.

PR #403, opened 2015-05-01
Refactorings

nignag:
+generics
+some type tightening
+reduced memoryfootprint of SimpleLocalUses


GUIpsp:
Do these include the performance changes you talked about in #402?

nignag:
not that much i think, but at least ArrayPackedSet.toList (commit 3aa1d5c29070b7090b49c43d555f80e005924787) should be a bit faster

PR #407, opened 2015-05-05
Migratation to SLF4J

wil3:
When using Soot as a library there was not enough flexibility to configure what information is written to the console. SLF4J allows the end user to control exactly what they want to be logged and how. This change replaces all (with the exception of soot.Main) System.out.println and G.v().out.println references with the SLF4J logger.


mbenz89:

jpstotz:
There is a new and fresh and conflict free pull request for migrating soot logging to slf4j: https://github.com/Sable/soot/pull/910.

Therefore I would suggest to close this pull request.

PR #412, opened 2015-05-07
Fast LocalDefs

nignag:
+rewritten SimpleLocalDefs
+some reduced HashMap pressure from SmartLocalDefs / SimpleLiveLocals
+complete rewritten FlowAnalysis
+introduced factories for LocalDefs/LiveLocals/LocalUses
+changed LocalSplitter to behave less optimistic (usage of undefined locals)
+removed the usage of SmartLocalDefsPool (it causes memoryleaks, and isn't required anymore)
+added "FlowSet.isSubSet"
+added some Div/Rem Constant checks for 	DeadAssignmentEliminator
+added some asserts to soot.validation.UsesValidator

The behaviour of SimpleLocalDefs and FlowAnalysis has changed a little bit: only reachable nodes within a graph will be visited. It's not a big deal but maybe this causes some errors that have been hidden before (see also soot.validation.UsesValidator.graphEdgesAreValid). This triggers especially bugs in UnitGraph, ExceptionalUnitGraph, etc and in the ThrowAnalysis (or in combination).

The newly introduced pooling was removed, it doesn't offer any benefit with the SimpleLocalDefs/FlowAnalysis and just silently fills the memory.

SimpleLocalDefs uses a mixed strategy that works best when used with the "semi-ssa" code that the LocalSplitter creates, so you should avoiding to use LocalPacker.

### DIFF
--- a/src/soot/baf/toolkits/base/LoadStoreOptimizer.java
+++ b/src/soot/baf/toolkits/base/LoadStoreOptimizer.java
@@ -82,7 +82,6 @@ class Instance {
     // Instance vars.
     private Chain<Unit> mUnits;
     private Body mBody;
-    private ExceptionalUnitGraph mExceptionalUnitGraph;
     private LocalDefs mLocalDefs;
     private LocalUses mLocalUses;
     private Map<Unit, Block> mUnitToBlockMap;     // maps a unit it's containing block
@@ -162,9 +161,8 @@ class Instance {
     
     private void computeLocalDefsAndLocalUsesInfo() 
     {        
-        mExceptionalUnitGraph =  new ExceptionalUnitGraph(mBody);
-        mLocalDefs = new SmartLocalDefs(mExceptionalUnitGraph, new SimpleLiveLocals(mExceptionalUnitGraph));
-        mLocalUses = new SimpleLocalUses(mExceptionalUnitGraph, mLocalDefs);
+        mLocalDefs = LocalDefs.Factory.newLocalDefs(mBody);
+        mLocalUses = LocalUses.Factory.newLocalUses(mBody, mLocalDefs);
     }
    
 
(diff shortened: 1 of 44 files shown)


StevenArzt:
I merged the pull request. For our external test suite, the processing time came down from 24/25 minutes to 22 minutes.

Can you please comment the FlowAnalysis class a bit better? I have a rough idea of what is going on there with the SCCs, but a better documentation (code comments) would be much appreciated. A short outline of the algorithm definitely wouldn't hurt.

There's one thing I don't understand about the new FlowAnalysis. If there is an SCC, you do not need to iterate within it as you already know the results of all flow functions within after a single run. This was however also implicitly done by the old version which did a fixpoint iteration on the nodes as such. If you computed the same flow value again, you knew that it was time to stop. So how does the SCC-based approach help in terms of performance given that you also have the additional overhead for computing the SCCs?

nignag:
Sure i can add some comments ;)
 
the scc has only a slighlty effect, it's just used to reduce the equals to computation, first i thought that would have a greater effect, but it's not that big deal.
what test-suite are you using? on my older system (core2 from 2007) the average soot runtime is reduced by around 30%-50% often more.

StevenArzt:
We run this test suite (https://github.com/Sable/soot-tests) on our nightly build server whenever a new Soot build has been performed. The basic idea of the test suite is to read in Java classes by Soot, write them out again, and run the tests included with these Java classes. We for instance do this with Apache Commons IO. If the pure Java->Java conversion did not break any of their test cases (and did not crash Soot), we use dx to convert them to Dalvik, then use Soot to convert them back to Java and run the tests again.

Concerning performance, this means that quite a lot of the time is spent in the test cases of the target projects, not in Soot. So your improvements might actually have reduced a lot, but only in the Soot part of the test suite, so it's not directly visible.

I'll also run the FlowDroid test suite next which should give us a better idea of the performance improvements. The Soot suite was however better for code coverage inside Soot.

nignag:
Thx :+1: , that same suite i used for testing too

PR #414, opened 2015-05-08
Implementation of a backend based on the ASM framework

LukasSommerTu:


ericbodden:
@StevenArzt Could you please have a look, AFTER the paper deadline?

PR #419, opened 2015-05-11
Change: Use higher API of Android.jar for when there is not target Android.jar

yuki2006:


ericbodden:
Hi. Thanks for the change but I would recommend to make this optional at best.

StevenArzt:
If a user wants to use a higher Android API version, he can use the -force-android-jar option. That option has a slightly different semantics (always use the specified version instead of take it as a default), but that should be sufficient in practice.

PR #427, opened 2015-05-26
Optimize processPhiNode().

GUIpsp:


GUIpsp:
Before anyone asks:
* Yes this was a bottleneck.
* Yes this increases memory usage.
* A bloom filter might be more appropriate if we had a way of knowing the size of the thing.

PR #428, opened 2015-05-26
Add CuckooHashMap to .get() hotspots.

GUIpsp:
Before this can be merged a cuckoo hash map implementation needs to be added. The one I used in my testing is Apache 2.

### DIFF
--- a/src/soot/shimple/internal/PhiNodeManager.java
+++ b/src/soot/shimple/internal/PhiNodeManager.java
@@ -346,7 +346,7 @@ public class PhiNodeManager
         */
     }
     
-    protected Map<Unit, Block> unitToBlock;
+    protected CuckooHashMap<Unit, Block> unitToBlock;
 
     /**
      * Returns true if champ dominates challenger.  Note that false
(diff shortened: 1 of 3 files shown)


StevenArzt:
Do you have numbers on how and whether this improves Performance?

GUIpsp:
How this improves performance? Cuckoo hash maps provide O(1) .get()

I don't have the numbers but as with my other optimizations, these are guided by profiling the application.

Without this optimization+ #427 it took me 18 minutes to class->shimple my data. With these optimizations it takes me 10 minutes.

nignag:
As the default JDK implementation of the Maps performs really well. I don't believe this request would change anything beyond introducing new map-class. if "get" is really a hotspot, than you changing the initial hashmap size should be enough. If it is not helping find out why "get" is a really hotspot

GUIpsp:
> I don't believe this request would change anything beyond introducing a new map-class.

But it did. 

> if "get" is really a hotspot, than changing the initial hashmap size should be enough.

This only works when you know the expected size, and even then, collisions will happen, unless you set the doubling threshold insanely low.

> If it isn't helping, find out why "get" is a really hotspot

It's a hotspot because the maps are being used as they are supposed to. We just need faster gets.

> As the default JDK implementation of the Maps performs really well.

Good. But this one performs better.

GUIpsp:
We already use apache 2 libs. Can we use an apache 2 Map implementation?

GUIpsp:
@StevenArzt

StevenArzt:
In general, we can switch to a different map implementation. To do so, we however need good evidence that it really buys us something and we would need to carefully check where in Soot to use the new list and where to stick with the old one (and why). This is nothing I want to do in the passing.

PR #432, opened 2015-06-05
Annotation instance creation

florian-kuebler:
Exchanged commons-lang lib with jboss-common-core as fixes problems with prim. types.


StevenArzt:
Could you elaborate a bit on why this replacement is necessary?

florian-kuebler:
The general problem is that Java's ClassLoader#loadClass method neither support classnames in the non-canonical representation used in Java Bytecode (and therefore also within soot's AnnotationElem) (i.e. [[[package/name/Classnamepackage.name.Classname instead of package.name.Classname[][][]) nor classes of primitive types (e.g. Integer.TYPE for 'I').

Even though a conversion is not that hard to implement I've wanted to reuse existing solutions, so I added commons-lang as library for this. I did not recognized that the capabilities to load primitive classes is only implemented in private methods and can not accessed transitively, where jboss-core offers both. It loads primitive classes using their char identifier and can accepts non-canonical array representations (one only has to replace '/' with '.')

I would also understand if you don't want to add the jboss lib. to soot. In this case i would implement the case distinction by myself.

PR #441, opened 2015-07-06
alloc-nodes for params of library methods

florian-kuebler:
For library analysis, points-to sets of parameters of public/protected methods are empty if there is no call to them. The resulting callgraph is unsound.

This PR introduces a new spar option "allocate-params". Whenever this option is set to true, allocation nodes to the parameters will be added for AnySupType of params type.


ericbodden:
Thanks that looks like a very useful addition!

florian-kuebler:
Thanks for the response. 

Within the next weeks I will add an opportunity to define which methods are treated as accessible. Furthermore the same feature will be added for the this local, as there exists the same problem.

You might have a better name for the option, that covers the intention and not the implementation.
I am quiet unhappy with the current one, but had no better idea.

PR #442, opened 2015-07-07
Adds a relatively simple validator. 

MarcMil:
It tries to check whether after each new-expression-statement there is a corresponding call to the <init> method before a use or the end of the method.


ericbodden:
Can you please still add the license header to the new file? Thanks!

MarcMil:
Thanks for the notice, I added the license :)

PR #448, opened 2015-07-22
fixed bug in ArrayPackedSet.complement

nignag:


StevenArzt:
I merged it, but it would be great to also have test cases to make sure that bugs which get fixed do not re-appear, especially if they happened in very basic data structures.

ericbodden:
What was the bug you fixed? Please explain...

nignag:
The complement was only calculated up to the highest bit set and not for the complete universe.

PR #451, opened 2015-07-22
Exception-Handler backport

nignag:
A generalisation of SmartLocalDefs control flow


StevenArzt:
I had to revert your pull requests because it breaks some of Soot's own test cases.

PR #457, opened 2015-07-31
lazy getExceptionDests result

nignag:
This allows to test the result for emptyness without calling mightThrow


StevenArzt:
Why is the size always 1?

nignag:
getExceptionDests, returns either the known result or a calculates a new one where the exception is not handeld within the same graph/body, so its a (single) jump out in the void.

I've also slightly improved the pull-request, so things should be more obvious now

PR #464, opened 2015-08-31
DexPrinter: Write out Throws-Annotation

MarcMil:
The dexprinter missed to write out the throws-annotation.

### DIFF
--- a/src/soot/toDex/DexPrinter.java
+++ b/src/soot/toDex/DexPrinter.java
@@ -84,6 +84,7 @@ import soot.SourceLocator;
 import soot.Trap;
 import soot.Type;
 import soot.Unit;
+import soot.dexpler.DexType;
 import soot.dexpler.Util;
 import soot.jimple.Jimple;
 import soot.jimple.NopStmt;
@@ -549,6 +550,7 @@ public class DexPrinter {
         	if (memberClassesItem != null)
         	  annotations.addAll(memberClassesItem);
         }
+    	
         
         for (Tag t : c.getTags()) {
             if (t.getName().equals("VisibilityAnnotationTag")){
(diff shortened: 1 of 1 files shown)


StevenArzt:
You explicitly add annotations. Have you double-checked that you don't get any duplicates when reading in an APK file, writing it out, and reading it in and writing it out once again? I want to avoid that we first explicitly write these annotations, then read them in as generic ones, and then write out those generic annotations plus yet another set of explicitly created ones for the exceptions

MarcMil:
In DexAnnotation, this annotation gets skipped:
} else if (atypes.equals("dalvik.annotation.Throws")) {
                // this is handled in soot.dexpler.DexMethod
                continue; 

In DexMethod, the thrown exception types are added to a list, which becomes the exception list of the method:

        for (Annotation a : method.getAnnotations()) {
            Type atype = DexType.toSoot(a.getType());
            String atypes = atype.toString();
            if (!(atypes.equals("dalvik.annotation.Throws")))
                continue;
            for (AnnotationElement ae : a.getElements()) {
                EncodedValue ev = ae.getValue();
                if(ev instanceof ArrayEncodedValue) {
                    for(EncodedValue evSub : ((ArrayEncodedValue) ev).getValue()) {
                        if(evSub instanceof TypeEncodedValue) {
                            TypeEncodedValue valueType = (TypeEncodedValue) evSub;
                            String exceptionName = valueType.getValue();
                            String dottedName = Util.dottedClassName(exceptionName);
                            thrownExceptions.add(SootResolver.v().makeClassRef(dottedName));
                        }
                    }
                }
            }
        }
...

            sm = new SootMethod(name, parameterTypes, returnType, accessFlags, thrownExceptions);

Thus, everything should be fine.

Issue #468, opened 2015-09-08
Soot output naming conflict


Waldo1637:
If Soot attempts to write a jar file along with other outputs (ex: dumping call graphs), one of the outputs will fail because it attempts to create a jar file and a directory with the same name. 

For example, if the sootOutput directory is empty and you run Soot with options
```
-outjar -dump-cfg bb
```
the call graphs will be dumped but the following error occurs when trying to write the jar file:
```java
soot.CompilationDeathException: Cannot open output Jar file sootOutput\out.jar
	at soot.PackManager.setupJAR(PackManager.java:548)
	at soot.PackManager.writeOutput(PackManager.java:518)
	at soot.Main.run(Main.java:273)
	at soot.Main.main(Main.java:147)
```

In an alternative scenario, the sootOutput directory already contains a jar file named out.jar. Then when you run Soot, it will recreate the jar file but will fail to dump the CFGs with the following error:
```java
java.io.IOException: Unable to mkdirs sootOutput\out.jar\inst.test.Test1\int A0(int)
	at soot.util.PhaseDumper.makeDirectoryIfMissing(PhaseDumper.java:169)
	at soot.util.PhaseDumper.nextGraphFileName(PhaseDumper.java:192)
	at soot.util.PhaseDumper.dumpGraph(PhaseDumper.java:374)
	at soot.toolkits.graph.BriefUnitGraph.<init>(BriefUnitGraph.java:71)
	at soot.toolkits.graph.ZonedBlockGraph.<init>(ZonedBlockGraph.java:64)
	at soot.baf.toolkits.base.LoadStoreOptimizer$Instance.buildUnitToBlockMap(LoadStoreOptimizer.java:125)
	at soot.baf.toolkits.base.LoadStoreOptimizer$Instance.go(LoadStoreOptimizer.java:93)
	at soot.baf.toolkits.base.LoadStoreOptimizer.internalTransform(LoadStoreOptimizer.java:79)
	at soot.BodyTransformer.transform(BodyTransformer.java:51)
	at soot.Transform.apply(Transform.java:105)
	at soot.BodyPack.internalApply(BodyPack.java:49)
	at soot.Pack.apply(Pack.java:126)
	at soot.baf.BafBody.<init>(BafBody.java:122)
	at soot.baf.Baf.newBody(Baf.java:560)
	at soot.PackManager.convertJimpleBodyToBaf(PackManager.java:978)
	at soot.PackManager.runBodyPacks(PackManager.java:934)
[...]

PR #470, opened 2015-09-09
Fix issue 468

Waldo1637:


ericbodden:
Thanks! Could you please describe your change in a few more words, for future reference? What problem does it address, and how?

StevenArzt:
It looks as if this directly implements the proposed fix for issue #468. To me, it looks fine content-wise.

Waldo1637:
In the previous implementation, `SourceLocator#getOutputDir()` would sometimes return the name of a directory and sometimes the name of a jar file depending on the values of `Options#output_jar()` and `Options#output_dir()`. However, when used throughout the rest of Soot, most places expected it to always return a directory name and one place (specifically `PackManager#setupJar()`) expected it to return the name of the jar file. This lead to the issues described in #468.

The new implementation adds a new method `SourceLocator#getOutputJarName()` to clearly differentiate this functionality between the two methods. This new method returns the name of a jar file located in the directory returned by `SourceLocator#getOutputDir()` if `Options#output_jar()` returns `true`, otherwise it returns an empty string. The new implementation of both methods handles the two possible cases for specifying the `-output-dir` (or `-d`) option:
* the path given ends with `.jar` -- we assume the user is specifying the name of the jar file so `SourceLocator#getOutputJarName()` will return the full path given by the user and `SourceLocator#getOutputDir()` will return the path given with the jar file name removed from the end
* the path given does not end with `.jar` -- we assume the user is specifying only the output directory so `SourceLocator#getOutputJarName()` will return the full path given by the user appended with a default jar name (`out.jar`) and `SourceLocator#getOutputDir()` will return the full path given by the user

PR #488, opened 2015-10-14
Library call graphs

florian-kuebler:
Added new options (cg library:*) to increase soundness of call graphs generated using the PAG when the target is a library rather than an application.
For all "accessible" method parameters, this local, caught exceptions... and fields, allocation sides to any possible subtype were added.
If simulate-natives is also activated, the return value of native methods will be over-approximated, too.
Accessibility can be defined using Scene#setClientAccessibilityOracle.

For the library mode library:signature-resolution, method calls with base type AnySubType[T] will be resolved in a call-by-signature fashion.


StevenArzt:
@johanneslerch Do you think the code is ready for production use? Apparently, you have already reviewed it at least partly.

johanneslerch:
@StevenArzt sorry, i obviously missed your question.
To me the changes look good and i have been using this already for quite a while. So i think the changes are solid and can be merged, especially with the knowledge that it is also backed by a good test suite (different pull request).

StevenArzt:
Please do at least a test build using out ant script before submitting pull requests. This pull request introduces code that does not compile with Java 1.6 source level. I know that the old source compatibility level is annoying, but we have people who rely on it. Personally, I'd love to get rid of that legacy as soon as possible, but for now, we have to deal with it.

The core message, however, stays the same: Test the builds. The ant script fails and the Eclipse projects outputs the same error - so please do check before you commit. I'll fix the issue for now.

PR #497, opened 2015-10-31
Java 6 Compatibility Fixes

domainexpert:
Updated smali libraries libs/dexlib2-2.1.0-dev.jar and libs/util-2.1.0-dev.jar to ones compiled using Java 6, also updated BlockGrahConverter class for Java 6 syntax compatibility.


ericbodden:
Hi. Is the change to BlockGraphConverter really necessary? Is this not Java 5 syntax it is using?

Also I wonder: @StevenArzt What is Soot's current source-code compatibility level? Is it Java 6?

StevenArzt:
@ericbodden It's Java 6 and Soot builds all fine with Java 6 code compatibility level in my Eclipse installation. This, however, only refers to Soot. Libraries that we use might be built for other Java versions and I find it hard to ensure continued compatibility to a totally outdated version of Java. I run Soot on Java 7 or 8, depending on the machine. The old Java 6 source compatibility level is still there, but we might have to drop that at some point anyways.

domainexpert:
My `javac` version is 1.6.0_65 from Mac OS X 10.6.8. Without the patch to BlockGraphConverter, it complains about illegal start of expression as follows:
```
/Users/andrew/Documents/workspace/soot/src/soot/toolkits/graph/BlockGraphConverter.java:197: illegal start of expression
    [javac]             return (Collections.<Unit>emptyList()).iterator();
    [javac]                                 ^
    [javac] 1 error
```
Using the smali jar libraries that currently come with Soot, `javac` reports no errors, but just a lot of warnings for example:
```
[javac] warning: org/jf/dexlib2/dexbacked/DexBackedOdexFile.class(org/jf/dexlib2/dexbacked:DexBackedOdexFile.class): major version 51 is newer than 50, the highest major version supported by this compiler.
    [javac] It is recommended that the compiler be upgraded.
```
Perhaps I should just follow what that last line instructs, but on Mac that means upgrading my OS, which I'm afraid to do for fear that it will slow down everything.

StevenArzt:
I will merge the pull request, but I assume that you will run into all sorts of funny problems with such a compiler. The old code definitely was Java 6-compatible and we have such constructs in other places inside Soot as well.

domainexpert:
Thank you @StevenArzt , but I actually don't mind upgrading the source compatibility to Java 7 whenever you decide to do so. But from this what I could think of is that regression test suite would be important for Soot in the long run (I don't know if there's any at present). I mean, the decision would have been easier in adopting (or otherwise) anyone's pull request had Soot got a regression test suite.

PR #516, opened 2015-12-30
Spelling

wjk20120522:
Spelling mistake


patricklam:
1) that should be in pretty much all Soot files, so it would be inconsistent to fix in just one file.
2) the goal was for Soot to work on not just Java but other bytecode targets as well.

wjk20120522:
understood ^_^

PR #531, opened 2016-02-07
Added Target Descriptions to build.xml

domainexpert:
Added target descriptions to main Ant targets in `build.xml`. Now users can view a brief explanation on the main targets with `ant -p`. Also added Eclipse `.settings` to `.gitignore`, and added option bootclasspath to the javac task in `build.xml` for cross-compiling to Java 6, in case Java 6 is available in the system. This also prevents compiler warning, in case the user's system has Java 6.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,10 +3,13 @@
 /classes
 /.classpath
 /sootOutput
-# IntelliJ IDEA project dir
+# IntelliJ IDEA project directory
 .idea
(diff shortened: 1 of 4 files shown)


domainexpert:
I removed bootclasspath setting in domainexpert/soot@d92528f thinking that it's better to live with compiler warning than to worry about cross compilation to Java 6.

PR #557, opened 2016-03-16
Throw a RuntimeException if an entry in -process-path or -cp does not exists.

florian-kuebler:


StevenArzt:
I have just added an option to silently ignore classpath errors, i.e., get back the old behavior. by default, we throw the exception if something's wrong, but a user can explicitly chose to ignore them. We need this, because some of our projects based on Soot a relying on the possibility to have multiple possible places where a JAR file can be on the classpath. The file will always only be in one of them, but where it actually is depends on the configuration of the machine. Yes, we could check before passing the classpath to Soot, but I don't want to refactor all of that code.

Issue #565, opened 2016-04-01
No Support for multiple dex files in APK


vitaliiavdiienko:
Hi all,

a few issues:
1) Soot doesn't load classes from all .dex files
2) Soot tries (after instrumentation) to write all classes to one .dex file and doesn't even try to split them across different .dex files (and reaches limits of .dex files quickly...).

This makes Soot a useless framework for analyzing real-world Android apps.

Do you plan any activities in this direction?

Thank you in advance.

Best Rergards,
Vitalii

ericbodden:
Hi Vitalii.

> a few issues:
> 1) Soot doesn't load classes from all .dex files
> 2) Soot tries (after instrumentation) to write all classes to one .dex file and doesn't even try to split them across different .dex files (and reaches limits of .dex files quickly…).

Yes those are known issues.

> This makes Soot a useless framework for analyzing real-world Android apps.

That’s a harsh statement. I would not say that one follows from the other.

> Do you plan any activities in this direction?

Yes we are. @StevenArzt would know more about the current state of things. But contributions are also always welcome!

Cheers,
Eric

vitaliiavdiienko:
Hi Eric,
> 
That’s a harsh statement. I would not say that one follows from the other.

Agree. I was under the impression of an app with 9 classes.dex inside :smile: 

> But contributions are also always welcome!

I have a code that supports loading of all *.dex files from the APK and I actually use it (I can share it also), but writing .jimple classes to *.dex files back seems to be more complicated and experts can do it faster and better.

ericbodden:
> I have a code that supports loading of all *.dex files from the APK and I actually use it (I can share it also), but writing .jimple classes to *.dex files back seems to be more complicated and experts can do it faster and better.

Sure, thanks. Let’s see whether Steven can enlighten us about the current status on this. I am not quite sure about what has already been implemented.

Cheers,
Eric

StevenArzt:
If you need to load multiple dex files into your Soot scene, you have to do this on your own at the moment. You extract the additional dex files from the APK file and then specify the APK (for its main classes.dex) as well as the additional dex files in the process-dir option. This will merge all files together.

Automating this merging is something we can implement as an optional feature. If you have code for it, feel free to share it. There are, however, good reasons for not making this the default:

1) What happens if there is a class Foo in two of the additional dex files? If the main code loads either dex file depending on some condition, that's no problem. In our Soot scene, we would, however, have them both at the same time which is impossible.

[...]

PR #580, opened 2016-04-21
Added processing of multiple DEX files in an APK

vitaliiavdiienko:
Hi @StevenArzt and @ericbodden,

there is my solution for the issue that we discussed in #565.

Maybe not the best one but can be helpful :smile: 

Best Regards,
Vitalii

### DIFF
--- a/doc/soot_options.htm
+++ b/doc/soot_options.htm
@@ -658,6 +658,11 @@ packages in the Java runtime.</p>
 </td><td colspan="2">Prepend the given soot classpath to the default classpath.</td>
 </tr>
 <tr>
+<td><tt>-process-multiple-dex </tt>
+<br>
+</td><td colspan="2">Process all DEX files found in APK.</td>
+</tr>
+<tr>
 <td><tt>-process-path <var>dir</var></tt>
 <br>
 <tt>-process-dir <var>dir</var></tt>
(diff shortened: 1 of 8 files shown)


StevenArzt:
Please add an option to make this behavior optional. As I wrote in the other thread, there are multiple ways in which this can break existing analyses and software. I agree that it is helpful in some cases, but not in all of them. One such example where we are dependent on the old behavior is, for instance, a fuzzing-based approach we are currently working on where we only want to merge what is on the current trace.

vitaliiavdiienko:
Please check my last commit.

ericbodden:
@vitaliiavdiienko Thanks for this. Could you still please add a few comments to the code (where suitable)?

Also, very importantly, Soot options should only be added by modifying soot-options.xml. Our build script then generates the required options code automatically.

StevenArzt:
Yes, that's one final important point. Add new options to options.xml and have the ant build script re-generate the code files and documentation accordingly. Otherwise, your changes will be overwritten during our automatic builds. Thanks a lot for your contribution!

vitaliiavdiienko:
Thanks! I have just learned the correct way of adding options. Here we go 53057c7f6cd615086dade582489035d01612359f

PR #583, opened 2016-04-29
Feature/insert on edge

arpith20:
```insertOnEdge(toInsert, source, target)``` adds instrumentation in a manner such that the resulting [CFG](https://en.wikipedia.org/wiki/Control_flow_graph) of the program will contain ```toInsert``` on an edge that is defined by ```source``` and ```target```.


StevenArzt:
The code change seems reasonable. However, please add Javadoc comments for your new methods. Otherwise, it's really hard for the user to understand what they do.

Btw, I appreciate the comments inside the method. They are really good.

arpith20:
@StevenArzt Thank you for your feedback. I added Javadoc comments to the methods.

PR #603, opened 2016-06-03
update android api version fetching function

zhhailon:
1. Using platformBuildVersionCode rather than targetSdkVersion and minSdkVersion as the API version for an APK.
2. Add jre/rt.jar to class path if using JDK.

### DIFF
--- a/src/soot/ClassSource.java
+++ b/src/soot/ClassSource.java
@@ -33,5 +33,9 @@ public abstract class ClassSource
      * or Types referenced by the class. */
     public abstract Dependencies resolve( SootClass sc );
     protected String className;
+    
+    public void close() {
+    	
+    }
 }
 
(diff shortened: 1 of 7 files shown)


StevenArzt:
Is "platformbuildversioncode" really a documented schema element? It seems to be valid and some APKs contain it, but I couldn't find any proper documentation on that element. In the Android developer docs, there is only the minSdkVersion and the targetSdkVersion. In general, what is the reasoning behind the proposed change?

Concerning the jre/rt.jar file on the default classpath: This should be done conditionally, only if the normal rt.jar does not exist. Otherwise, we might be picking up unnecessary files. I can see that the rt.jar has a different path whe nrunning the JDK's own java executable instead of the one from the JRE, but as I said, make it conditional.

zhhailon:
'platformBuildVersionCode' is add by aapt during the compilation of resources. It's value is the versionCode of the framework resources that an app is built against. This feature is [add in 2014](https://android.googlesource.com/platform/frameworks/base/+/ad2d07d2d98a46babb2a9472413fe9ce5080ca76), thus some apps may not have that.

The reason of using 'platformBuildVersionCode' is that it is the version actually used to compile the code, while 'targetSdkVersion' is claimed by developers.

PR #617, opened 2016-07-27
Adds Maven pom.xml configuration

akbertram:
The project can now be built simply with `mvn install`. As a temporary workaround, I've pushed any dependencies that aren't available in Maven Central to our Maven Repository at https://nexus.bedatadriven.com. 

I've also taken the liberty of using the groupId `ca.mcgill.sable` which seems more appropriate than an unqualified `soot`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,0 +1,154 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>ca.mcgill.sable</groupId>
+  <artifactId>soot</artifactId>
+  
+  <name>Soot</name>
+  <version>2016-07-27</version>
+  <description>A Java Optimization Framework</description>
+  
+  <properties>
+    <maven.compiler.source>1.7</maven.compiler.source>
+    <maven.compiler.target>1.7</maven.compiler.target>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+
+  <build>
+    <sourceDirectory>src</sourceDirectory>
+    <testSourceDirectory>tests</testSourceDirectory>
+
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>generate-sources</phase>
+            <goals><goal>add-source</goal></goals>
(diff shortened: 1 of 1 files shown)


StevenArzt:
Thank you for creating the POM file. I'm not so happy with the remote repository, though. Some dependencies are updated quite frequently and we would make sure that we're not dragging in old components. Can't we just have them in a local repository that is located in, e.g., the "libs" directory? This would retain maximum compatibility and would also be more suitable for nightly builds, unless we really want to upload to some repository every night,

akbertram:
I would definitely recommend versioning and uploading the nightly builds of the dependencies as well as Soot itself. Right now, with the jars in the lib directory, it's very difficult to figure out where they come from and to what source they correspond. 

The approach that's worked really well with [Renjin](https://github.com/bedatadriven/renjin) is setting up a job for each project on Jenkins (or whatever CI) that builds and deploys a uniquely versioned artifact following each commit to the master branch. You could also stick with nightly builds and just use the date as a version. So then soot-201608014.jar would transparently depend on heros-20160801.jar or heros-b28.jar or whatever, so long as each build gets its own number.

This works well because downstream projects (like Renjin!) can safely depend on a specific build of these libraries that have been tested against Renjin, without imposing the burden of formal releases on Soot or heros. 

I've set up heros as an example on our own build server: 
* Each commit to the develop branch triggers a Jenkins [build](.
* If successful, the heros-b${BUILD_NUMBER}.jar is deployed to the maven repository at [nexus.bedatadriven.com](https://nexus.bedatadriven.com/content/groups/public/heros/heros/)

Depending on how the projects interact, you can either set up Soot then to always use the latest available build of heros, or manually bump the version of heros used by Soot as required. 

Happy to host the build process at renjin.org, or can also help set up a similar process on your side.

maenu:
I build with Maven and Soot is not a standalone tool for me. I don't want to build my own version of Soot, but reuse it as a dependency. It lives on the class path with a lot of other artifacts, but it does not integrate well when used as a dependency. Without a POM I have only unsatisfying options:

- Take the shaded JAR, with all it dependencies inside. Sounds good enough, but actually conflicts with Maven's dependency resolution: E.g. slf4j is shaded and I have no way of excluding it without ripping it out of the JAR.
- Take the shallow classes JAR: Now I have to manage the dependencies manually. That's Maven's job, not the developers.

If you hesitate because of frequently updated dependencies, then this tells me that there is a problem with versioning in those dependencies too. I'd propose the same approach as @akbertram: Publish each build as a release and depend on that. Then evolve that into working with milestones, where you release less often, but with clearly defined changes. You can still continuously build snapshots.

akbertram:
@maenu +1 this sounds very similar to the challenges we're having integrating Soot. @StevenArzt Like I said before, I'm happy to setup and host all of the projects on build.renjin.org. Heros is already there, and if you don't mind merging this pom.xml than we can setup the same for Soot.

StevenArzt:
Then let's adopt this pom file as a secondary channel through which Soot can be built in addition to the ant script. I feel that sorting out some organizational issues for the overall project is overdue anyway, but that's a different story.

PR #621, opened 2016-08-12
Adds type-based resolution of invoke() calls

jtoman:
Adds an unsound approach for statically resolving calls to invoke().
The OnFlyCallGraphBuilder hooks into Spark and uses the reaching types
of the 'obj' argument and contents of the 'args' array to find methods
to invoke. In some cases it is possible to compute the positions in
the argument array of these reaching types, in which case the
resolution can be made more precise. This types-for-invoke approach
is still unsound:

* It doesn't handle arrays receivers
* It doesn't handle static methods
* It fails to account for default nulls in the argument array

### DIFF
--- a/src/soot/jimple/spark/solver/OnFlyCallGraph.java
+++ b/src/soot/jimple/spark/solver/OnFlyCallGraph.java
@@ -22,7 +22,12 @@ import soot.Context;
 import soot.Local;
 import soot.MethodOrMethodContext;
 import soot.Scene;
+import soot.Type;
+import soot.jimple.IntConstant;
+import soot.jimple.NewArrayExpr;
+import soot.jimple.spark.pag.AllocDotField;
 import soot.jimple.spark.pag.AllocNode;
+import soot.jimple.spark.pag.ArrayElement;
 import soot.jimple.spark.pag.MethodPAG;
 import soot.jimple.spark.pag.Node;
 import soot.jimple.spark.pag.PAG;
@@ -72,7 +77,7 @@ public class OnFlyCallGraph {
     private void processReachables() {
         reachableMethods.update();
         while(reachablesReader.hasNext()) {
-            MethodOrMethodContext m = (MethodOrMethodContext) reachablesReader.next();
+            MethodOrMethodContext m = reachablesReader.next();
             MethodPAG mpag = MethodPAG.v( pag, m.method() );
             mpag.build();
             mpag.addToPAG(m.context());
@@ -80,7 +85,7 @@ public class OnFlyCallGraph {
     }
     private void processCallEdges() {
         while(callEdges.hasNext()) {
-            Edge e = (Edge) callEdges.next();
+            Edge e = callEdges.next();
             MethodPAG amp = MethodPAG.v( pag, e.tgt() );
             amp.build();
             amp.addToPAG( e.tgtCtxt() );
@@ -89,6 +94,20 @@ public class OnFlyCallGraph {
     }
 
     public OnFlyCallGraphBuilder ofcgb() { return ofcgb; }
+    
+    public void updatedFieldRef(final AllocDotField df, PointsToSetInternal ptsi) {
+    	if(df.getField() != ArrayElement.v()) {
+    		return;
+    	}
+    	if(ofcgb.wantArrayField(df)) {
+    		ptsi.forall(new P2SetVisitor() {
+				@Override
+				public void visit(Node n) {
+					ofcgb.addInvokeArgType(df, null, n.getType());
+				}
+			});
+    	}
+    }
 
     public void updatedNode( VarNode vn ) {
         Object r = vn.getVariable();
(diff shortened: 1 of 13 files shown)


src/soot/jimple/spark/solver/OnFlyCallGraph.java
```diff
@@ -72,15 +77,15 @@ public void build() {
     private void processReachables() {
         reachableMethods.update();
         while(reachablesReader.hasNext()) {
-            MethodOrMethodContext m = (MethodOrMethodContext) reachablesReader.next();
```
jtoman: My eclipse auto-remove this pointless cast. Although strictly not part of the functionality of this commit I see no reason to remove this?

src/soot/jimple/spark/solver/OnFlyCallGraph.java
```diff
@@ -72,15 +77,15 @@ public void build() {
...
+            MethodOrMethodContext m = reachablesReader.next();
             MethodPAG mpag = MethodPAG.v( pag, m.method() );
             mpag.build();
             mpag.addToPAG(m.context());
         }
     }
     private void processCallEdges() {
         while(callEdges.hasNext()) {
-            Edge e = (Edge) callEdges.next();
```
jtoman: see above

src/soot/jimple/spark/solver/PropWorklist.java
```diff
@@ -69,6 +69,9 @@ public final void propagate() {
 						public final void visit(Node n) {
 							AllocDotField nDotF = pag.makeAllocDotField(
 									(AllocNode) n, target.getField());
+                            if(ofcg != null) {
```
jtoman: I have not been able to figure out the indentation style preferred by Ondrej in the Spark code, so the indentation in this change is inconsistent. Let me know the appropriate scheme to use and I will happily adjust.

src/soot/jimple/toolkits/callgraph/InvokeCallSite.java
```diff
@@ -0,0 +1,54 @@
+package soot.jimple.toolkits.callgraph;
+
+import soot.Local;
+import soot.SootMethod;
+import soot.jimple.InstanceInvokeExpr;
+import soot.jimple.Stmt;
+import soot.jimple.toolkits.callgraph.ConstantArrayAnalysis.ArrayTypes;
+
+public class InvokeCallSite {
+	public static final int MUST_BE_NULL = 0;
```
jtoman: These constants overlap with the NullnessAnalysis constants but they are protected so they are duplicated here.

src/soot/jimple/toolkits/callgraph/OnFlyCallGraphBuilder.java
```diff
@@ -82,12 +100,60 @@
  */
 public final class OnFlyCallGraphBuilder
 { 
+	private static final PrimType[] CHAR_NARROWINGS = new PrimType[]{
```
jtoman: These constants are for widening primitive argument types. If there is functionality in the soot codebase that would handle this already I would love to get rid of these.

src/soot/jimple/toolkits/callgraph/OnFlyCallGraphBuilder.java
```diff
@@ -399,10 +507,15 @@ public OnFlyCallGraphBuilder( ContextManager cm, ReachableMethods rm ) {
         }
         
         if(options.reflection_log()==null || options.reflection_log().length()==0) {
-        	reflectionModel = new DefaultReflectionModel();
+        	if(options.types_for_invoke() && new SparkOptions(PhaseOptions.v().getPhaseOptions("cg.spark")).enabled()) {
```
jtoman: I could not find a better way to determine if spark was enabled.

src/soot/jimple/toolkits/callgraph/OnFlyCallGraphBuilder.java
```diff
@@ -414,57 +527,375 @@ public void processReachables() {
...
+			 */
+			for(PrimType narrowings : narrowings(primType)) {
+				if(reachingTypes.contains(narrowings.boxedType())) {
+					return true;
+				}
+			}
+			return false;
+		} else {
+			// impossible?
+			return false;
```
jtoman: I could be missing something here but I think this branch should not be reachable.

jtoman:
There are still some question marks on this commit where I'm not sure I've done things correctly. I have noted these places with comments in the commit.

jtoman:
I see now that I'm using the diamond operator which is unsupported in 1.6. I will fix and update this PR.

jtoman:
I've realized there are two problems with this commit, both in the constant array analysis. First, on control-flow merge if a position in the array is not assigned into that information will be lost, e.g.

```java
Object[] foo = new Object[1];
if(*) {
   foo[0] = "blah";
}
invoke(..., foo);
```

In this case, the analysis should conservatively conclude that any unary method is callable as the implicit null in position 0 could potentially match any argument type.

Also, there is a precision bug that refuses to use the constant array information if the argument array may be null:

```java
Object[] args;
if(*) {
   args = new Object[]{"fizz", "bizz"};
} else {
   args = null;
}
invoke(args);
```

In this case the analysis should simply use the may-be-null information and constant array information to resolve invoke to nullary methods and methods that take two strings.

jtoman:
Alright, I foresee no further big changes.

I understand that this PR is quite big, if it is preferable I can try to split it up into more manageable chunks. Happy to help in whatever way I can.

src/soot/jimple/spark/solver/OnFlyCallGraph.java
StevenArzt: This ocde is quite old, and back in the old days, there were no generics in Java. Consequently, we still have some pointless casts in the code that date back to times when everything was java.lang.Object.

src/soot/jimple/spark/solver/PropWorklist.java
StevenArzt: Let's not discuss the coding style in the older parts of Soot, because this won't lead us anywhere. Just stick to what, e.g., the Eclipse code formatter produces or what yu see in newer parts of the code, ad you're fine.

src/soot/jimple/toolkits/callgraph/InvokeCallSite.java
StevenArzt: Fine with me, unless there are any cross-dependencies that require them to be kept in sync. If they are distinct and just happen to be similar, that's ok.

StevenArzt:
You might be interested in what I did here: f0165f9dadee9eb633f2c9a0ab15ef4eec289c5d SPARK can now handle object allocation through Class.forName() / Class.newInstance() if the class name is a constant string. That ties in perfectly with what you built for method calls.

PR #624, opened 2016-08-16
Fixes type masks for classes with phantom parents

jtoman:
When a class has a parent that is a phantom class, it will not be processed during the breadth first search in TypeManager, as it is technically not reachable from the java.lang.Object type. This causes a missing type mask that will lead cause an error for some use patterns: specifically if a class is instantiated reflectively (via Soot's dynamic classes).

### DIFF
--- a/src/soot/jimple/spark/internal/TypeManager.java
+++ b/src/soot/jimple/spark/internal/TypeManager.java
@@ -21,6 +21,7 @@ package soot.jimple.spark.internal;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


src/soot/jimple/spark/internal/TypeManager.java
```diff
@@ -197,6 +209,9 @@ final public void caseAnySubType(AnySubType t) {
     }
 
     final private BitVector makeClassTypeMask(SootClass clazz) {
+    	if(typeMask.get(clazz.getType()) != null) {
```
jtoman: We need this because we no longer traverse in topological order.

src/soot/jimple/spark/internal/TypeManager.java
```diff
@@ -138,6 +147,9 @@ final public void makeTypeMask() {
                 if (sc.isInterface()) {
                     makeMaskOfInterface(sc);
                 }
+                if(!visitedTypes.get(t.getNumber()) && !((RefType)t).getSootClass().isPhantom()) {
```
jtoman: If we didn't visit this type already and we're not phantom, then we need to make sure to process the type here.

StevenArzt:
Can you comment on why the BitVector is not redundant? You're checking the typeMask map anyway, so in my opinion, having this extra vector is just a duplicate check and a duplicate data structure to maintain.

PR #632, opened 2016-08-31
Change the exit code when RuntimeException occurs

gsridhar53:
When using scripts to run soot, changing the exit code will help in determining if the execution of soot failed.


ericbodden:
Seems way overdue, thanks!

PR #635, opened 2016-09-13
Updated pom.xml and project README

akbertram:
Synced the pom with the dexlib2 dependencies checked into the repository. Is there any difference between the version of dexlib2 manually checked and the [official release](https://bitbucket.org/JesusFreke/smali/downloads) of dexlib2 2.3? I have linked to the [artifact](http://search.maven.org/#artifactdetails%7Corg.smali%7Cdexlib2%7C2.1.3%7Cjar) hosted on Maven Central.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![Build Status](http://build.renjin.org/job/Soot/job/soot/badge/icon)](http://build.renjin.org/job/Soot/job/soot/)
+
 # What is Soot?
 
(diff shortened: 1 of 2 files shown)


StevenArzt:
There shouldn't be any difference. We uograde to the newest version of dexlib every now and then, make the necessary modifications to Soot (if any) and then check the updated code back in together with a copy of the library so that it's available to people who use Soot from within Eclipse or some other IDE.

PR #637, opened 2016-09-13
Basic support for Method Handle Constants (WIP)

akbertram:
Added enough support to ensure that a classfile with MethodHandles can be read in and written back out by Soot. Still need to add a test case for actually invoke method handle values.

I'm assuming that there's no interest in adding this to Jasmin, but @StevenArzt let me know if this is important and where to find the latest version of the project's fork of Jasmin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ a dependency via Maven, Gradle, SBT, etc using the following coordinates:
   <dependency>
     <groupId>ca.mcgill.sable</groupId>
     <artifactId>soot</artifactId>
-    <version>RELEASE/version>
+    <version>RELEASE</version>
   </dependency>
 </dependencies>
 <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -21,6 +21,15 @@
     <sourceDirectory>src</sourceDirectory>
     <testSourceDirectory>tests</testSourceDirectory>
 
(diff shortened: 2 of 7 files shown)


akbertram:
Next step seems to be to handle signature polymorphism in general. Are there any existing plans to handle this inside Soot that I should follow?

StevenArzt:
Jasmin is legacy code and I'm all for dropping it entirely in a new, legacy-free of Soot once we do a major overhaul. Integrating features of modern Java language versions into Jasmin is a lost case anyway, because Jasmin doesn't write out the type annotations that are required by modern JVMs. We have the ASM backend, and it will become the default in the foreseeable future.

StevenArzt:
The test case MethodHandleTest.testInvoke() fails, because it looks for an invoke() method that retrurns a double, but the invoke() method always returns an object. From your commit message, I assume that this test is expected to fail, so I can mark it with a JUnit ignore annotation. Is that correct? Are you going to work on this issue in the future?

akbertram:
Yes, this can be `@Ignore`d for the moment. MethodHandle constants should now work properly but actually invoking them is the focus of this failing test. I'll open a separate issue for signature polymorphic methods so the method handles constant issue can be closed.

PR #645, opened 2016-09-23
Load Dex in Jar's and Zip's not just Apk's and fix issue when loading dex from directory

agorski3:
This fixes two issues.

1. It adds the ability to load dex classes from jar's and zip's instead of just apk's. This is useful when trying to load the jar files contained within the /system/framework of android. 

2. It fixed an issue where the getClassesUnder method in SourceLocator searched for dex files recursively in the file/directory tree of a directory but the buildDexIndex method in DexClassProvider searched for dex files only in the listing of files in a given directory. This created an issue where class names were discovered by getClassesUnder and then their source could not be found in buildDexIndex.

On a related note you may want to fix the other issue with the getClassesUnder method in SourceLocator and dex files in directories. That being since getClassesUnder does a recursive search for dex, jimple, class, and java files, it reconstructs the package name of each class by appending the current directory's name to all the class names returned after a recursive call. This makes sense for jimple, class, and java files as it is common to portray their package through the directories in a path to a class. It does not really make sense for dex files as these contain multiple classes with their full class name already stored inside the dex file. So appending the directories in a path to a dex file to all the class names in the dex makes no sense and is likely incorrect.

Finally, in DexlibWrapper I noticed the line "int api = 23; // TODO:" which hard-codes the api version used when loading dex files. I am not sure if this is important but is seems like it is important since dex2lib uses the api version to lookup the dex opcodes to be used when parsing the dex file. It seems like something the should probably be an option in soot.


agorski3:
Because of a need to be able to process dex from android N, I updated the dexlib2 to version 2.2b3 and fixed a number issues that arose because of design changes in dexlib2 2.2. 

It should be noted that because of these design changes the api version actually matters now when looking up the opcodes, an action that is required by several of the functions soot uses in dexlib2. He actually updated the mapping between the api level and the opcodes available at that api level so that they return the proper opcodes instead of the same ones for every api level after 20. 

This means for instances where you cannot infer the api level from say the manifest of an apk (i.e. when loading a .dex or .jar containing .dex), the api must be supplied manually. Therefore, a soot option needs to be created to supply the api level to dexlib2. 

Note I tried to mark the instances where I found the api level being required by dexlib2 with todo's when I encountered them but I may have not caught all instances.

agorski3:
This latest set of commits fixes a bunch of resource leaks involving open archive files. The archive files would remain open in many cases after soot had exited or been reset because the archive files were being opened, a stream was grabbed for an file in the archive, and then only the stream was closed but the archive file remained open. The was often because the stream was being treated as a generic InputStream to a file when really files in zip need a bit of special wrapping to ensure everything is closed correctly.

This occurred in two places. The first was ClassProvider and ClassSource for various types of input classes. The second was the ZipFilePart, PathPart, Program, and Attribute classes. These classes are in the generated directory of Soot's source. I am not sure if more needs to be done to correctly integrate my changes into the generated code since the classes are generated but hopefully you can see what I changed and make the appropriate additions. 

I also fixed a few minor bugs I saw when I was looking for the code leaving the archive files open.

StevenArzt:
I'm working on this right now. First of all, let me thank you for all the work you put into this. I will address the issue with the hard-coded API version. Since Soot already automatically selects the right platform JAR file based on a set of rules, we already know a suitable API version, we just need to pass it to the right place in the code and supply it to dexlib. If the automatic decision fails or just gets it wrong, the user can already manually specify an explicit platform JAR file to use. If this option is chosen, we might be able to parse the API version out of the file path if it's taken from the normal SDK directory. For cases in which that fails, we should have another option. That essentials leads to the following process for chosing the platform JAR and the API version:

1) User specifies android-jars and nothing else: Soot chooses automatically. If it fails, we just use the default (15 at the moment, I guess).
2) User specifies android-jars and the API version (new option): Soot picks the platform JAR that corresponds to the given API version. If it doesn't exist, we throw an exception.
3) User specifies force-android-jar and nothing else: We take that platform JAR and try to parse the API version out of it. If that fails, we take the default.
4) User specifies force-android-jatr and the API version (new option): We take the specified platform JAR file and the specified API version

That should give us all the freedom we need while still making it not too complex for the users, because the automatic choices should mostly be fine.

The issue of streams not being properly closed is a long-standing one. I have already fixed several code positions at which this happened on my own and I am not surprised that you found more, especially in the older parts of the code.

To make things bad, some of your changes are in the generated/ folder. I will look into where the respective code comes from and make sure that it's generated the correct way now, because just changing the generated code will lead to your changes being silently overwritten the next time the ant build script runs. I strongly feel that we need to make a cut at some point (call it Soot 2.0 or whatever you like) and get rit of all the legacy stuff. Let's face it: JastAdd can only cope with Java source up to Java 1.5 which makes it pretty much useless nowadays, but it's a pain in the butt to maintain that legacy stuff.

Let's see if more crops up while I'm working on this.

StevenArzt:
Manual merge.

agorski3:
Thanks. Double checking your changes right now since parts of the merge were manual. Did you already take care of the generated code or is there something more you have to do at a later time to make it so the code gets properly generated from the build script?

agorski3:
Just fyi. The reason I did not make any changes to JavaClassProvider is because of this check here:

```
if( file.file == null ) {
    throw new JarException(className);
}
```

Which basically prevents the loading of java files in archives for whatever reason. However, I guess it is still possible for the class to be passed a FoundFile with an open archive. Better to be safe than sorry I suppose.

StevenArzt:
I changed the inputs for the code generation, but could not test it yet. As I mentioned, this part is based on JastAddJ 1, which no longer exists officially. I modified the local copy we have on our build system, and will run it there to see whether I broke things as soon as the machine has finished its other builds.

agorski3:
I think it would be better to have DexPrinter and DexClassProvider read the api version from Options.v().android_api_version() instead of Scene. 

The only way to get the android api version returned by scene to be set to that in options is to call defaultAndroidClassPath(). However, suppose I am not reading in a apk but am instead reading in something in /system/framework. None of these require the android.jar that defaultAndroidClassPath() requires or it throws an exception. Moreover, it is in fact incorrect to try to load them with the android.jar as they are not apps. So in this situation what we really need is the be able to set the option externally and have DexClassProvider or DexPrinter read from the options version of the api not the scene version.

agorski3:
> I changed the inputs for the code generation, but could not test it yet. As I mentioned, this part is based on JastAddJ 1, which no longer exists officially. I modified the local copy we have on our build system, and will run it there to see whether I broke things as soon as the machine has finished its other builds.

Understood thanks. I tried to be as generic as possible with my coding so hopefully everything is compatible with java 1.5 but I don't know for sure because that is such an old version of java now.

agorski3:
Nothing else stands out as being problematic other than what I mentioned about the api version and reading it from options not scene.

StevenArzt:
The build ran without issues on the internal build system, so I can make it available to the public download site soon.

The issue with Android API version from the Scene was actually a mistake. I planned to make getAndroidAPIVersion() return the value from the Soot options if we don't have a value that was set when constructing the default classpath. Directly accessing the options would not be enough, though, because when the default classpath is used, we need to make sure that the API version is in line with the platform JAR that we picked. Commit ba54b434cd785c3b58f6316c73e001b4c5eed759 should do the trick.

agorski3:
Ah ok that makes sense and looks good. Thanks.
Any thoughts on the jack and jill method name issue while I have still go you (I know it is getting late over there)? I thought of a couple of general solutions that might work and sent them in an email but they both have their downsides.

PR #655, opened 2016-10-20
Reduce the time it takes to load classes from archives

agorski3:


agorski3:
These changes reduced the time it took to run my tool by about 70%. My tools is heavily dependent on reading classes from archives (multiple times) so that is why there is such a large impact. The changes should still be significant for normal use cases though.

StevenArzt:
I merged the request, but I wonder whether there are really situations in which there is more than one open input stream for a given code file. Normally, this should never happen. If you want to look into this, we can further simplify and speed up the code by getting rid of the list of streams.

agorski3:
We could do that as I am pretty sure the getInputStream method is never called more than once on any given FoundFile. I was writing the code with the idea of what if at some point we wanted to use FoundFile's code properly (maybe in Soot's rewrite) so that we have some unique mapping between a class name and a FoundFile (so the FoundFile would get used multiple times). For instance, you may want to pass FoundFile to the Jimple parser instead of using the ZipFilePart. But these are just ideas we could always just simplify it for now if you wanted.

The real code improvement I saw was related to the code in method "doJDKBugWorkaround" which is now in FoundFile. All the archive entries input streams are simply read completely into a ByteArrayInputStream. This input stream is then returned by getInputStream. What this means is that there is no particular reason to keep the ZipFile open as the ZipEntry input stream has already been completely read into memory. The only reason I left this alone and did not modify the code so ZipFile was not closed immediately after reading the entire ZipEntry into a ByteArrayInputStream is because the code is called a workaround and may be removed at some point. I am not even sure if it is still needed or what the purpose of it was.

StevenArzt:
The workaround method is just badly named. Previous versions of the code tried to red the file completely in one large chunk, On some JVM / OS combinations, this truncated the stream, i.e., the read operation read less than the specified amount of bytes, and returned this lesser number. Since the code, however, did not expect such a behavior, it never checked whether this file was fully read and just carried on. The "workaround" was to read one chunk at at time until we have all the data. Personally, I wouldn't call it a bug if a JVM can't read megabytes at a time, but well, that's just naming. Anyway, I'll simplify the code a bit to make sure that we close streams early that we don't need later on.

StevenArzt:
Here's what I did: 6be4dfaa149869734159401e6c583fe0a2796633 I kept the list around, but at least close the stream from which we copied.

agorski3:
Ah good point. Looks good to me.

patricklam:
I should hope the jdk bug is gone by now. Used to be that when you
requested a ZipInputStream for a jar file, the jdk would eventually give
you all 0s as input. There is no reason to keep the ZipInputStream open.
Or, try getting rid of the ByteInputStream intermediary and see if that
works now.

pat

PR #657, opened 2016-10-22
Fix class name parsing in dexpler and add outer/inner class hierarchy construction to jimple parser

agorski3:
This pull request fixes three issues:

    1. When determining the outer class of a inner class, dexpler used several different regular
       expressions, some were equivalent and others were not. Moreover, because of the ways these
       were written, some would return the outer most class instead of the first outer class of
       an inner class. The changes in this commit make sure that given a string such as
       'com.package.path.OuterMostClassName$1$MiddleOuterClass$InnerClass', dexpler returns
       'com.package.path.OuterMostClassName$1$MiddleOuterClass' as the outer class.

    2. Lambda classes generated by the jack and jill compiler contain extremely abnormal class
       names that may include the '$' which is normally reserved for indicating the outer and
       inner class hierarchy for a specific class. As such, a '$' does not always indicate
       the separation of the name of an outer class from one of its inner classes. Luckily,
       all the jack and jill generated lambda class names begin with a '-' and are always the
       inner class of class containing the lambda operation. Moreover, generated lambda
       classes are always the last class in the outer/inner class hierarchy specification of
       a class name. Therefore, dexpler can use the first occurrence of '$-' as a special indicator
       that we are dealing with a jack and jill generated lambda class and also as a separator
       for the inner class name and the outer class name.

    3. Previously, the jimple parser completly omitted the setup of the inner/outer
       class hierarchy as it read in classes from jimple. This just adds a bit of code
       to the JimpleClassSource file that uses the class hierarchy laid out in a classes
       full name to determine if it has an outer class, what the outer class is, and
       to set the classes outer class. The name parsing procedure is the same as the
       parsing procedure used in dexpler (i.e. it handles the jack and jill generated
       lambda classes).

In terms of the changes in dexpler, I am pretty sure I caught all the places in dexpler that handled inner/outer classes. I am also fairly certain that the parsing procedure I added works in all cases since I ran it on all the classes in the system image of AOSP 7.0 and everything seemed to work fine. However, I did replace code using 'replaceAll' with code that searches for the last or first occurrences of string indicators and then splits the string based on the index of these indicators. In other words, the functionality is not exactly equal. However, for all the classes of the system image of AOSP 7.0, I never encountered a inner class whose name was more than something like `Lcom.package.path.OuterMostClassName$1$MiddleOuterClass$InnerClass;` (i.e. no class names inside of '<>', more class names after the ';', more that one ';', or more than one 'L') so I did not see much reason in complicating the procedure by using a 'replaceAll'.

In terms of the changes to the jimple parsing procedure, when constructing the outer/inner class hierarchy, I simply copied the procedure used in dexpler and asm. Therefore, except for the fact that there are not any tags being set for inner classes or enclosing methods, everything should be fine. I did not set the tags because I was not entirely sure how and I was also not sure if they were really needed.

### DIFF
--- a/src/soot/JimpleClassSource.java
+++ b/src/soot/JimpleClassSource.java
@@ -46,22 +46,45 @@ public class JimpleClassSource extends ClassSource {
         
         InputStream classFile = null;
         try {
+        	//Parse jimple file
         	classFile = foundFile.inputStream();
         	soot.jimple.parser.JimpleAST jimpAST = new soot.jimple.parser.JimpleAST(classFile);
         	jimpAST.getSkeleton(sc);
+        	
+        	//Set method source for all methods
         	JimpleMethodSource mtdSrc = new JimpleMethodSource(jimpAST);
-
         	Iterator<SootMethod> mtdIt = sc.methodIterator();
         	while(mtdIt.hasNext()) {
         		SootMethod sm = mtdIt.next();
         		sm.setSource(mtdSrc);
         	}
+        	
+        	//set outer class if not set (which it should not be) and class name contains outer class indicator
+        	String outerClassName = null;
+        	String className = sc.getName();
+        	if(!sc.hasOuterClass() && className.contains("$")) {
+        		if(className.contains("$-")) {
+        			/* This is a special case for generated lambda classes of jack and jill compiler.
+					 * Generated lambda classes may contain '$' which do not indicate an inner/outer 
+					 * class separator if the '$' occurs after a inner class with a name starting with
+					 * '-'. Thus we search for '$-' and anything after it including '-' is the inner
+					 * classes name and anything before it is the outer classes name.
+					 */
+        			outerClassName = className.substring(0, className.indexOf("$-"));
+        		} else {
+        			outerClassName = className.substring(0, className.lastIndexOf('$'));
+        		}
+        		sc.setOuterClass(SootResolver.v().makeClassRef(outerClassName));
+        	}
 
+        	//Construct the type dependencies of the class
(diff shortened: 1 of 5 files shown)


StevenArzt:
I'll just merge this and run our test suite on it. In general, I feel that we need to centralize all this parsing stuff, especially because the special-casing for the Jack and Jill compilers is a mere heuristic. They might change their compiler at any time and then we need to adapt our code again. It would be better to only have one place to change.

agorski3:
I agree. It makes things a lot easier to change if the inner class parsing procedure is centralized. To do that though, you may want to switch around the ordering of things such as the performing the removal of "L" and ";" first in dexpler and performing the "/" to "." substitution first in dexpler. That way whatever function you write gets a consistently formatted string as input that uses soot's class naming procedure.

You should also be aware that while all of the locations require the other class name, a few require the short version of the inner class name as well (i.e. just the class name without the package or outer classes). 

Other than these changes, creating a centralized parsing method should be fairly simple.

StevenArzt:
@agorski3 I now had the requirement to touch that code again, so I centralized it in a minimalistic fashion by just refactoring those parts that deal with InnerClassTag objects into a static method in a separate helper class. All these places share the same semantics as to short/long names, so that was fairly simple. Semantically, I changed the code to skip over inner class annotations that have a null reference to the outer class and no clue in their name as to where they could belong. Throwing an exception is a bad option, it's better to keep the show running as well as you can, especially if you're dealing with malware or obfuscated APKs that contain all kinds of weird stuff that doesn't exactly follow the specification.

agorski3:
That all looks good.

PR #659, opened 2016-10-26
Supports AnySubType for reaching types

jtoman:
Our previous implementation only worked on concrete reference types.
Under Spark this was a MOSTLY a fine assumption, except that the native
simulators handlers could still introduce anysubtype types into the PAG.

This commit correctly handles this and, in principle, opens the door for types-for-invoke to be used with the CHA builder. However, the machinery to plug into the CHA is not written (yet).

### DIFF
--- a/src/main/java/soot/jimple/toolkits/callgraph/OnFlyCallGraphBuilder.java
+++ b/src/main/java/soot/jimple/toolkits/callgraph/OnFlyCallGraphBuilder.java
@@ -31,6 +31,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -38,6 +39,7 @@ import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import soot.AnySubType;
 import soot.ArrayType;
 import soot.Body;
 import soot.BooleanType;
@@ -325,6 +327,39 @@ public final class OnFlyCallGraphBuilder {
       }
     }
   }
+  
+  private Set<Type> resolveToClasses(Set<Type> rawTypes) {
+    Set<Type> toReturn = new HashSet<Type>();
+    for (Type ty : rawTypes) {
+      if (ty instanceof AnySubType) {
+        AnySubType anySubType = (AnySubType) ty;
+        RefType base = anySubType.getBase();
+        Set<SootClass> classRoots;
+        if (base.getSootClass().isInterface()) {
+          classRoots = fh.getAllImplementersOfInterface(base.getSootClass());
+        } else {
+          classRoots = Collections.singleton(base.getSootClass());
+        }
+        toReturn.addAll(getTransitiveSubClasses(classRoots));
+      } else if (ty instanceof ArrayType || ty instanceof RefType) {
+        toReturn.add(ty);
+      }
+    }
+    return toReturn;
+  }
+
+  private Collection<Type> getTransitiveSubClasses(Set<SootClass> classRoots) {
+    LinkedList<SootClass> worklist = new LinkedList<>(classRoots);
+    Set<Type> resolved = new HashSet<>();
+    while (!worklist.isEmpty()) {
+      SootClass cls = worklist.removeFirst();
+      if (!resolved.add(cls.getType())) {
+        continue;
+      }
+      worklist.addAll(fh.getSubclassesOf(cls));
+    }
+    return resolved;
+  }
 
   private void resolveInvoke(Collection<InvokeCallSite> list) {
     for (InvokeCallSite ics : list) {
@@ -343,7 +378,7 @@ public final class OnFlyCallGraphBuilder {
       // yet, then generate nullary methods
       if (mustBeNull || (ics.nullnessCode() == InvokeCallSite.MAY_BE_NULL
           && (!invokeArgsToSize.containsKey(ics.argArray()) || !reachingArgTypes.containsKey(ics.argArray())))) {
-        for (Type bType : s) {
+        for (Type bType : resolveToClasses(s)) {
           assert bType instanceof RefType;
           SootClass baseClass = ((RefType) bType).getSootClass();
           assert !baseClass.isInterface();
@@ -371,7 +406,7 @@ public final class OnFlyCallGraphBuilder {
         }
         assert reachingTypes != null && invokeArgsToSize.containsKey(ics.argArray());
         BitSet methodSizes = invokeArgsToSize.get(ics.argArray());
-        for (Type bType : s) {
+        for (Type bType : resolveToClasses(s)) {
           assert bType instanceof RefLikeType;
           // we do not handle static methods or array reflection
           if (bType instanceof NullType || bType instanceof ArrayType) {
@@ -393,7 +428,7 @@ public final class OnFlyCallGraphBuilder {
 
   private void resolveStaticTypes(Set<Type> s, InvokeCallSite ics) {
     ArrayTypes at = ics.reachingTypes();
-    for (Type bType : s) {
+    for (Type bType : resolveToClasses(s)) {
       SootClass baseClass = ((RefType) bType).getSootClass();
       Iterator<SootMethod> mIt = getPublicMethodIterator(baseClass, at);
       while (mIt.hasNext()) {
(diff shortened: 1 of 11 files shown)


mbenz89:

ericbodden:
Hi all. @jtoman usually one access all subtypes of a SootClass using Scene.v().getFastHierarchy(). In this particular case, however, I am not sure that the FastHierarchy is going to be in the correct state.  Would you mind checking this out?

jtoman:
Hi @ericbodden. My understanding is that AnySubType can be used with interface types, in which case we would have to first find all sub-interfaces of that interface and then find the concrete implementers of THOSE interfaces. In other words, given a type T (interface of class), finding all concrete classes that are subtypes of T requires different strategies depending on whether T is a class or interface.

As for the fast hierarchy, I _think_ it is available during call graph construction: the on-the-fly builder uses it transitively via the VirtualResolver (which uses getOrMakeFastHierarchy() _not_ makeFastHierarchy).

ericbodden:
You are right but I think that the Hierarchy or FastHierarchy already provide that functionality, also for interfaces.

Cheers
Eric

--
Sent from my mobile
________________________________

ericbodden:
@jtoman I am still considering merging this... did you find some time to look into using FastHierarchy?

jtoman:
Hello!

Yes I did, but my distro's Java package was so out of date I was unable to build against the latest master. I recently upgraded, so this is a perfect time to revisit this merge.

Last I checked, I do not believe the FastHierarchy provides a method that returns all concrete implementers of an interface, and all subinterfaces, but not both. The way AnySubType(T) is handled in the call graph builder, we need all the concrete classes that are subtypes of T. For interfaces, this means finding all subclasses of all implementers of T (and it's subiterfaces). Finding all implementers of T (and it's subinterfaces) is provided by FH's `getAllImplementersOfInterface`, but that method's comment says it does not return subclasses. Further, I do not believe there is anything in FH that provides  "find all transitive subclasses of class C." `getSubclassesOf` says it only returns _direct_ subclasses, but that comment may be out of date.

That said, rereading the code in this PR, the iterator approach is an awful nightmare. Last time I revisited this PR I significantly simplified it, and now that I can hopefully build Soot, I will push up this updated version.

jtoman:
Hello. I've pushed the simplified version. It is _not_ ready for merge yet; I have pushed the updated version for comments and feedback. (The maven build is still failing and I haven't been able to get a unit test set up yet.)

@mbenz89 I noticed you wrote some unit tests for the types for invoke features (thanks by the way), is there a way to include test classes without manually constructing the jimple representation?

mbenz89:
@jtoman Maybe I can help if you propose me with Maven's error message. Our build server seems to be fine with your changes.

Unfortunately, there is currently no mechanism to generate jimple classes for test cases together with the tests themselves. I've done this in another project and might feed it back into Soot but I haven't found the time for that yet.
The test cases I've written for the types for invoke feature are very sparse and only tackle the case where a reflective call is executed on a string literal as a receiver. Since the case is so simple, I've generated the jimple target programmatically.

jtoman:
@mbenz89 Here is the maven debug output when maven fails: https://gist.github.com/jtoman/94f629e83cda1eb5d0cd764e9b643f1c

Meanwhile, I'm working on a retest using the intermediate files class files produced by maven.

jtoman:
Sorry, I figured it out; apparently my local copy of polyglot was corrupted. Manually deleting maven's local copy and forcing a redownload did the trick.

jtoman:
Hello all,

I apologize for the delay, I have done some testing, and this patch now works. [1] I've confirmed that when an invoke() is performed with a receiver that is an AnySubType(T) where T is an interface or a class, the resolution mechanism dispatches to the appropriate methods. The test itself is rather involved and requires 9 classes to fully test the functionality. Again, any advice on how to make a unit test would be appreciated, but maybe that's too much to hope for now?

(I also tested the `getSubclassesOf` method of the FastHierarchy and the comment is correct, it does not return transitive subclasses. I updated the `resolveToClasses` method to a worklist approach to reflect this.)

[1] Modulo an unrelated bug I discovered in the types-for-invoke resolution. Apparently the dot() method of an AllocNode does not on-demand create DotFields, they have to be explicitly created through the pag. I have not included the fix in this PR, as it seems orthogonal (but it is a one line change and I can easily add it if need be).

mbenz89:
add to whitelist

mbenz89:
Could you propose a separate pr for the bug fix you did in `dot()`?

Unfortunately, Soot is missing a well-designed testing framework. There are some hacky possibilities right now, though. E.g., you could add the testing targets (jimple or class) as resources and when setting up your unit test, copy them to a temporary folder which you then analyze with Soot. 
It's not a nice solution, but it should work.

mbenz89:
I am just preparing the integration of my small integration/system testing framework for Soot. I'll might be able to still merge it today. It would be cool if you could try to implement your tests with it, so we can already see if its useful for the general purpose as well.

jtoman:
@mbenz89 #971 has the `dot` fix, I'm hoping to fold testing for that PR into the tests I write for this one.

jtoman:
Hello @mbenz89 I have checked out your branch, but I'm getting what appear to be unrelated test failures in the default tests (not the system tests). Namely:
```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   EnumTest>AbstractASMBackendTest.runTestAndCompareOutput:131 Expected line private final static synthetic [Lsoot/asm/backend/targets/MyEnum; $VALUES, but got private final static synthetic [Lsoot/asm/backend/targets/MyEnum; ENUM$VALUES in line 16 for class soot.asm.backend.targets.MyEnum
[ERROR]   ExtendedArithmeticLibTest>AbstractASMBackendTest.runTestAndCompareOutput:131 Expected line L0, but got GOTO L0 in line 103 for class soot.asm.backend.targets.ExtendedArithmeticLib
[ERROR]   InstanceOfCastsTest>AbstractASMBackendTest.runTestAndCompareOutput:131 Expected line CHECKCAST [Lsoot/asm/backend/targets/Measurable;, but got ARETURN in line 22 for class soot.asm.backend.targets.InstanceOfCasts
[ERROR]   ModifiersTest>AbstractASMBackendTest.runTestAndCompareOutput:131 Expected line // access flags 0x902, but got // access flags 0x102 in line 60 for class soot.asm.backend.targets.Modifiers
[ERROR]   MonitorTest>AbstractASMBackendTest.runTestAndCompareOutput:131 Expected line TRYCATCHBLOCK L0 L1 L2 java/lang/Throwable, but got ALOAD 0 in line 20 for class soot.asm.backend.targets.Monitor
[ERROR]   StoresTest>AbstractASMBackendTest.runTestAndCompareOutput:131 Expected line INVOKESPECIAL java/lang/StringBuilder.<init> ()V, but got LDC 2343249 in line 40 for class soot.asm.backend.targets.Stores
[ERROR]   TryCatchTest>AbstractASMBackendTest.runTestAndCompareOutput:131 Expected line TRYCATCHBLOCK L0 L1 L3 java/lang/Throwable, but got TRYCATCHBLOCK L0 L3 L4 java/lang/Throwable in line 18 for class soot.asm.backend.targets.TryCatch
```

I can hack up a test regardless, but I would also like to hear what your envisioned folder structure should be.

mbenz89:
I just checked out a fresh copy of the branch on another machine and `mvn clean test` runs just fine. The failing tests on your side check for equality of expected input with given input based on file positions. Might it be that you reformatted or changed some of the test files? This would break the tests. 
Generally, you could also just ignore the failing tests if you cannot figure out the reason as long as you do not push your changes. 

Regarding the folder structure, I do not quite get the question. As I mentioned, please put java code you want to analyze into `src/systemTest/targets` and the test cases (JUnit tests) themselves into `src/systemTest/java`. I would put both under the package in which the Soot component you want to test is located.

Hope that helps.

jtoman:
Hi Manuel,

Yeah I was mostly asking about the package structure, poorly worded question. I will look into these failing tests and hopefully have something soon.

jtoman:
Well, that merge isn't ideal, but unfortunately I can't think of a better way to get this code up without merging in feature/systemTest. Once #970 and #971 are merged I'll untangle this commit and it will be ready to merge. In the meantime, marking as WIP.

jtoman:
(The tests I wrote are currently failing because of the bug fixed by #971)

mbenz89:
Hi John, 
I just merged the stated pull requests. Thanks for your contribution!
There was a minor change in the API of the `AbstractTestingFramework` class. In your code, you can just remove the parameter of `setupSoot` and it should work fine.

jtoman:
Hi Manuel,

I've pushed a squashed version of the changes of this PR (I don't want to clog up soot with 5 different commits) and we should be good to go. Let me know if something needs still doing here.

mbenz89:
Thanks for your efforts!

PR #663, opened 2016-11-05
Update `dexlib2` version in `pom.xml` to 2.2b3.

cfallin:
Previously, a `mvn compile` in a clean Soot clone on the `develop` branch would not succeed, because of API changes in recent versions of `dexlib2` leveraged by recent Soot commits. This change updates the version of `dexlib2` pulled in by the Maven POM file to the one expected by the code.

---
This change is [Reviewable](https://reviewable.io/reviews/sable/soot/663)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
     <dependency>
       <groupId>org.smali</groupId>
       <artifactId>dexlib2</artifactId>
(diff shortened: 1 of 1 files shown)


StevenArzt:
I'll manually change this, because we're already at 2.2b4 now - with yet another change in the interface.

PR #681, opened 2016-12-09
Fixes #385

### LINKED ISSUE #385: JVM VerifyError after transformation from .class via shimple
When running soot via Shimple with compiled class files as source, some try-catch blocks are not processed correctly. Specifically, when a variable is assigned inside the try block (before a statement that the ThrowAnalysis directs to the catch block) and used after the try-catch, the output class file fails the JVM bytecode verification with an error such as:

Exception in thread "main" java.lang.VerifyError: (class: VerifyErrorTest, method: readFromConsole signature: ()Ljava/lang/String;) Accessing value from uninitialized register 1
        at java.lang.Class.getDeclaredMethods0(Native Method)
        at java.lang.Class.privateGetDeclaredMethods(Unknown Source)
        at java.lang.Class.getMethod0(Unknown Source)
        at java.lang.Class.getMethod(Unknown Source)
        at sun.launcher.LauncherHelper.getMainMethod(Unknown Source)
        at sun.launcher.LauncherHelper.checkAndLoadMain(Unknown Source)

However, this does not seem to happen when soot is given the java source as input.

A simple test case is available here: https://www.dropbox.com/sh/gvur3vfkykkqd7i/AABAkdu4AUtyku6IVg3qpmOTa?dl=0

You will see the error when trying to run the output given by the following:

```
C:\Soot_VerifyError> java soot.Main -via-shimple -pp -cp class VerifyErrorTest
```

However, the output given by the following passes JVM verification:

```
C:\Soot_VerifyError> java soot.Main -via-shimple -pp -cp src VerifyErrorTest
```

In that link you will also find an image of the ExceptionalBlockGraph and the jimple that are given when VerifyErrorTest.class is used as input to soot. On examining the jimple file, you can see the statement on line 34 is what causes the VerifyError because r0_2 may not be initialized due to the exception handler covering label1 to label2.

Waldo1637:


mbenz89:

ericbodden:
This part of Soot is super complex, so I cannot fully verify this but the fix does seem to make sense to me and I think we should accept it. @Waldo1637 sorry for taking so long. Unfortunately the code bases have diverged in the meantime. Would you mind updating the pull request so that it can be merged?

mbenz89:
Thanks for your contribution!

PR #701, opened 2017-02-28
Better lambda support via modeling of LambdaMetaFactory

balliet-gt:
This patch implements "modeling" of LambdaMetaFactory by creating a real implementation of the functionality of LambdaMetaFactory (by genenerating Jimple from the data in the invokedynamic callsite).

We create a new 'thunk' class containing two methods:

static bootstrap$(...) which just creates a new instance of the thunk class (this could just be a direct call to the constructor, the static method exists mostly for impedance matching with the invokedynamic callsite.

The actual function required by the functional interface, which is invokes the real method implementing the body.

This essentially implements lambdas like new Lambda implements FunctionalInterface () { public void foo() { ... body ... } }. This is sort of like what RetroLambda does.

This pull request isn't quite complete. I still have some regression tests which aren't quite properly integrated. I might also do some cleanup of the AsmMethodSource interface to LambdaMetaFactory. I think the meat of the implementation is in LambdaMetaFactory.java, which is included.

I suspect our discussions will lead me to write up some more informative comments for the functionality in that file too, but I figured I'd get it out there as is to start the discussion.

### DIFF
--- a/src/main/java/soot/LambdaMetaFactory.java
+++ b/src/main/java/soot/LambdaMetaFactory.java
@@ -1,0 +1,298 @@
+/* Soot - a J*va Optimization Framework
+ * Copyright (C) 1997-2014 Raja Vallee-Rai and others
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+package soot;
+
+import soot.jimple.*;
+import soot.coffi.Util;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public final class LambdaMetaFactory {
+    public static SootMethodRef makeLambdaHelper(
+                        List<? extends Value> bootstrapArgs,
+                        int tag,
+                        String name,
+                        Type[] types) {
+        if(bootstrapArgs.size() < 3 || 
+           !(bootstrapArgs.get(0) instanceof ClassConstant) ||
+           !(bootstrapArgs.get(1) instanceof MethodHandle) ||
+           !(bootstrapArgs.get(2) instanceof ClassConstant) ||
+           (bootstrapArgs.size() > 3 && !(bootstrapArgs.get(3) instanceof IntConstant))) {
+            G.v().out.println("warning: LambdaMetaFactory: unexpected arguments for LambdaMetaFactor.metaFactory");
+            return null;
+        }
+        String samMethodType = ((ClassConstant) bootstrapArgs.get(0)).getValue();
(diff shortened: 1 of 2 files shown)


ericbodden:
Thanks a lot. This will require us a bit to double-check... Please bear with us.

mbenz89:

mbenz89:
Could you also upload your test suite? That would help me to get through this pr.

src/soot/LambdaMetaFactory.java
```diff
@@ -0,0 +1,272 @@
...
+                if(va >= bootstrapArgs.size()) {
+                    G.v().out.println("warning: LambdaMetaFactory: unexpected arguments for LambdaMetaFactory.altMetaFactory");
+                    return null;
+                }
+                Value v = bootstrapArgs.get(va++);
+                if(!(v instanceof ClassConstant)) {
+                    G.v().out.println("warning: LambdaMetaFactory: unexpected arguments for LambdaMetaFactory.altMetaFactory");
+                    return null;
+                }
+                bridges.add((ClassConstant)v);
```
mbenz89: You gather bridging types here but never use them. Is it planned to extend the factory by the corresponding bridge methods?

balliet-gt: Yes (support for altMetaFactory was a little hurried and mostly just wanted to get it to parse these without blowing up). Let me revisit this and add the bridge methods.

src/soot/Scene.java
```diff
@@ -686,7 +686,7 @@ private void modifyHierarchy() {
      * as a library class and invalidates the class hierarchy.
      * @param c The class to add
      */
-    public void addClass(SootClass c) 
+    public synchronized void addClass(SootClass c)
```
mbenz89: Do you remember why it was necessary to make these method synchronized? As far as I know, class-resolution is serial in Soot and, thus, this wasn't necessary before. I am confused why your PR should change this.

balliet-gt: Hmm... I certainly saw multiple threads running my code (without specifying an option, which I forget the name of, to limit it to one thread). It is possible something has changed since I originally did this (which was quite a while ago).

src/soot/LambdaMetaFactory.java
balliet-gt: b28f85d (now in the pull request, below) adds support for the bridge methods

src/soot/Scene.java
balliet-gt: I think this may be unnecessary now, at least my test case which triggered the problem before doesn't with current HEAD, and my changes. It is possible that some other bug was fixed it the mean time. We'll try it without this change.

StevenArzt:
ok to test

mbenz89:
Could you provide me with your test cases as well?

balliet-gt:
I've been testing with this test case mostly (at least as far as the details of "thunk" construction, etc).

https://github.com/GrammaTech/j8-tests/tree/master/src/apps/church

I considered adding this whole thing as a unit test somehow, but wasn't quite clear how this should fit in to the rest of the tests (it might've made since as part of the old 'systests').

I've also been running the whole Java 8 test suite we developed (although it is mostly "real world" sized apps, and probably a little heavy weight): https://github.com/GrammaTech/j8-tests/blob/master/Readme.md

ericbodden:
@mbenz89 Can we merge this change?

mbenz89:
I have to recheck this with the provided test cases and the current state. Last time I had a look, the bridging methods weren't implemented and the MetaFactory wasn't quite correctly wired to the ASMMethodSource.

mbenz89:
retest this please

mbenz89:
retest this please

StevenArzt:
@mbenz89 Are there any updates to this MR? Lamdas and invokedynamic calls are becoming more and more popular with newer versions of Java.

mbenz89:
@StevenArzt One of my student assistants is currently investigating the MR, I hope we can finish this up next week as well.

mbenz89:
Merged with #1078.

Thanks for your contribution!

PR #707, opened 2017-03-07
Remove nops before building Shimple or else Jimple validation may fail

Waldo1637:
Must remove nops prior to building the CFG because NopStmt appearing before the IdentityStmt in a trap handler that is itself protected by a trap cause Phi nodes to be inserted before the NopStmt and therefore before the IdentityStmt. This introduces a validation problem if the Phi nodes leave residual assignment statements after their removal.

Generalized NopEliminator to work on any type of Body by removing the cast of the body to JimpleBody


mbenz89:
Thanks for your contribution!

PR #711, opened 2017-03-20
Fix quoting and unquoting issues for .jimple format

jonmathews-ensoft:
Escaping subparts of names introduced in
9a00dc779dbb1c2b29dc40dfa47604c66db644b9

Other related commits:
9a00dc779dbb1c2b29dc40dfa47604c66db644b9
8275282dec8c3dbc72307f25469e75ff4ddfa9f6
5f46195dcb3b5381c92a443b1932fffd0cb57969
804bf1f58d81df7a9ff8a0947815339fdc901b09
5bd57e5aa3734def55d592e18386cb5d6c196ed4
9ab0dc724d7e7f90dc1bf4571e297bc3891ecf9b

### DIFF
--- a/src/soot/ArrayType.java
+++ b/src/soot/ArrayType.java
@@ -131,6 +131,20 @@ public class ArrayType extends RefLikeType
 
         return buffer.toString();
     }
+    
+    /** Returns a textual representation, quoted as needed, of this type for serialization, e.g. to .jimple format */
+    @Override
+    public String toQuotedString() {
+        StringBuilder buffer = new StringBuilder();
+
+        buffer.append(baseType.toQuotedString());
+
+        for(int i = 0; i < numDimensions; i++)
+            buffer.append("[]");
+
+        return buffer.toString();
+    }
+
 
     public int hashCode()
     {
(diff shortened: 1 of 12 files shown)


jonmathews-ensoft:
This patch undoes / fixes various changes made to quoting when Jimple text format is involved.  The goal is to ensure that quotes only exist in the text format.  Prior to this patch, soot was inconsistently quoting and unquoting, which could cause unresolved references if either performing a round-trip through Jimple text, or loading from both class and Jimple text.

[jimple-quoting.zip](https://github.com/Sable/soot/files/856248/jimple-quoting.zip)

See attached project with legal Java source code that triggers the issue when converted into Jimple.

* To reproduce issues
 * compile the Java into class files with javac
 * load class files with Soot, write Jimple text format
 * load Jimple text format with Soot, write Jimple text to another location
 * compare both versions of Jimple text output

Prior to this fix Soot either errors out or the two copies of Jimple would not match (assuming local variables have the same names).

mbenz89:
The quoting was indeed rather inconsistent before the proposed fix and led to the described errors.
Thanks for your contribution!

PR #745, opened 2017-06-06
new option: ignore-dex-loading-errors

jpstotz:
If Soot fails to load one dex file from an APK file that contains multiple dex files the overall loading process fails. 

If ignore-dex-loading-errors is set to true those dex files that can not be loaded are now ignored so that at least the classes from the other dex files can be used via Soot.

### DIFF
--- a/doc/soot_options.htm
+++ b/doc/soot_options.htm
@@ -684,6 +684,11 @@ packages in the Java runtime.</p>
 </td><td colspan="2">Process all DEX files found in APK.</td>
 </tr>
 <tr>
+<td><tt>-ignore-dex-loading-errors </tt>
+<br>
+</td><td colspan="2">Ignore errors when loading DEX files found in APK/ZIP</td>
+</tr>
+<tr>
 <td><tt>-process-path <var>dir</var></tt>
 <br>
 <tt>-process-dir <var>dir</var></tt>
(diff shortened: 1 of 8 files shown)


mbenz89:
Thanks for your contribution!
Can you provide an example app where the error occurs and your option would help to solve it?

jpstotz:
According to my logs the following apps contain(ed) invalid dex files:

* com.arteoning.chestsimulatorforcr (`lib/armeabi/mixz.dex`)
* com.arteoning.fut18draftsimulator (`lib/armeabi/mixz.dex`)
* com.zijayrate.fidgetspinner (`lib/armeabi/mixz.dex`)
* com.candidate.chestsimulatorforcr (`lib/armeabi/mixz.dex`)

mbenz89:
I encapsulated the dex loading process and replaced all accesses to dex files by the new process. This seems to eliminate the problem on my side.
Can you try analyzing the problematic apks again with the newest soot version and tell me if the problem still remains on your side?

jpstotz:
I will look into your changes. However I already noticed one potential problem:

Your dex index maps a class name to an implementation. I already have encountered APK files with multiple dex files where one class is contained in more than one dex file. 

By ` index.put(className, dexSource);` in DexClasProvider.java you replace existing classes.
At least the should be a warning.

In my code for loading dex files (non-soot) I had a similar algorithm This ended up that an obfuscated class from classes.dex was replaced by another obfuscated class that had a totally different interface...

From my experience it is better to sort the dex files based on their priority (classes?.dex first) and then keep the first version of a class. This makes it impossible that a minor dex file found somewhere in the APK overwrites a class from the main dex files.

mbenz89:
Good point, I'll take your suggestion.

mbenz89:

mbenz89:
Since the problem does not seem to persist with the updated newly introduced dex loading process, I close this now.

Issue #772, opened 2017-09-05
Does soot suport multidex?


jacoryjin:
Does soot suport multidex?  An apk has class.dex, class1.dex, raise warn with not found activity

XingweiLin:
It seems `Soot` has not supported multidex now. #565

mbenz89:
Have you tried the `-process-multiple-dex` option?
Soot should be able to handle APKs with multiple dex files with that option. It should still have problems with writing out, though.

jacoryjin:
@XingweiLin ，@mbenz89，Thank you for the answers！
`Options.v().set_process_multiple_dex(true); `  doesn`t work in my code；
I uses another way：
1、translate .apk to .jar with [enjarify](https://github.com/google/enjarify);
2、soot-infoflow-android->SetupApplication.java->initializeSoot(), changed:
`Options.v().set_process_dir(Collections.singletonList(apkFileLocation));
Options.v().set_src_prec(Options.src_prec_apk_class_jimple);`
To
`Options.v().set_process_dir(Collections.singletonList("E:/app-release.jar"));
Options.v().set_src_prec(Options.src_prec_java);`

mbenz89:
Have you tried setting `Options.v().set_process_multiple_dex(true);` inside of soot-infoflow-android->SetupApplication.java->initializeSoot()?
The Soot options are overwritten there, thus it doesn't change anything if the options are altered before any calls to `SetupApplication`.

Btw. you do not need to change the source code of `SetupApplication`. Instead, you can override the `SootConfigForAndroid` class and pass it to `SetupApplication`:

```
SetupApplication setupApplication = new SetupApplication(jarPath,apkPath);
        
SootConfigForAndroid sootConf = new SootConfigForAndroid() {
            @Override
            public void setSootOptions(Options options) {
                // we need to specify soot options here since FlowDroid resets them
                super.setSootOptions(options);
                options.set_process_multiple_dex(true);
            }
};
setupApplication.setSootConfig(sootConf);

ShengliangD:
It took me quite a while to find this solution :(

mbenz89:
It's always a good idea to look at Soot's options to see what is possible and what might not be.

Unfortunately, we do not have the manpower to maintain Soot's documentation very well. We would thus be happy if you could contribute and maybe add a wiki page on how to use Soot with android.

ShengliangD:
[...]

PR #785, opened 2017-09-29
Added support for multi Dex write-out

mbenz89:
- Updated dexlib2 library
- Refactored DexLibWrapper to list all Dex files and warn the user if only processing one
  - Fixed a bug where all classes were mapped to a single Dex file even if processing in multi dex mode
- Added write-out of multiple dex files if the content is too large for a single one.
  - Only implemented for Android versions post Lollipop. An exception is thrown for previous versions

To allow dex splitting for pre Lollipop versions, it would be necessary to compute the needed classes of the main dex file. This isn't necessary for post Lollipop versions since Art will extract and compile all classes out of all dex files anyway.

Maybe you want to have a look at this before merging @StevenArzt.


mbenz89:
This is also partly related to #772.

mbenz89:
I found a few apps for which the new write-out process does corrupt the apps. I'll investigate this more and reopen the request when fixed.

mbenz89:
The mentioned issue was caused by some unrelated error and not by the dex splitting.
I think the splitter is ready to be merged.

StevenArzt:
Looks fine to me, thanks. It's clear that the generated code will not run on older versions of Android (dalvik runtime), but that's a fair design choice to make.

PR #792, opened 2017-10-11
Added java.lang.Exception to basic classes

anddann:
In the set of basic classes the class java.lang.Exception is missing.
This leads to the fact that the resolution `superClassPath` of a `java.lang.RuntimeException` stops at Exception, and thus `java.lang.Object` is not contained in the hierachy in method `BytecodeHierarchy.public static RefType lcsc(RefType a, RefType b, RefType anchor)`.

### DIFF
--- a/src/soot/Scene.java
+++ b/src/soot/Scene.java
@@ -1437,6 +1437,7 @@ public class Scene // extends AbstractHost
 		addBasicClass("java.lang.Error");
 		addBasicClass("java.lang.AssertionError", SootClass.SIGNATURES);
 		addBasicClass("java.lang.Throwable", SootClass.SIGNATURES);
+		addBasicClass("java.lang.Exception", SootClass.SIGNATURES);
 		addBasicClass("java.lang.NoClassDefFoundError", SootClass.SIGNATURES);
 		addBasicClass("java.lang.ExceptionInInitializerError");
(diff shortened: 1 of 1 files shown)


anddann:
This problem occurs if the defaults excludes are activated, which include classes in java.*.
Thus, the same problem showed up for `java.lang.ReflectiveOperationException`

mbenz89:
`java.lang.ReflectiveOperationException` was introduced in Java 7. Adding it to the basic classes might break backward compatibility. Maybe we should insert some checks on the java versions.
Also, if we do so, it would be cool to have an analysis that extracts those classes with versions for us, so that we can get a complete list.

anddann:
Yeah, you are absolutely right is overlooked the fact that `ReflectiveOperationException` was introduced in Java 1.7.
Anyway, the class `java.lang.invoke.LambdaMetafactory` is contained in the set of basic classes and was added in Java 1.7, too.
However, this raises up the question "how do the basic classes differ for each java version?".
Do we need different basic classes for java 1.5,1.6,...,1.9, android 21,...android 25?
Do we know which java/android version we are analyzing?

mbenz89:
Good point! We should discuss further proceedings for such corner cases. 
Merging your change for `java.lang.Exception`.

PR #795, opened 2017-10-13
MethodHandle constants and invocation

akbertram:
This corrects the output of MethodHandle constants implemented in the previous pull request (#637) and adds support for MethodHandle.invoke() and MethodHandle.invokeExact(), which are signature polymorphic.

### DIFF
--- a/src/soot/SootMethodRefImpl.java
+++ b/src/soot/SootMethodRefImpl.java
@@ -164,10 +164,6 @@ class SootMethodRefImpl implements SootMethodRef {
 
 	private SootMethod tryResolve(StringBuffer trace) {
 
-		if(isSignaturePolymorphic()) {
-			return new SootMethod(name(), parameterTypes(), returnType(), Modifier.PUBLIC);
-		}
-
 		if (declaringClass.getName().equals("java.dyn.InvokeDynamic")) {
 			throw new IllegalStateException("Cannot resolve invokedynamic method references at compile time!");
 		}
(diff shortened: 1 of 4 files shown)


akbertram:
@StevenArzt Anything missing from this PR? Happy to make any changes or improvements as requested.

src/soot/SootMethodRefImpl.java
```diff
@@ -152,6 +163,11 @@ private SootMethod checkStatic(SootMethod ret) {
 	}
 
 	private SootMethod tryResolve(StringBuffer trace) {
+
+		if(isSignaturePolymorphic()) {
+			return new SootMethod(name(), parameterTypes(), returnType(), Modifier.PUBLIC);
```
StevenArzt: Please don't create SootMethods out of thin air. This method won't be registered in the Scene, will have no parent class, and will therefore violate various reasonable assumptions held by analysis clients.

akbertram:
I've now ensured that the new SootMethod gets added to the Scene graph. Is there any way to test that this works correctly. Right now I run Soot's *intra*procedural optimization on a classfile that uses signature polymorphism, but that was passing before this change.

mbenz89:

akbertram:
Hi @StevenArzt, do these changes address your concerns or is there anything else I need to fix?

ericbodden:
@StevenArzt could you please look into this again? Thanks!

mbenz89:
ok to test

mbenz89:
Signaturepolymorphic is now supported by #1135

PR #818, opened 2017-11-05
fix jbco MethodRenamer, FieldRenamer and CollectConstants

gnom7:
Fix `MethodRenamer` "Unresolved compilation error" in renamed method's bodies.
Refactoring `SootMethod#setName` and `SootClass` (remove `subSigToMethods` field and use `methodList` instead, no more "keep in sync" work between those two fields, e.g. to rename method just rename it, not remove-rename-add).

### DIFF
--- a/src/soot/jbco/jimpleTransformations/CollectConstants.java
+++ b/src/soot/jbco/jimpleTransformations/CollectConstants.java
@@ -97,7 +97,7 @@ public class CollectConstants extends SceneTransformer implements IJbcoTransform
 
 		for (SootClass cl : appClasses) {
 			for (SootMethod m : cl.getMethods()) {
-				if (!m.hasActiveBody() || m.getName().indexOf("<clinit>") >= 0)
+				if (!m.hasActiveBody() || m.getName().contains(SootMethod.staticInitializerName))
 					continue;
 
 				for (ValueBox useBox : m.getActiveBody().getUseBoxes()) {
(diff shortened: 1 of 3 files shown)


gnom7:
please comment

mbenz89:
I haven't had the time to go through everything yet, but there is an issue with your removal of `subSigToMethods`. SootClass#getMethod may be called a lot of times in client code and, thus, is performance critical. With your changes, the list of all methods will have to be traversed each time the method is called. Can you please change the behavior back to use a map? You may remove the list of methods and use Map.values(), though.

mbenz89:

gnom7:
Hello, @mbenz89, I agree with you in whole, but I just cannot expect much from such optimization (from both code quality and performance perspective), there should be better opportunities for that.

A few arguments from mine

I use soot in my project and with that change I'm really happy (as a client and as a contributor - I made this refactoring after I faced a bug in `MethodRenamer`), and I didn't note any performance drop - do we have any sort of benchmark to check whether it is really critical?

(Also is there anyone who uses soot in their project's runtime + response time sensitive? I think that soot is another sort of thing, and it is by nature cannot have performance, appropriate for direct runtime use. It is already runs for me for 10+-2 seconds, and I just cannot notice any change)

But anyway my main argument is that I'm sure code quality is more critical than performance, I really don't like that `SootClass` goes with state of `SootMethod`, rather than asks later for that state (bad practice, isn't it?) and your proposal doesn't changes that. (later we have bugs as with `MethodRenamer` - iteration among class methods and renaming them causes `ConcurrentModificationException`, doh! - and more, we should write messy code when working with methods to update class state too, that was nasty surprise for me so I made this refactoring and proposed it as PR).

mbenz89:
ok to test

gnom7:
@StevenArzt could you please merge this or comment what should be done?

StevenArzt:
I agree with @mbenz89 here. While there are definitely places in Soot where the code is both ugly and slow, there are also deliberate design decisions that trade off performance and cleanliness. There are many different use cases for Soot, ranging from analyzing or transforming smaller targets where the whole process takes only a couple of seconds, to large targets on which Soot runs for multiple minutes. You also have to take into account that clients may be complex analyses that run from anything between seconds and hours. Therefore, I am very reluctant to integrate this change. In fact, in some of our projects, we even keep own caches for some of Soot's data objects to avoid costly checks that need to loop over collections.

There are clean ways to fix method renaming. Classes that keep state can register as listeners, or can check a modification counter to see whether their cached state is still up-to-date.

ericbodden:
The optimizations/caches were put in place for a reason, not because people felt they had nothing better to do. Hence please make sure they stay intact.

Cheers
Eric

gnom7:
@StevenArzt Ok, I got it, thanks for your comment, I'll try to fix it without touching SootMethod and SootClass classes. Does soot has any sort of benchmarks?
@ericbodden While I didn't have in mind that people used it without a reason or whatever, I can disagree with this design decision and propose this PR. I hope you understand.

Thanks you all.

gnom7:
@StevenArzt please review new PR for this issue.

mbenz89:
ok to test

mbenz89:
retest this please

gnom7:
@mbenz89 
Hello, what exactly should I retest?

mbenz89:
This just instructs our ci to rebuild and test the pr. Sorry for any confusions caused.
The pr looks good to me, I'm gonna merge it. Thanks for your contribution!

gnom7:
Happy to hear :)
Thank you

PR #821, opened 2017-11-06
Natural mvn

mbenz89:
Refactoring of Soot's project structure to be closer to typcial Maven/Gradle standard


mbenz89:
ok to test

mbenz89:
@StevenArzt what do you think about this pr?
I think it's a good idea to get Soot a bit more up to date to current directory standards. Should be a bit easier for new contributors to get in. Separation of resources and source is a bit more consistent. As a nice bonus, the pom file gets a bit less bloated.

Alexeyan:
I would recommend to add <pluginManagement></pluginManagement> tags around the <plugins></plugins> tag in the pom.xml. 
This way projects that include soot inherit the plugin configuration by default, which seems like a sane default.
See https://maven.apache.org/pom.html#Plugin_Management

mbenz89:
You are right, I may indeed add this. 
I'm just asking my self if it is a realistic scenario that someone uses Soot's pom as parent pom. Can you think of such a scenario?

Alexeyan:
I came across this issue while using the development version of flowdroid 
(as in: https://github.com/secure-software-engineering/soot-infoflow and https://github.com/secure-software-engineering/soot-infoflow-android).
 
The Eclipse Projects depended on the soot project and adding the tag fixed an "Plugin execution not covered by lifecycle configuration"-error for me in the soot project.

StevenArzt:
Making Soot compliant with the Maven directory structure sounds good to me. Can you update the changes with the latest Soot code files to get around the merge conflicts?

FlowDroid will be a set of maven projects starting with the upcoming 2.5 release as well, which should make it a lot easier to build everything you need. FlowDroid will reference Soot as a maven dependency and everything should fit neatly into the maven model.

mbenz89:
Sure, I will merge back the changed files into this pr and also merge the pr into the develop branch when I have time for it.

mbenz89:
retest this please

PR #838, opened 2017-12-04
fixed scc calculation for headless graph

### LINKED ISSUE #836: StronglyConnectedComponentsFast missing sccs
`StronglyConnectedComponentsFast` does not seem to identify any sccs if the given `DirectedGraph` does not have any heads.

```
MutableDirectedGraph<String> graph = new HashMutableDirectedGraph<String>();
graph.addNode("foo");
graph.addNode("bar");
graph.addEdge("foo", "foo");
graph.addEdge("foo", "bar");
System.out.println(new StronglyConnectedComponentsFast<>(graph).getComponents());
```

This outputs an empty list: `[]`

I would have expected an output like `[[foo], [bar]]`.

stefandollase:
StronglyConnectedComponentsFast now runs over all nodes instead of only the possibly empty list of heads of the graph.

Fixes #836


anddann:
ok to test

mbenz89:
Looks good to me. Thanks for your contribution!

PR #847, opened 2017-12-15
MethodValidator

gnom7:
I have added validator to check that static initializers have 'static' modifier (if not exception with message "<clinit> is not flagged correctly wrt/ static" will be thrown in android runtime, see d46683feb30ec21ca913219da365e6b0df058dff commit)
and removed change added in 382971d0cb32f9ecf2c6d363e1626821d03ca05e commit to IdentityValidator , as we don't need it now since MethodValidator ensures that all static initializers have 'static' modifier (thus check for static is sufficient)


mbenz89:
ok to test

PR #850, opened 2017-12-18
Update java and some dependencies versions

gnom7:
Enable Java 1.8 source/target level (I hope adcd0b620f9cd46feb39592c3105bf184c07bb40 was about that, if not, I'll change it back to 1.7)
Update asm, ant, dexlib2 versions (without introducing new dependency conflicts)
Update maven-javadoc-plugin version
Move plugin versions to pom properties


mbenz89:
ok to test

mbenz89:
I am ok with the version changes you did and I also appreciate the extraction of plugin versions to the configuration section. Nevertheless, I using "java.version" does not seem to be the standard way of defining the java version in Maven. It seems to be a spring boot specific ([source](https://stackoverflow.com/questions/38882080/specifying-java-version-in-maven-differences-between-properties-and-compiler-p)). If you have problems with IntelliJ detecting the correct version, you should apply a "reimport from maven" action. I guess IntelliJ does just not recognize the target version has changed, for a new import it should actually work!

mbenz89:
If you change back "java.version" to ``` <maven.compiler.source>1.8</maven.compiler.source><maven.compiler.target>1.8</maven.compiler.target>``` I would be happy to merge this pr.

gnom7:
I have updated PR.
Also I have tried for the third time to compile project with only your adcd0b620f9cd46feb39592c3105bf184c07bb40 commit, and that time it worked (although I have used `mvn clean package` command, not IntelliJ, and I have double checked this earlier), sorry for that, I still have no idea where from was that problem.

PR #852, opened 2017-12-21
fix issue with MethodRenamer

gnom7:
when class already contains method-1 and field with the same name and we decide to rename another method-2 with same field's(method-1's) name, so getting for a while two different methods with same name (if method-2 renamed before method-1) in one class, causing Exception)

### DIFF
--- a/src/main/java/soot/jbco/jimpleTransformations/MethodRenamer.java
+++ b/src/main/java/soot/jbco/jimpleTransformations/MethodRenamer.java
@@ -20,11 +20,12 @@
 package soot.jbco.jimpleTransformations;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Vector;
+import java.util.stream.Stream;
+
 import soot.Body;
 import soot.FastHierarchy;
 import soot.G;
@@ -43,6 +44,8 @@ import soot.jbco.util.BodyBuilder;
 import soot.jbco.util.Rand;
 import soot.jimple.InvokeExpr;
 
+import static java.util.stream.Collectors.toList;
+
 /**
  * @author Michael Batchelder
  * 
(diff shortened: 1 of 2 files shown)


gnom7:
@mbenz89 may I ask you to merge this pr please?

src/main/java/soot/jbco/jimpleTransformations/MethodRenamer.java
```diff
@@ -214,51 +214,52 @@ public static String getNewName() {
...
+                .filter(SootClass::isLibraryClass)
+                .anyMatch(c -> c.declaresMethod(method.getName(), method.getParameterTypes(), method.getReturnType()));
+    }
+
+    private static List<SootClass> getAllInterfacesOf(SootClass sc) {
+        Stream<SootClass> superClassInterfaces = sc.isInterface() ? Stream.empty() : hierarchy.getSuperclassesOf(sc)
+                .stream()
+                .map(MethodRenamer::getAllInterfacesOf)
+                .flatMap(Collection::stream);
+        Stream<SootClass> directInterfaces = Stream.concat(sc.getInterfaces().stream(), sc.getInterfaces().stream()
```
mbenz89: You concat the same stream here, is this correct? If so, could you please explain the semantics behind it?

src/main/java/soot/jbco/jimpleTransformations/MethodRenamer.java
```diff
@@ -85,81 +85,80 @@ protected void internalTransform(String phaseName,
...
-									+ c.getName());
+									+ sc.getName());
+						}
 						rename = false;
 						break;
 					}
 				}
 
 				if (rename) {
-					// TODO: This is flawed since it all methods of a similar
```
mbenz89: Where you able to fix t this? Where?

src/main/java/soot/jbco/jimpleTransformations/MethodRenamer.java
gnom7: Please re-check this, second stream is not the same as first.

src/main/java/soot/jbco/jimpleTransformations/MethodRenamer.java
gnom7: I will provide commit for this in a while (after I check changes on my project and ensure that all changes work fine), for now I have returned this TODO in f31aba60e76f438973c61abcf85d78b2e33d5392

mbenz89:
OK to test

mbenz89:
Thanks for your contribution!

PR #854, opened 2017-12-23
Added support for dot visualization of icfg

kshitijgorde:
Added support to traverse an InterproceduralCFG from heros and save it to a file in DOT extension. Graph Traversal is done in depth-first manner!


kshitijgorde:
@ericbodden can verify this patch

kshitijgorde:
Could not add reviewers but @ericbodden should be able to verify this patch

PR #855, opened 2017-12-29
Resolved compilation errors due to formatting of soot_options.xml

rajatkhanna1994:


mbenz89:
OK to test

mbenz89:
Hi.

I'm a bit confused by this pr. It seems that you formatted the options XML file so that it can be compiled. That is not your task! The file is already formatted so that it correctly compilea. Your task is to modify the xslt file so that "soot_options. XML" can be arbitrarily formatted and the generated Java class is still compilable.

Is that clear?

Manuel

rajatkhanna1994:
Oh okay, sorry I misinterpreted the task.
It is clear now.

PR #868, opened 2018-01-14
Made changes to 'make-soot_options.xsl' so that options.java still ge…

rajatkhanna1994:
…ts compiled(without errors) when 'soot_options.xsl' is formatted in any way


mbenz89:
ok to test

PR #881, opened 2018-01-31
Adding support for output_jar flag

pnest:
* Made DexPrinter write to jar file when source APK is absent. To enable it set `Options.v().set_output_jar(true)`
* replaced manual closing streams with try with resources
* replaced writing to System out with slf4j logger

### DIFF
--- a/doc/soot_options.htm
+++ b/doc/soot_options.htm
@@ -1,5860 +1,7204 @@
 <html>
-<head>
-<META http-equiv="Content-Type" content="text/html; charset=UTF-8">
-<title>Soot Command Line Options</title>
-<META NAME="description" CONTENT="Soot command-line options">
-<META NAME="keywords" CONTENT="usage">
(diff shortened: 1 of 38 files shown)


mbenz89:
ok to test

mbenz89:
@pnest Would you mind adding some additional information on how the `set_output_jar` option changes output semantics when analyzing an apk to the soot_options.xml?

StevenArzt:
@pnest Any news on the merge request?

pnest:
Added usage notes to ``soot_options.xml`` and also updated ``.xls`` files to generate clearer ``.java`` classes - will definitely need this done when code formatting checker will be applied

mbenz89:
Thanks for your contribution! 
Btw. we are not planning to check the formatting of generated files. These should have a consistent formatting anyway.

StevenArzt:
Please do not reduce the visibility of fields or methods. If something is protected, there may be code that depends on being able to access the respective member through a subclass. I'll change the affected members back to protected.

Issue #874, opened 2018-01-19
Problems with whole program analysis and Java 8?


ChuckFoo:
While investigating some Soot call graph issues in some analysis code I'm working on, I updated to the latest & greatest Soot to pick up bugfixes, which resulted in me having to switch from using Java 7 (which I had been using previously after hitting issues w/ Java 8 which seemed to be known to the developers at the time) to Java 8 due to recent commits.  I assumed that this meant that the problems with using Java 8 to run the analysis had gone away, but apparently some still exist. :)

The first problem can be seen by doing whole program analysis on something like a recent JEdit installer (not what I was working with but a simple test case that shows the same issue w/ Soot standalone, jar with deps downloaded from Soot nightly builds 1/16/18), like so:

```
java -classpath ~/Downloads/sootclasses-trunk-jar-with-dependencies.jar soot.Main -w -p cg verbose:true -p cg.spark "enabled:true,verbose:true,pre-jimplify:true" -verbose -debug -soot-class-path ~/Downloads/jedit5.4.0install.jar:/usr/lib/jvm/java-1.8.0-openjdk-amd64/jre/lib/rt.jar:/usr/lib/jvm/java-1.8.0-openjdk-amd64/jre/lib/jce.jar -process-dir ~/Downloads/jedit5.4.0install.jar
```

This results in the following exception:

```
java.lang.RuntimeException: Failed to convert <java.util.stream.FindOps: java.util.stream.TerminalOp makeRef(boolean)>
	at soot.asm.AsmMethodSource.getBody(AsmMethodSource.java:1883)
	at soot.SootMethod.getBodyFromMethodSource(SootMethod.java:126)
	at soot.SootMethod.retrieveActiveBody(SootMethod.java:385)
	at soot.jimple.spark.builder.ContextInsensitiveBuilder.preJimplify(ContextInsensitiveBuilder.java:58)
	at soot.jimple.spark.SparkTransformer.internalTransform(SparkTransformer.java:80)
	at soot.SceneTransformer.transform(SceneTransformer.java:39)
	at soot.Transform.apply(Transform.java:90)
	at soot.RadioScenePack.internalApply(RadioScenePack.java:57)
	at soot.jimple.toolkits.callgraph.CallGraphPack.internalApply(CallGraphPack.java:49)
	at soot.Pack.apply(Pack.java:118)
	at soot.PackManager.runWholeProgramPacks(PackManager.java:564)
	at soot.PackManager.runPacksNormally(PackManager.java:455)
[...]

PR #882, opened 2018-02-01
Fix for #874

anddann:
Fixing the following bugs:
- the inner classes java.util.stream.FindOps$... are not properly resolved
- the order of arguments when constructing dynamic invoke expressions is wrong

Fix:
- always resolve all inner classes in `SootClassBuilder.java`
- fixing the order of arguments when dynamic invoke instructions are created `AsmMethodSource.java`


mbenz89:
Looks good to me. Thanks for your contribution!

PR #889, opened 2018-02-06
Some fixes

gnom7:


pnest:
I liked that you fixed java version detection and ``CollectConstants`` class - I'm facing this problem now.
Didn't like your formatting. You updated this line https://github.com/Sable/soot/pull/889/files#diff-212ae91a7c0ea161ef5a1c7bdd104330R110 and above with saving tabs. But the next one is formatted using spaces. That is not good. 
In every class you updated more than a half. In this case you'd better format every edited file using only one approach

gnom7:
Formatting for edited jbco classes has been [fixed](https://github.com/Sable/soot/pull/889/commits/3b5ac35ce930f0287fb0b4ab2288c370c2b062f5) as @pnest suggested, thanks.

StevenArzt:
To avoid the formatting problems, we have configured the Eclipse project to automatically re-format the code whenever a code file is saved. That should make the formatting uniform for at least those contributors that work with Eclipse.

PR #895, opened 2018-02-12
Fix ArithmeticTransformer division handling

gnom7:
when denominator is not a multiple of 2 we cannot just ignore reminder (i.e. transforming 1000/5 to 1000 >> -30), for now just skip such cases, ideally at some point we should come up with more sophisticated approach


StevenArzt:
ready to test

mbenz89:
add to whitelist

PR #899, opened 2018-02-13
Fixed some issues with LibraryMethodWrapperBuilder, MethodRenamer

gnom7:
We cannot extract calls to 'super' to the static methods, as it is not permitted on the language level to call 'super' on objects from the outside.
So without this fix we get such (not working) wrappers:
```
public static void run(Object arg, Object libObj) {
    super.doSomething(arg); //ignoring libObj
}
```
the fix is to skip `specialinvoke` instructions


gnom7:
Update: MethodRenamer now properly handles some cases of invokeinterface when library class overrides/implements method of app class (through child which extends this library class)

PR #900, opened 2018-02-14
Introduced class loading strategy for plugins

bergerbd:
Soot uses reflection to create instances of plugins.  This approach does not work in all execution environments, e.g. OSGi. In OSGi one cannot load an arbitrary class by name as long as it is not a explicit dependency. 

The changes in this pull request introduce a class loading strategy for plugin instantiation. The default implementation still employs Java reflection and does not change the current behavior. Nevertheless, it is possible to write new strategies coping with special execution environments and register them using ```PluginLoader.setClassLoadingStrategy```.


StevenArzt:
add to whitelist

PR #907, opened 2018-03-01
ClassRenamer could rename different classes to the same name

pnest:
There were a bug with ``soot.jbco.jimpleTransformations.ClassRenamer``: it could rename different classes to the same name. I ran into this bug when these classes were at the same package. As a result, one class were overwritten by another when transformed code was put into output jar

Here is my solution.
I created singleton from ``soot.jbco.jimpleTransformations.ClassRenamer`` and made sure that there are no puts into ``oldToNewClassNames`` between the moment when we check there are no duplicates in that map and the moment we put new name into it.

As I hid internal representations from other code, I added methods that provide managing functionality: 
```java
soot.jbco.jimpleTransformations.ClassRenamer#addClassNameMapping
soot.jbco.jimpleTransformations.ClassRenamer#getClassNameMapping
```
I believe this PR might cause arguments as it breaks backward compatibility. So there my arguments:
* ``soot.jbco.jimpleTransformations.ClassRenamer``'s work is unexpected: we cannot rely on it as it can break output code and we won't know about it before running on JVM
* ``soot.jbco.jimpleTransformations`` package is deprecated: there were no guarantee that backward compatibility will be saved
* users still can use recent stable release

### DIFF
--- a/src/main/generated/singletons/soot/Singletons.java
+++ b/src/main/generated/singletons/soot/Singletons.java
@@ -23,18 +23,21 @@
 package soot;
 
 /** A class to group together all the global variables in Soot. */
+@javax.annotation.Generated(value = "Saxonica v3.0", date = "2018-02-28T23:07:31.883+03:00", comments = "from singletons.xml")
 public class Singletons {
+
     public final class Global {
         private Global() {}
     }
+
     protected Global g = new Global();
 
     private soot.PhaseOptions instance_soot_PhaseOptions;
     public soot.PhaseOptions soot_PhaseOptions() {
-        if( instance_soot_PhaseOptions == null ) {
+        if (instance_soot_PhaseOptions == null) {
 	       	synchronized (this) {
-		        if( instance_soot_PhaseOptions == null )
-	        		instance_soot_PhaseOptions = new soot.PhaseOptions( g );
+		        if (instance_soot_PhaseOptions == null)
+	        		instance_soot_PhaseOptions = new soot.PhaseOptions(g);
 	       	}
        	}
         return instance_soot_PhaseOptions;
@@ -45,10 +48,10 @@ public class Singletons {
 
     private soot.jimple.toolkits.callgraph.VirtualCalls instance_soot_jimple_toolkits_callgraph_VirtualCalls;
     public soot.jimple.toolkits.callgraph.VirtualCalls soot_jimple_toolkits_callgraph_VirtualCalls() {
-        if( instance_soot_jimple_toolkits_callgraph_VirtualCalls == null ) {
+        if (instance_soot_jimple_toolkits_callgraph_VirtualCalls == null) {
 	       	synchronized (this) {
-		        if( instance_soot_jimple_toolkits_callgraph_VirtualCalls == null )
-	        		instance_soot_jimple_toolkits_callgraph_VirtualCalls = new soot.jimple.toolkits.callgraph.VirtualCalls( g );
+		        if (instance_soot_jimple_toolkits_callgraph_VirtualCalls == null)
+	        		instance_soot_jimple_toolkits_callgraph_VirtualCalls = new soot.jimple.toolkits.callgraph.VirtualCalls(g);
 	       	}
        	}
         return instance_soot_jimple_toolkits_callgraph_VirtualCalls;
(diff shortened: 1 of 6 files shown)


StevenArzt:
If the existing functionality breaks code, there's little point in arguing about backwards compatibility. In such a case, we  need a fix and there is no way around it. To the best of my knowledge, jbco isn't in such a widespread use either, so even if there is code that expects a certain behavior, that should not be a big impact.  Lastly, it's an obfuscator, so I would assume some degree of randomness as to how the renaming works.

StevenArzt:
ok to test

pnest:
Could someone, please, merge this PR?

mbenz89:
There are some conflicts after Steven's recent merge of the "logging"-branch. Can you resolve them first?

pnest:
@mbenz89 updated

src/main/java/soot/jimple/internal/JInterfaceInvokeExpr.java
```diff
@@ -19,54 +19,65 @@
...
 
         //Check that the method's class is resolved enough
         //CheckLevel returns without doing anything because we can be not 'done' resolving
         methodRef.declaringClass().checkLevelIgnoreResolving(SootClass.HIERARCHY);
+
         //now check if the class is valid
-        if(!methodRef.declaringClass().isInterface() && !methodRef.declaringClass().isPhantom()) {
-        	throw new RuntimeException("Trying to create interface invoke expression for non-interface type: " +
-        			methodRef.declaringClass() +
-        			" Use JVirtualInvokeExpr or JSpecialInvokeExpr instead!");
```
mbenz89: Here, you allow methods that aren't declared by interfaces but normal classes as well. Ist this correct? As I understand the language specification, this is illegal. Can you craft an example where an invokeinterface instruction can actually invoke a class declared method?

src/main/java/soot/jimple/internal/JInterfaceInvokeExpr.java
```diff
@@ -19,54 +19,65 @@
...
+        if (methodRef.isStatic()) {
```
mbenz89: Here, you allow methods that aren't declared by interfaces but normal classes as well. Ist this correct? As I understand the language specification, this is illegal. Can you craft an example where an invokeinterface instruction can actually invoke a class declared method?

src/main/java/soot/jimple/internal/JInterfaceInvokeExpr.java
pnest: You must be right. I'll remove this commit.
Btw, I haven't found any mentions in the JVM spec that this replacement is impossible. Do yo have this information at hand?

mbenz89: https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.10.1.9.invokeinterface

To me, the spec is a bit ambiguous. 

"Its first operand, CP, refers to a constant pool entry denoting an interface method named MethodName with descriptor Descriptor that is a member of an interface MethodIntfName."

I interpret this as "The method has to be an interface method".

However, this would be redundant then:

"MethodName is not <init>.

MethodName is not <clinit>."

The safest way to check this would be crafting an example using an invokeinterface instruction to call a non-interface method and see if it works.

pnest: To check this I created next class:
```java
public class TestClass {
    public void testMethod() {
        this.toString();
    }
}
```
To call `toString` method I used `invokeinterface`:
```java
Jimple.v().newInterfaceInvokeExpr(body.getThisLocal(), toString.makeRef());
```
When I loaded this class and tried to call `testMethod` I got this exception:
```
Caused by: java.lang.IncompatibleClassChangeError: Found class java.lang.Object, but interface was expected
	at TestClass.testMethod(Unknown Source)
```
So we cannot replace `invokeinterface` with `invokevirtual` without consequences. The check in `soot.jimple.internal.JInterfaceInvokeExpr#JInterfaceInvokeExpr` was correct.

*I reverted this commit and the PR is ready to be merged*

mbenz89: Thanks for clarifying!

PR #910, opened 2018-03-05
Use slf4j instead of G.v().out 

jpstotz:
As discussed in https://github.com/Sable/soot/pull/401#issuecomment-341720455 I changed all the G.v().out requests to slf4j.


StevenArzt:
ready to test

StevenArzt:
ok to test

StevenArzt:
@jpstotz Can you have a look at the test failures, please?

jpstotz:
I accidentally modified the code that was analyzed (`MethodThrowableSetClass.java`) by the TestCase that failed. Seems like the TestCase relies on `e.printStracTrace()`. reverted that file and now it passes the test.

The strange part is that on my local system the EnumTest and some other tests now fail but on the ci those tests work. Can it be that some tests are not reliable/deterministic at the moment?

mbenz89:
@jpstotz Thanks for your contribution!

I've never stumbled across any non-determinism regarding the tests outcomes. I wouldn't say there isn't, though. Were you able to track down the issue locally?

jpstotz:
I just performed some test using different branches. In the end I executed the branch with the failing tests using the command `mvn clean package  -U` (clean and update snapshots). Afterwards no test fails anymore.

I don't fully understand it but my guess is that it was caused by an outdated snapshot of a dependency or a sub-dependency of a snapshot. The failing String comparison assertion e.g. of `EnumTest` were looking not completely wrong, but had minor differences. It looked like it had been formatted with a different version.

dibollinger:
Looks like you also replaced all the calls to `G.v().out` in Main.java with `logger.debug()`. Combined with the fact that the default logging level is set to `INFO` means that there is now by default absolutely no console output anymore, apart from errors...

mbenz89:
Good point! @jpstotz Could you fix this?

jpstotz:
I already did what Github suggested and deleted the slf4j branch. Therefore I made a new pull request: https://github.com/Sable/soot/pull/916

PR #914, opened 2018-03-13
Fixing LibraryMethodWrapperBuilder

pnest:
Fixed `soot.jbco.jimpleTransformations.LibraryMethodWrappersBuilder` that could produce incomplete methods.
Let's pretend we have the following code:
```java
public class CustomWatcher implements android.text.TextWatcher {
    // code
    @Override
    public void afterTextChanged(android.text.Editable editable) {
        if (editable.toString().equals("something")) {
            return;
        }
        // code
    }
}
```
Expression `editable.toString()` is `invokeinterface`. Method `soot.jimple.InvokeExpr#getMethod` returns `Object#toString()` that is absolutely correct because jar was compiled with `android-api.jar` (not the implementation). So in this case expression `editable.toString()` is not `invokeinterface` anymore, but `invokevirtual`. But `LibraryMethodWrappersBuilder` treated it as `invokeinterface` that led to runtime exception. As exception was ignored the transformer left new generated method incomplete.
Another problem in this case is incorrect treating with API methods: if we allow `LibraryMethodWrappersBuilder` replace `invokeinterface` with `invokevirtual` that will lead to replacing `android.text.Editable#toString()` with `Object#toString()` that is wrong.
So my solution is checking if we "lose" interface (and find class instead) and skip this transformation

---
Also I added `@Deprecated` annotation to `soot.jbco.IJbcoTransform#out` (will be replace with logger), `soot.jbco.IJbcoTransform#output` (added `soot.jbco.IJbcoTransform#isVerbose` method that gets state at runtime) and `soot.jbco.IJbcoTransform#debug`  (added `soot.jbco.IJbcoTransform#isDebugEnabled` method that gets state at runtime)

### DIFF
--- a/src/main/java/soot/jbco/IJbcoTransform.java
+++ b/src/main/java/soot/jbco/IJbcoTransform.java
@@ -19,20 +19,61 @@
 
 package soot.jbco;
 
-import java.io.PrintStream;
 import soot.G;
 
+import java.io.PrintStream;
+
 /**
- * @author Michael Batchelder 
- * 
- * Created on 19-Jun-2006 
+ * @author Michael Batchelder
+ * <p>
+ * Created on 19-Jun-2006
  */
 public interface IJbcoTransform {
 
-  public PrintStream out = soot.G.v().out;
-  public boolean output = G.v().soot_options_Options().verbose() || soot.jbco.Main.jbcoVerbose;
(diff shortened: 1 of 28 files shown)


pnest:
Can someone look at it?

mbenz89:
add to whitelist

PR #916, opened 2018-03-15
slf4j calls reverted to System.out.println/System.err.println

jpstotz:
As requested in https://github.com/Sable/soot/pull/910#issuecomment-373384675

I don't know why git so stupid and thinks everything has changed.

### DIFF
--- a/src/main/java/soot/Main.java
+++ b/src/main/java/soot/Main.java
@@ -47,7 +47,6 @@ import com.google.common.base.Joiner;
 
 /** Main class for Soot; provides Soot's command-line user interface. */
 public class Main {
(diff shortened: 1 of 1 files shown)


mbenz89:
I would have expected to change the log level instead of using System.out. Why have you gone in this direction?

jpstotz:
Sorry, my mindreader is disfunctional at the moment ;)

Anyway as you pointed out some months ago there is no output without logback. Therefore changing the loglevel will not help at all because only slf4j-api is currently included in soot.

Anyway IMHO a cli application should not communicate using a logger with it's users.

mbenz89:
Thinking about it, I guess you are right.

PR #921, opened 2018-03-23
Fixing soot.jbco.jimpleTransformations.MethodRenamer

pnest:
Fixing `soot.jbco.jimpleTransformations.MethodRenamer` to handle "indirect" inheritance. Consider the next class hierarchy:
diagram-one
Method `method()` cannot be renamed in class `C` because it implements library one. As we leave this method name as is, none of superclasses of class `C` cannot change this name too. As we do not rename `A#method()` we must not change its name in any implementation of `A` including `D`. So this case boils down to indirect connection between `D#method()` and `E#method`: "indirect" inheritance. Although updated version of `MethodRenamer` recognizes this connection and skips renaming of such methods, it still processes rest methods with same name but different number of parameters. Thus, `D#method(java.lang.String, int)`, `C#method(java.lang.String)` and `C#method(long, int)` will be renamed, but will have _the same_ name. The restriction to have the same new method name for the methods with the same name but different parameters, was introduced to support _renaming map_: the result of call `soot.jbco.jimpleTransformations.MethodRenamer#getRenamingMap(String className)` for any class must return map with old-to-new method names. The structure of map does not allow us to store different new method names for one method (I mean we will not know exactly the new method name if we receive a collection, not a single value). Note, that this rule applies to every single _tree_ of classes. Thus, a method with name `method` in other class hierarchy not related to this one, will have other name.

Also I extracted generation names logic into separate class; updated javadoc and formatted printing logs

### DIFF
--- a/src/main/java/soot/Hierarchy.java
+++ b/src/main/java/soot/Hierarchy.java
@@ -25,6 +25,10 @@
 
 package soot;
 
+import soot.jimple.SpecialInvokeExpr;
+import soot.util.ArraySet;
+import soot.util.Chain;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -34,10 +38,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-
-import soot.jimple.SpecialInvokeExpr;
-import soot.util.ArraySet;
-import soot.util.Chain;
 
 /**
  * Represents the class hierarchy. It is closely linked to a Scene, and must be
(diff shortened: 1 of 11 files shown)


pnest:
Hello! Can anyone take a look at it?

pnest:
I'm kindly sorry, but could someone, please, check my pull request?

src/main/java/soot/Hierarchy.java
```diff
@@ -202,70 +202,103 @@ protected void checkState() {
...
+			current = current.getSuperclass();
 		}
 
-		return Collections.unmodifiableList(l);
+		return Collections.unmodifiableList(superclasses);
 	}
 
-	/** Returns a list of subinterfaces of c, including itself. */
-	public List<SootClass> getSubinterfacesOfIncluding(SootClass c) {
-		c.checkLevel(SootClass.HIERARCHY);
```
mbenz89: Why have you removed the check for the resolving level here while still preserving it everywhere else? This seems to be an issue to me.

PR #926, opened 2018-04-14
Fixing GotoInstumenter

pnest:
There were cases when `GotoInstumenter` might create `Trap` with `null` end unit. That made generated code wrong (and actually it fails on validation).
This case appeared when we move only one statement to the end of method body. When we create try-catch block that covers this statement, we cannot find the end statement (because of random process).

Also I created `HorrorNameGenerator` that uses _similar_ letters from different alphabets.

### DIFF
--- a/src/main/java/soot/jbco/jimpleTransformations/GotoInstrumenter.java
+++ b/src/main/java/soot/jbco/jimpleTransformations/GotoInstrumenter.java
@@ -19,186 +19,226 @@
 
 package soot.jbco.jimpleTransformations;
 
-import soot.*;
-import soot.util.*;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Iterators;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import soot.Body;
+import soot.BodyTransformer;
+import soot.Local;
+import soot.PatchingChain;
+import soot.RefType;
+import soot.Scene;
+import soot.SootMethod;
+import soot.Trap;
+import soot.Unit;
+import soot.UnitBox;
 import soot.jbco.IJbcoTransform;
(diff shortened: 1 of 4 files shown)


StevenArzt:
add to whitelist

StevenArzt:
@pnest Thanks a lot for your work on improving jbco. Do you happen to have JUnit test cases for the examples with which you found the problems you fixed? It would be great to include those tests into our continuous integration process, along with the (unfortunately way too few) tests we already have.

pnest:
@StevenArzt I've been thinking a lot about testing jbco. There are two problems I see for now:
1. filling the Scene on which transformers will be tested, and
2. running Soot multiple times on the same JVM

As to first problem, I see the only solution is simply writing java code and running Soot to load them. Using Soot to fill Scene directly is much harder and not so obvious.

As to second problem, I haven't found any solution. Soot (and jbco transformers) use static fields to store their data. Thus to create reproducible testing environment we must either clear all this fields manually, or refactor code in order to forbid using them. Also there is the third option - run every single test in different process. But this will increase testing time significantly. 

Step by step I move to the second solution - removing usage of static fields. In this case we can just recreate transformer, or use `G.reset()` if transformer is singleton. 

Do you see any other solution?

StevenArzt:
For some other test cases, we have classes on disk that we compile when Soot is compiled and that we load into Soot when the test is started. That is indeed the easiest way to create tests for transformers in Soot, although it is a system test and no longer a unit test in its usual intention.

For the static fields provided by Soot, G.reset() is indeed the only reasonable solution that we have. A transformer should not have static fields on its own. Usually, you do not need state that is kept beyond the lifetime of the transformer - at least, I don't see any reason for such a requirement right now. If you really need to, you can convert the transformer itself into a singleton.

PR #928, opened 2018-04-18
Codingstyle with changes

rajatkhanna1994:


StevenArzt:
ok to test

StevenArzt:
In the Eclipse properties, we have already enabled code formatting on save, i.e., whenever a code file is saved after a modification, ther complete file gets reformatted automatically, Can you please check whether your style is compatible with what Eclipse does? Especially concerning indentation, it seems as if you use spaces, while we normally use tabs.

mbenz89:
Hi Steven.

With this pr we aim to enforce a consistent coding style for Soot. Therefore, we provide different configurations for ecplise and IntelliJ to automatically apply formatting to help with merging (e.g., indentation, import sorting), but also a lot of other stuff that ensures code quality (e.g., removing redundant modifiers, surrounding one-liner control blocks with bracers).  

The coding guidelines are as close to Google's as possible. Sadly, we had to lower some of the rules to not break public interfaces or require too many manual changes to the codebase.

Furthermore, we enforce the style by having maven check for compliance before every build. This especially helps on the build server side. 

We've already written some tutorials on how to apply the formatting on your side. We would certainly be happy if you would find the time to check them out and report if you miss some steps.

[Contributors notice]( https://github.com/Sable/soot/wiki/Contributing-to-Soot)
[Formatting with eclipse]( https://github.com/Sable/soot/wiki/Formatting-for-Eclipse)
[Formatting with IntelliJ](https://github.com/Sable/soot/wiki/Formatting-for-IntelliJ)
[Help with merging after this pr](https://github.com/Sable/soot/wiki/Merging-changes-after-introduction-of-formatting-guidelines)

Regarding tabs vs spaces, I think there are pros and cons for both. I would rather not discuss this here :).
Regarding the config file, where exactly is it in the repository? I couldn't find the configuration. I guess, though, after configuring eclipse as stated in the wiki, the automatic save will apply the correct formatting.

mbenz89:
We checked this and eclipse format on save will use the formatting as configured in [our tutorial](https://github.com/Sable/soot/wiki/Formatting-for-Eclipse).
Only the 'clean up' step has to be invoked manually. But in theory it is sufficient to do this once before a commit (or at least in a dedicated commit before pushing)

mbenz89:
@StevenArzt Are there any concerns left from your side?

StevenArzt:
I'm fine. Someone will have to resolve the conflicts, but beyond that, I'm fine.

rajatkhanna1994:
I've resolved the conflicts and reformatted the conflicting files.

pnest:
Don't you think that line length of 150 symbols is too much? Who on Earth uses such line length?

mbenz89:
@rajatkhanna1994 I remember there was a reason for us setting the line length that high, do you remember it?

I found [this article](http://hilton.org.uk/blog/source-code-line-length) and I liked the reasoning about the standard being 125 chars due to GitHub diff views supporting it.

@rajatkhanna1994 Is there anything speaking against a smaller line width?
@pnest Would you consider something between 100 and 125 reasonable or what is your preference?

rajatkhanna1994:
@mbenz89 Earlier we tried setting the maxLineLength to 100, there were instances where formatter couldn't wrap long lines to chunks of 100 symbols. For such violations, we had to manually perform the wrapping. (but we didn't want too many manual changes in our codebase)
By setting the maxLineLength to 150, these violations were reduced.

I could give it a try (setting maxLineLength to 125), we could get around the aforementioned issue by relaxing the line length check in the checkstyle file (by setting maxLineLength to a higher value), while in the formatter configuration, maxLineLength stays as 125.

pnest:
@mbenz89 Thanks for article!

125 symbols seems to be reasonable solution: setting the same length as on github will make us use most of its features

mbenz89:
@rajatkhanna1994 I would rather not diverge automatic formatting rules from enforcement rules. Even if we shortened all lines to max 125 chars manually now, we couldn't prevent this from being necessary over and over again if we actually allow 150 chars in the checkstyle configuration. Could you please lower maxLineLength for both and see how much effort it would be to manually fix all violations that are not automatically fixed?

rajatkhanna1994:
@mbenz89 I lowered the maxLineLength in both the files, there are about 70 violations in 30 different files, which need to be manually corrected.
However, there is a slight issue with wrapping assignment statements. It's kind of hard to explain here, I'll email you the details.

PR #943, opened 2018-05-07
solve the bug in Tarjan's algorithm of Geom Points-to Analysis

fyuac:
Solve a bug in the function 

  private void callGraphDFS(int s)

in the file

  GeomPointsTo.java

which use Tarjan's algorithm to contract the SCCs.

The problem of this bug is that it may cause Soot to report an empty points-to set for some of the pointers when using Geom Points-to Analysis. In fact, this is a typical bug when implementing Tarjan's algorithm.


StevenArzt:
ok to test

Issue #937, opened 2018-04-27
Android: initialization values of static non-final fields missing in Jimple representation


jpstotz:
I encounter a problem with static fields in the Android/DEX part of Soot. Consider the following simple class:

    public class SimpleClass {
       public static String A = "!AABBCCDD!";
    }

If I compile this class to SimpleClass.class and then generate the Jimple code from it everything is fine.
`java -cp  soot-3.1.0-20180424.084712-9-jar-with-dependencies.jar soot.Main -w -f jimple  -cp target/classes/;"%JAVA_HOME%\jre\lib\rt.jar";"%JAVA_HOME%\jre\lib\jce.jar" SimpleClass`

The generated Jimple code contains the correct `<clinit>`:
    
    static void <clinit>()
    {
         <SimpleClass: java.lang.String A> = "!AABBCCDD!";
         return;
    }

If I do the same using a DEX file containing this class the `<clinit>` misses the initial value of the field:

    static void <clinit>()
    {
        return;
    }

I did this using different versions of Soot from the [Snapshot repository](https://soot-build.cs.uni-paderborn.de/nexus/#browse/browse/components:soot-snapshot). It does not matter if I use the oldest or the newest (3.1) version the result is always the same as above.

For making sure that the variable is really present in the dex code I also decompiled the APK using different versions of `apktool` and checked the Smali code - the initialization value was always present:

    # static fields
    .field public static A:Ljava/lang/String; = "!AABBCCDD!"

MarcMil:
For the case that this field is not final, I would suggest adding an appropriate assign statement to the *start* of the <clinit> method of the same class to emulate old behavior.
Some classes might assume that the presence of a constant tag (even on a non-final field) means that the corresponding field is in fact a constant and cannot be changed.

StevenArzt:
Soot normally uses tags to deal with such data. Have you tried to call getTags() on the SootField instance? You should then find an object of a class derived from ConstantValueTag, i.e., IntegerConstantValueTag, which contains the initialization value from the bytecode. If these tags are missing, that's a bug. If these tags are there, client analyses should evaluate them and no changes to the static initializer are required.

In fact, we have code that enforces that these tags are available. The Dexpler frontend loads the constant table from the dex bytecode. If we load Jimple files from disk (Jimple doesn't have a constant pool), we even infer the tags from the code in the static initializer. In the end, there should always be a tag - but not necessarily a static initializer,

jpstotz:
[...]

PR #944, opened 2018-05-07
Handle initial field value when creating a new SootField instance fro…

anantoni:
A minor modification to the way Soot handles DexFields and their conversion to SootFields. I added a field of String type to model the initial value for Android fields, since I see it right now it is not preserved.I'm also wondering whether this change can help with issue #937 in case it is accepted.

Best regards,
Tony

### DIFF
--- a/src/main/java/soot/Scene.java
+++ b/src/main/java/soot/Scene.java
@@ -1868,6 +1868,10 @@ public class Scene // extends AbstractHost
     return new SootMethod(name, parameterTypes, returnType, modifiers, thrownExceptions);
   }
 
+  public SootField makeSootField(String name, Type type, int modifiers, String initialValue) {
+    return new SootField(name, type, modifiers, initialValue);
+  }
+
   public SootField makeSootField(String name, Type type, int modifiers) {
     return new SootField(name, type, modifiers);
   }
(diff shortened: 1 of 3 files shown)


StevenArzt:
I don't think we should extend the SootField class. This change would mean that every analysis would need to be extended to look into this field and see whether there is an initial value. Furthermore, it's dalvik-specific, and I'd rather keep the Jimple classes as generic as possible. Lastly, we would have one more construct that is in the data object, but not in the Jimple language.

My proposal to add support for Dalvik initial values is as follows: Create an assignment in the clinit method of the respective class. That exactly models the behavior of an initial value. It is assigned when the class is loaded and can afterwards be overwritten at any time. With that solution, analyses would not have to be modified, because they have support for assignments anyway.

anantoni:
Thanks for the response Steven, yes I understand that your proposal is the best way to model this. Could you instruct me on how to write a pass that handles all field initial values for dex fields?

StevenArzt:
You need to write a transformer. For each class, you need to check whether there is already a clinit method. If so, you add the assignment. If not, you create the method and add the assignment. That's essentially a normal instrumentation (technically a SceneTransformer).

ericbodden:
I don't think this really needs to be a transformer. We could just generate an appropriate Jimple statement in Dexpler. @anantoni were you planning to still look into this?

StevenArzt:
I have looked into the code again. We have tags for these default values, so I don't even think we need anything like a transformer or a Jimple statement. Actually, the code for writing the tags is in the DexField class, right above the proposed change. Client code can just call getTags() and see if there is a ConstantValueTag.

mbenz89:
Related to #937

mbenz89:
@anantoni Any updates on this?

PR #954, opened 2018-06-08
Test struct. equivalence in equivTo of JimpleLocal

### LINKED ISSUE #953: Structural equality of JimpleLocal
I'm using Soot for a bytecode analysis task, and I want to compare two Jimple expressions containing `JimpleLocal`s on structural equality. However, the `equivTo` method does not test structural equality: It falls back to referential equality (a standard Java `.equals()` call). I would expect the structural equality of two `Local`s to be defined as their `type`s being equal, ignoring the name of the local. The relevant file in this repo can be found [here](https://github.com/Sable/soot/blob/master/src/main/java/soot/jimple/internal/JimpleLocal.java).

Would you welcome a PR that changes the implementation of the `equivTo` method (and the `equivHashCode` method) to only compare types?

gandreadis:
The JimpleLocal class has an `equivTo` and an `equivHashCode` method. Both claim to test for structural equality (as described in their Javadocs). However, as mentioned in #953, this is not actually the case. This change replaces the referential equality check with a type equality check.

Closes #953.

### DIFF
--- a/src/main/java/soot/jimple/internal/JimpleLocal.java
+++ b/src/main/java/soot/jimple/internal/JimpleLocal.java
@@ -56,7 +56,9 @@ public class JimpleLocal implements Local, ConvertToBaf {
   /** Returns true if the given object is structurally equal to this one. */
   @Override
   public boolean equivTo(Object o) {
-    if (!(o instanceof JimpleLocal)) return false;
+    if (!(o instanceof JimpleLocal)) {
+      return false;
+    }
 
     JimpleLocal that = (JimpleLocal) o;
     return Objects.equals(this.type, that.type);
--- a/src/main/java/soot/jimple/internal/JimpleLocal.java
+++ b/src/main/java/soot/jimple/internal/JimpleLocal.java
@@ -24,6 +24,7 @@ package soot.jimple.internal;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 import soot.Local;
 import soot.Scene;
@@ -55,7 +56,10 @@ public class JimpleLocal implements Local, ConvertToBaf {
   /** Returns true if the given object is structurally equal to this one. */
   @Override
   public boolean equivTo(Object o) {
-    return this.equals(o);
+    if (!(o instanceof JimpleLocal)) return false;
+
+    JimpleLocal that = (JimpleLocal) o;
+    return Objects.equals(this.type, that.type);
   }
 
   /**
@@ -64,10 +68,7 @@ public class JimpleLocal implements Local, ConvertToBaf {
   @Override
   public int equivHashCode() {
     final int prime = 31;
-    int result = 1;
-    result = prime * result + ((name == null) ? 0 : name.hashCode());
-    result = prime * result + ((type == null) ? 0 : type.hashCode());
-    return result;
+    return prime + ((type == null) ? 0 : type.hashCode());
   }
 
   /** Returns a clone of the current JimpleLocal. */


mbenz89:
ok to test

mbenz89:
Have you checked if the methods aren't used anywhere else where your new implementation would break functionality? There might be good reasons for checking the local name as well. Why don't you just compare the types of the locals in your client code instead of creating specific methods in the local class for it?

FWDekker:
@mbenz89 We haven't exhaustively checked all usages, but we'll do that in a second. We'll keep you up-to-date.

While it's possible to check the types of `Local`s in client code, this does not extend when the `equivTo` method is applied recursively. For example, if an `EqExpr` uses two `Local`s and compares it, the `EqExpr`'s `equivTo` method recursively calls `equivTo` on the two `Local`s. It is not possible to override this behaviour, unless we are to override all `equivTo` methods.
Perhaps there is a more elegant solution you know of?

FWDekker:
@mbenz89 From what I can see `equivTo` is mostly used in the other implementations of `equivTo`. In addition, there are some classes such as `JasminClass` and `BuildIntermediateAppClasses` that use the `equivTo` method, but I cannot say that I know Soot's codebase well enough to judge whether our change would break their implementations.

mbenz89:
I would rather not mess with this. At least not without being sure about the consequences of such a change.

Can't you use something like this:
```
  private boolean compareValueTypes(Stmt first, Stmt sec) {
    final List<Type> firstTypes = first.getUseBoxes().stream().map(v -> v.getValue().getType()).collect(Collectors.toList());
    final List<Type> secTypes = sec.getUseBoxes().stream().map(v -> v.getValue().getType()).collect(Collectors.toList());
    return firstTypes.equals(secTypes);
  }
```

FWDekker:
I understand your hesitation and perhaps an alternative would be preferable.

I've considered your suggestion, but I think that this could lead to false positives. Consider the following two diagrams:

Diagram A | Diagram B
:---:|:---:
Diagram A | Diagram B

Calling `getUseBoxes()` on either root will result in the list `D, B, C, A` (because they are visited in depth-first order), so it will look like the two lists are the same even though they aren't. I think it's necessary that the elements are compared recursively like in the `equivTo` method, since the structure of the tree is lost when it's flattened, but as far as I know this recursive comparison happens only in the `equivTo` method.

FWDekker:
The problem is not so much that `getUseBoxes()` uses depth-first ordering, since any ordering at all will cause the problem that there are two trees that will be visited in the same order. The solution to this problem requires that nodes are compared recursively _during_ iteration rather than after iteration.

ericbodden:
Thanks for the effort but I am not sure I follow your line of reasoning for the change. Why would one want to compare in terms of the type, but not the name? When would that ever make sense? Sounds like a very special use case to me. The reason for the current behaviour is that it's sensible in most cases. Some parts of Soot may even rely on this. Hence I am quite reluctant to just accept this change.

FWDekker:
@ericbodden Thank you for your response!

The reason we want to compare `Local`s by their type only is because we're writing a program that detects patterns in code. One part of detecting the patterns is determining whether two statements are equivalent. However, the name of a variable is not relevant in this case. For example, if we have
```java
ArrayList<String> list = new ArrayList<>();
list.add("foo");
return list.size();
```
and
```java
ArrayList<String> arrayList = new ArrayList<>();
arrayList.add("foo");
return arrayList.size();
```
then our program should conclude that these two snippets are equivalent and thus constitute a pattern, even though we have `list` in the first snippet and `arrayList` in the second snippet.

I understand your hesitance to merge these changes into Soot because the users of the behaviour we're changing may assume that `Local`s are equivalent only if their name _and_ type are the same. Is there perhaps another way to achieve the desired behaviour that results in compatibility with the current implementation of `equivTo`?

ericbodden:
Thanks for the explanation Felix.

I can see that this is currently hard to achieve with Soot as it is. A clean solution would be to have a variant of equivTo that accepts some sort of Comparator callback, similar to compareTo in Comparable interface. This mechanism would then be extensible, you could just plug in your own comparator.

We are currently working towards a new major release of Soot that will have many breaking changes. 
@mbenz89 we should certainly include such support there. But I think it also would not hurt to add such an addition to this branch of Soot already, as it would just be a backwards-compatible addition.
@FWDekker do you think you would be able to implement an appropriate solution and update the pull request accordingly?

linghuiluo:
this PR is too old, please reopen it if it is needed.

PR #957, opened 2018-06-12
Android manifest reading - improvements for obfuscated APKs

jpstotz:
Obfuscated APKs often remove the attribute names from AndroidManifest.xml and use the resourceId instead. In such a  case Soot does not find the entries targetSdkVersion and minSdkVersion.

This patch extends the loading mechanism with a fallback mechanism that checks for the known static resourceIds of targetSdkVersion and minSdkVersion.


mbenz89:
add to whitelist

ericbodden:
Can this be merged? @mbenz89: Are we automatically checking license headers also for pull requests?

mbenz89:
@ericbodden Yes, we do check the license header for pull requests. The pr fails if the license is absent or incorrect. You can see the passing/failing checks by clicking "show all checks" in the merge UI.
 
In this pr only a few lines of an existing file have changed, so the license header is unaffected.

PR #958, opened 2018-06-19
Fix CHA call graph building was failing on methods of type 'abstract finalize()'

agorski3:


mbenz89:
ok to test

src/main/java/soot/jimple/toolkits/callgraph/OnFlyCallGraphBuilder.java
```diff
@@ -550,6 +550,25 @@ public void addType(Local receiver, Context srcContext, Type type, Context typeC
...
           VirtualCalls.v().resolve(type, receiver.getType(), site.subSig(), site.container(), targetsQueue, appOnly);
+          if(!targets.hasNext()) {
+            /* In the situation where we find nothing to resolve an invoke to in the first call, this
+             * might be because the type for the invoking object is a abstract class and the method is
+             * declared in a parent class. In this situation, when the abstract class has no classes
+             * that extend it in the scene, resolve would not find any targets for the invoke, even
+             * if the parent contained a possible target. 
+             * 
+             * This may have been by design since without a concrete class, we have no idea if the 
+             * method in the parent class is override. However, the same could be said for any non
```
mbenz89: How could you possibly override a private method?

src/main/java/soot/jimple/toolkits/callgraph/OnFlyCallGraphBuilder.java
```diff
@@ -550,6 +550,25 @@ public void addType(Local receiver, Context srcContext, Type type, Context typeC
...
+             * private method in the abstract class (and these all resolve fine inside the abstract
+             * class even though there are no sub classes of the abstract class). This makes this
+             * situation a corner case.
+             * 
+             * Where as, it used to not resolve any targets in this situation, I want to at least
+             * resolve the method in the parent class if there is one (as this is technically a
+             * possibility and the only information we have).
+             */
+            VirtualCalls.v().resolveSuperType(type, receiver.getType(), site.subSig(), targetsQueue, appOnly);
```
mbenz89: As you mention, the decision to not have any targets for an invitation because we do not know if the invoked method is overridden by a subclass might be a design decision. 

Since this behavior should be configurable to the needs of an analyst, I suggest introducing an option to Soot which by default uses the current behavior and can be used to configure Soot to work as you propose.

src/main/java/soot/jimple/toolkits/callgraph/OnFlyCallGraphBuilder.java
agorski3: You can't. The comment says this corner case only applies to **non private methods**. It is using the same resolution procedure you would use to resolve any method to a method in a parent class normally so there should not be any issue here.

agorski3:
This one should be good to go as well @mbenz89

mbenz89:
Thank's for your contribution!

PR #959, opened 2018-06-19
Minor bug fixes to dava output

agorski3:
- Move "outputDAVA" method outside of "postProcessDAVA" to make
  debugging easier.
- Make sure non-public empty constructors are not removed as
  these are no longer default constructors.


mbenz89:
ok to test

PR #961, opened 2018-06-19
Use global soot options when creating an ExceptionalUnitGraph

agorski3:
All JimpleBasedICFG were forced to use the UnitThrowAnalysis and and omit
exceptional unit edges when constructing an ExceptionalUnitGraph. There
are global settings for these options, so it makes more sense to use the
global settings than to have them hardcoded. Esepeically since this
exception to the global settings is not stated anywhere.


mbenz89:
ok to test

mbenz89:
Thanks for all of your contributions btw!

agorski3:
No problem. These were all changes I made about a year ago that I have been sitting on because of all the refactoring you guys have been doing. (hence why there are so many). I have been meaning to submit pull requests for a while I just had to find time to sync things back up.

mbenz89:
I see this can be a major pain at times. Thanks for the efforts!

PR #962, opened 2018-06-19
Make the LoopFinder easier to call and clean it up

agorski3:
- Reformatted extremly old code in the LoopFinder so that it uses more
  up-to-date java standards.
- Made it much easier to call the LoopFinder without having to go through
  the transofmation procedure since this was not really being used by the
  LoopFinder. It is still possible to call it that way though.
- Replaced a few calls in other parts of the code to account for the
  change in the way to get the loops.


mbenz89:
ok to test

mbenz89:
Please fix the styling issues. Otherwise, this is fine to be merged.
You can find instructions on how to set up a code formatter according to our guideline in the wiki: https://github.com/Sable/soot/wiki/Contributing-to-Soot

PR #964, opened 2018-06-19
BugFix: Issues with throwable and phantom classes

agorski3:
- The previous fix for handling Throwable not being the base class
  for phantom exception classes had some holes in it. This fixes
  those.
- Exception super class hierarchy does not really need to check
  if things are phantom, it just needs to make sure the first
  element is always Throwable and that Throwable only occurs once.
  Ensuring this is sufficant to cover phantom Exceptions and exceptions
  with a well formed super class hierarchy (since they will already have
  Throwable as their root minus Object).


mbenz89:
ok to test

mbenz89:
I fixed it.

PR #966, opened 2018-06-19
Changes related to the LocalNameStandardizer

agorski3:
- Append 0's to local names to make all local names in a method body
  the same length. This makes it easier to perform substitutions on
  the local names when only the string of the local name is available
  (i.e. when code and code segments are output in string form to be
  post-processed at a later time).
- Ensure the LocalNameStandardizer performs the last modification
  of a jimple body generated from dex to make sure the local names
  remain in the sorted order after all other changes are made.
- Make sure the phase options used by the LocalNameStandardizer are
  actually set based on the global options when generating jimple
  from dex. These are the same phase options that JimpleBodyPack
  sets based on the global options.
- Clean up some poorly formated comments in DexBody.
- Use the options object already available to check for options
  in JimpleBodyPack.

modification to the jimple body for stability reasons. Also make
sure it considers its options when running


ericbodden:
Thanks Albert! I am not too happy about the padding of the local names, as it may make the Jimple harder to read. It seems like this is really something particular to your use case. Maybe you rather keep that change to your Soot extension?

agorski3:
I could also make it so it only pads the names when the fixed ordering on
the variable names option is enabled since that is an option I basically
added a while back.

Or we could add a separate phase option just for padding.

ericbodden:
Hi Albert. Yes please let's bind it to the fixed-ordering option then. Maybe you can then also update the option's description. Thanks!

mbenz89:
ok to test

agorski3:
This one should be good now.

PR #967, opened 2018-06-19
A number of multi-threaded fixes to SootField and SootMethod

agorski3:
- Removed the restriction from setModifiers in SootField and SootMethod
  that a class needs to be an application class to set the modifiers of
  these objects. This seems to be an old restriction and I could not find
  any reason it needed to be around.
- Allowed SootField and SootMethod to be set to phantom even if their
  declaring class is not phantom. This fixes some holes where one might
  not have exactly the right library class but phantom methods and fields
  can be added to fill in some of the holes.
- Fixed a number of multi-thread race conditions related to the
  MethodSource and activeBody fields of SootMethod. Mainly,
  retrieveActiveBody, setActiveBody, and getActiveBody did not properly
  take into account changes that might occur to the class fields when
  not properly synchronized. This ensures that all modification to the
  MethodSource and activeBody fields occur in synchronized blocks and
  that any other field accesses these two depend on also occur in said
  blocks.
- Set all fields to private because modifying them without going through
  properly synchronized set methods might lead to race conditions. All
  child classes should use the seter method to set the fields.
- Cached the strings for the signature and subsignature in both SootField
  and SootMethod.
  - Constantly creating the signature strings in a multi-threaded
    analysis is extremly wasteful and time consuming (especially when the
    strings do not change).
  - This change caches the strings and ensures they only change when
    a modification to the data making up the signature or subsignature
    changes.
  - These changes are thread safe.
- Added a way to create a signature string from the declaring class and
  the subsignature string (without having to go through the trouble of
  decomposing the subsignature string into its parts if you already have
  the subsignature string).


mbenz89:
ok to test

agorski3:
I can make an option for the first part. 

For the second, I can change it back to protected and you would just have to make sure no changes in the Soot code try to modify the fields of these classes without going through the synchronized setter methods. This would give people who are extending the classes access to the fields (they would just have to ensure thread safety in their own code). How does this sound @mbenz89 ?

Also apologies for the delayed responses on all of these, I got busy with something else right after I submitted them all.

mbenz89:
@StevenArzt I would like to hear your opinion on this pr.

StevenArzt:
I would not make any fields private that used to be protected in earlier versions of Soot. In our projects, we heavily rely on being able to derive our own classes from the Soot classes. We also run our own concurrency model on top of Soot, which means that we need to take care of consistency and possible race conditions anyway. I think this is a suitable model: If you create a derived class, you need to know what you're doing. Having a cache for signatures and subsignatures is definitely reasonable. In general, we must test such a change really well. If it works for a reasonable number of apps, I'm fine with it (once we change the fields back to protected).

mbenz89:
Have you tested these changes already for a reasonable number of apps?

mbenz89:
Thanks for the rework and your contribution in general!

PR #974, opened 2018-06-25
fix NPE in soot.dexpler.DexBody.jimplify(Body b, SootMethod m)

arkbriar:
Detailed:
If the last instruction of body is InvodeStaticInstruction, the last dangling instruction's finalize will not be called. This leads to a NullPointerException in the following fixLineNumbers() method. This situation is not very common but still exists, e.g. some invokes android.util.Log.i() at the end of the method body.

And further more, maybe we could use some **code formatter such as clang-format to make our codes more readable**?
Anyway, please merge this into mainline to fix this bug.

### DIFF
--- a/src/main/java/soot/dexpler/DexBody.java
+++ b/src/main/java/soot/dexpler/DexBody.java
@@ -528,6 +528,9 @@ public class DexBody {
         instruction.setLineNumber(prevLineNumber);
       }
     }
+    if (dangling != null) {
+      dangling.finalize(this, null);
+    }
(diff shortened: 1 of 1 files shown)


mbenz89:
ok to test

mbenz89:
We have coding convetions and corresponding checkstyle and formatter rules ([see](https://github.com/Sable/soot/blob/develop/CONTRIBUTING.md))

arkbriar:
👍

mbenz89:
Thanks for your contribution!

Issue #976, opened 2018-06-28
Validation bug: InvokeArgumentValidator when validating DynamicInvokeExpr


xenox1886:
When validating a method body with the `InvokeArgumentValidator` (manually or when writing a class file with Jasmin), the `validate` method calls `getMethod` on the `InvokeExpr` of the `Stmt` currently being checked:
```java
InvokeExpr iinvExpr = s.getInvokeExpr();
SootMethod callee = iinvExpr.getMethod();
```
However, if `iinvExpr` happens to be of type `DynamicInvokeExpr`, and therefore not resolvable at compile time, `SootMethodRefImpl.resolve` (which gets called by `iinvExpr.getMethod`) will throw a `ClassResolutionFailedException`. 

This is particularly annoying, because `InvokeArgumentValidator` is a basic validator, i.e. (afaik) disabling soot's validation option doesn't make a difference, since it gets called anyway.
It also entails that one can't read dynamic expressions (e.g. when using lambdas) and writing them, without any changes, back with Jasmin, since the validation fails.

Possible (quick and dirty) fix in `InvokeArgumentValidator.validate`: 
```java
InvokeExpr iinvExpr = s.getInvokeExpr();
if (iinvExpr instanceof DynamicInvokeExpr){
    continue;
}
SootMethod callee = iinvExpr.getMethod();
```

(Also I assume the first line of `SootMethodRefImpl.tryResove` should actually throw an exception in this case, but it fails to compare the class name to "soot.dummy.InvokeDynamic")

mbenz89:
I find it a bit strange that this problem wasn't observed before. Having lambdas in client code (and thus DynamicInvokes) shouldn't be a corner case.
Is there a reason for which you are using Jasmin? I am pretty sure that Soot defaults to the ASM Backend in the current version.

Can you recheck with ASM and if it still faile propose a PR with your fix(es)?

xenox1886:
No, I have no particular reason to use Jasmin, I just followed this tutorial: https://github.com/Sable/soot/wiki/Creating-a-class-from-scratch
I've switched to the BafASM backend now (that's what you meant, right?), but I get the same error (see stacktrace below). I'll make a PR, as soon as I got the time.
Thanks
```java
[...]

PR #979, opened 2018-06-29
dynamic method invocations now handled as phantom classes

xenox1886:
References issue #976 
When resolving methods of dynamic method invocations, a `soot.SootMethodRefImpl$ClassResolutionFailedException` was thrown, since the declaring class of such invocations is unknown at compile time. However, this leads to unwanted bevahior when e.g. validating a method containing lambdas (which use dynamic method invocations), as well as writing a `SootClass` to a .class file (which also validates its methods), because an exception is thrown.
According to https://github.com/Sable/soot/wiki/Java%27s-Invokedynamic the dummy declaring class `soot.dummy.InvokeDynamic` of dynamic method invocations must be treated as a phantom class, regardless if phantom classes are enabled. This is now implemented.
Additionally, I've removed the `IllegalStateException` at the beginning of `tryResolve(StringBuffer trace)`, because it seems to address the same issue but wrongly (when compared to the wiki article). First, the class name the declaring class is compared to is hardcoded and wrong (`java.dyn.InvokeDynamic` instead of `soot.dummy.InvokeDynamic`) and second, it throws an exception instead of treating the class as a phantom class.

### DIFF
--- a/src/main/java/soot/SootMethodRefImpl.java
+++ b/src/main/java/soot/SootMethodRefImpl.java
@@ -167,10 +167,6 @@ public class SootMethodRefImpl implements SootMethodRef {
   }
 
   private SootMethod tryResolve(StringBuffer trace) {
-    if (declaringClass.getName().equals("java.dyn.InvokeDynamic")) {
-      throw new IllegalStateException("Cannot resolve invokedynamic method references at compile time!");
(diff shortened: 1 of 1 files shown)


mbenz89:
Thanks or your contribution!

PR #985, opened 2018-07-10
Fixed "remove" method of ArraySparseSet iterator.

Lex999:
The "remove" method of the ArraySparseSet iterator did not work correctly because it used a wrong index. This has been fixed now. A test for this method has been added.


mbenz89:
Thanks for your contribution!

PR #987, opened 2018-07-11
Call graph generation optimized

jpstotz:
# Overview
The soot call graph implementation is pretty slow especially with apps that make massive use of Lambda expressions, because every Lambda expression is implemented as anonymous inner class. Therefore the class hierarchy of `Runnable` and similar classes can get extremely large.

This pull request contains a patch with can reduce the call graph generation time by 10 to 600% depending on the app and the number of available CPU cores.

# In Detail
While analyzing several Android apps I encountered some that require a lot of time (com.ubercab.driver 6h and com.google.android.googlequicksearchbox 13 h) for generating a simple CHA call graph. Using VisualVM I found out that `FastHierarchy.canStoreClass(..)` takes up to 95% of the used CPU cycles - in detail the following loop and the contained calls to `it.hasNext()` and `it.next()` take the majority:

```
      for (Iterator<SootClass> it = impl.iterator(); it.hasNext();) {
        parentInterval = classToInterval.get(it.next());
        if (parentInterval != null && parentInterval.isSubrange(childInterval)) {
          return true;
        }
      }
```

I replaced this loop using the Java stream API wich allows to perform the task in multiple threads. Further tests showed that always using the parallel approach can slow down the generation of certain apps. Therefore in the end I end up in a mixed-approach wich is pretty easy thanks to the Java stream api:

If `impl` has less than 1000 elements the sequential approach is used, if there are more elements the multi-threaded parallel approach is used.

In the end the call graph generation time was reduced from 6 to about 1 hour and from 13 to about 2 hours.

BTW: In my tests the sequential approach was already 10 to 20% faster than the for loop plus iterator (Oracle Java 8.0_171).

### DIFF
--- a/src/main/java/soot/FastHierarchy.java
+++ b/src/main/java/soot/FastHierarchy.java
@@ -30,6 +30,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import soot.jimple.SpecialInvokeExpr;
 import soot.util.ConcurrentHashMultiMap;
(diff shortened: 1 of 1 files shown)


mbenz89:
Thanks for your contribution and the detailed explanation!
Any ideas why the sequential stream is faster than an iterator? I would have expected the opposite.

jpstotz:
I don't have a real explanation why the stream API is faster. As long as it isn't a error in measurement my guess would be that the streaming `anyMatch` implementation is highly optimized.
Another guess would be that the Java JIT compiler can handle and optimize a small lambda expression like the used predicate better than a code fragment from a larger method.

PR #992, opened 2018-07-17
check apk with magic and classes.dex but not the extension

gaojun0816:


src/main/java/soot/Scene.java
```diff
@@ -597,7 +598,7 @@ private String defaultAndroidClassPath() {
       String targetApk = "";
       Set<String> targetDexs = new HashSet<String>();
       for (String entry : classPathEntries) {
-        if (entry.toLowerCase().endsWith(".apk")) { // on Windows, file
+          if (isApk(entry)) { // on Windows, file
```
StevenArzt: This comment isn't relevant anymore, right? We don't deal with file names anymore after your change.

gaojun0816: Yes, you're right, I didn't take care about the comments. I'm going to remove it.

mbenz89:
Thanks for your contribution!

PR #996, opened 2018-07-18
Code refactoring in Dexpler and ToDex

MarcMil:
Code refactoring, allowing for more flexibility

### DIFF
--- a/src/main/java/soot/dexpler/DexAnnotation.java
+++ b/src/main/java/soot/dexpler/DexAnnotation.java
@@ -505,11 +505,13 @@ public class DexAnnotation {
           clazz.setOuterClass(SootResolver.v().makeClassRef(outerClass));
           assert clazz.getOuterClass() != clazz;
         }
+        // Do not add annotation tag
         return;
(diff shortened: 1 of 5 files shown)


src/main/java/soot/dexpler/DexAnnotation.java
```diff
@@ -483,25 +505,17 @@ public void add(AnnotationTag a, int visibility) {
...
         }
-
-        // EnclosingClass comes in pair with InnerClass.
-        // Those are generated from a single InnerClassTag,
-        // that is re-constructed only for the InnerClass Dalvik
-        // annotation.
-        continue;
-
-      } else if (atypes.equals("dalvik.annotation.EnclosingMethod")) {
+        return;
```
StevenArzt: Sometimes cases end with "break", sometimes with "return". Can we have a single concept here?

StevenArzt: Just comment the return with "do not add tag" or the like, that would be sufficient

PR #1000, opened 2018-07-24
WIP: Add support for invoke-polymorphic and invoke-custom in dex and fix some issues with normal invoke-dynamic

agorski3:
Explanation is in the commit messages

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -370,7 +370,7 @@
         <dependency>
             <groupId>org.smali</groupId>
             <artifactId>dexlib2</artifactId>
-            <version>2.2.2</version>
+            <version>2.2.4</version>
         </dependency>
         <dependency>
             <groupId>org.ow2.asm</groupId>
--- a/src/main/java/soot/asm/AsmMethodSource.java
+++ b/src/main/java/soot/asm/AsmMethodSource.java
@@ -194,6 +194,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.objectweb.asm.Handle;
+import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.tree.AbstractInsnNode;
 import org.objectweb.asm.tree.FieldInsnNode;
 import org.objectweb.asm.tree.IincInsnNode;
@@ -268,6 +269,7 @@ import soot.jimple.JimpleBody;
 import soot.jimple.LongConstant;
 import soot.jimple.LookupSwitchStmt;
 import soot.jimple.MethodHandle;
+import soot.jimple.MethodType;
 import soot.jimple.MonitorStmt;
 import soot.jimple.NewArrayExpr;
 import soot.jimple.NewMultiArrayExpr;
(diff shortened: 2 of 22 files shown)


agorski3:
Ok so apparently there are a few dependencies on changes I made to dexlib2 that I forgot about. This means these soot changes depend on the most up to date version of dexlib2 from the git repo (once he accepts my pull request). Not sure how you will work that though since everything is pulled from maven.

mbenz89:
Okay, please keep me updated on the status of your pr to dexlib2.

agorski3:
Dexlib2 has finally been updated to include my pull request but he has not released a new version yet with the changes. It is not clear when he will do that either. However, it is in the repo so it can be pulled and built manually.

mbenz89:
Thanks for letting me know! 
If this feature is requested by Soot users a lot, it might make sense to build and deploy dexlib2 our selves. Since this is currently not the case, I would rather wait for a stable release of dexlib2 before merging.

Thanks for your efforts, the pr looks quite nice!

agorski3:
No problem. I will let you know once he releases a new version.

StevenArzt:
@mbenz89 Can you please double-check whether the 2.2.5 release of Dexlib is new enough to finally accept this merge request? It has been a while since this MR was updated and we should integrate it before it gets fully outdated.

agorski3:
The 2.2.5 release of Dexlib 2 is available now and it has the changes I made to make my code here work. Sorry it was apparently released back at the beginning of Sep. I just have been busy with a seperate project and haven't checked since then.

StevenArzt:
That means we now need to get the MR fixed (most likely incompatibilities due to changes to the code base in the meantime) and are then ready to merge, right?

agorski3:
Yes it was ready for merging back when I made the pr. The dexlib2 version just needs to be bumped up to 2.2.5 and then whatever changes in the code base were made need to be accounted for.

StevenArzt:
@mbenz89 Can you have a look at this?

mbenz89:
@StevenArzt Sure, I'll finish this up on Monday.
@agorski3 Thanks for keeping us updated!

mbenz89:
I've raised the dexlib2 dependency to 2.2.5 and added some very basic test cases in #1047.

The PR looks good so far. However, since `invokeCustom` calls are transformed to Soot's `JDynamicInvokeExpr`, we will break when writing back out code using `invokeCustom`. This happens since we do not handle `invokeDynamic` for dex write-out at all:  https://github.com/Sable/soot/blob/develop/src/main/java/soot/toDex/ExprVisitor.java#L154

@StevenArzt Should we still merge this PR now and open a new issue for write-out support, or try to get this completed first?
@agorski3 Would you be able to implement such a write-out support? It should be sufficient to transform `invokdeDynamic` statments back to `invokeCustom`.

Regarding write-out of `invokePolymorphic` statments, as far as I undestand, we only loose the prototype when transforming these to `invokeVirtual`. Thus, using `invokeVirtual` instead of `invokePolymorphic` on write-out should still keep the code executable. Is that correct, @agorski3 ?

StevenArzt:
@mbenz89 At the moment, we do not support ```invokeCustom``` at all. Attempts to read in an apps that contains such code will either fail or deliver incomplete Jimple code. Neither possibility will result in a working app after writeout. Thus, we'll not make anything worse when adding support for reading in such constructs. So let's go ahead and add the write-out support later.

agorski3:
I second the idea of worrying about writing out invokeCustom later. It was not required for what I was doing (just need to read in dex code with invokePolymorphic or invokeCustom) so I have not written any code to write it out and don't really have time to tackle it at the moment. As for invokePolymorphic since it it transformed into an invokeVirtual you are correct in assuming that there should be no issue with writing this stmt back and still having the code execute.

mbenz89:
I agree. Thanks again for your contribution, @agorski3 .

mbenz89:
Merged by #1047

PR #1010, opened 2018-08-15
Moving insert on edge to patching chain, since it actually implies

MarcMil:
… that the chain is only handling units.


src/main/java/soot/PatchingChain.java
```diff
@@ -124,7 +135,136 @@ public void insertBeforeNoRedirect(E toInsert, E point) {
    *          the target point of an edge
    */
   public void insertOnEdge(E toInsert, E point_src, E point_tgt) {
-    innerChain.insertOnEdge(toInsert, point_src, point_tgt);
+    List<E> o = new ArrayList<E>();
```
StevenArzt: You can use Collection.singletonList() instead

src/main/java/soot/PatchingChain.java
```diff
@@ -124,7 +135,136 @@ public void insertBeforeNoRedirect(E toInsert, E point) {
...
+    if (validEdgeFound) {
+      innerChain.insertBefore(toInsert, point_tgt);
+
+      if (originalPred != point_src) {
+        if (originalPred instanceof GotoStmt) {
+          return;
+        }
+
+        Unit goto_unit = Jimple.v().newGotoStmt(point_tgt);
+        innerChain.insertBefore((List<E>) goto_unit, toInsert.iterator().next());
```
StevenArzt: I don't think it's a good idea to cast this

src/main/java/soot/PatchingChain.java
```diff
@@ -124,7 +135,136 @@ public void insertBeforeNoRedirect(E toInsert, E point) {
...
+    // In certain scenarios, the above code can add extra 'goto' units on a
+    // different edge
+    // So, an edge [src --> tgt] becomes [src -> goto tgt -> tgt].
+    // When this happens, the original edge [src -> tgt] ceases to exist.
+    // The following code handles such scenarios.
+    final Unit succ = getSuccOf(point_src);
+    if (succ instanceof GotoStmt) {
+      if (succ.getUnitBoxes().get(0).getUnit() == point_tgt) {
+
+        getSuccOf(point_src).redirectJumpsToThisTo(toInsert.iterator().next());
```
StevenArzt: Don't call getSuccOf() again, use succ instead

src/main/java/soot/PatchingChain.java
```diff
@@ -124,7 +135,136 @@ public void insertBeforeNoRedirect(E toInsert, E point) {
...
+
+        if (point_src instanceof GotoStmt) {
+
+          box.setUnit(toInsert.iterator().next());
+          innerChain.insertAfter(toInsert, point_src);
+
+          Unit goto_unit = Jimple.v().newGotoStmt(point_tgt);
+          if (toInsert instanceof List) {
+            List<Unit> l = (List<Unit>) toInsert;
+            innerChain.insertAfter((E) goto_unit, (E) l.get(l.size() - 1));
```
StevenArzt: Why do we need this cast?

src/main/java/soot/PatchingChain.java
```diff
@@ -124,7 +135,136 @@ public void insertBeforeNoRedirect(E toInsert, E point) {
...
+          } else {
+            innerChain.insertAfter((E) goto_unit, (E) toInsert.toArray()[toInsert.size() - 1]);
```
StevenArzt: Why do we need this cast?

src/main/java/soot/PatchingChain.java
```diff
@@ -139,7 +279,7 @@ public void insertOnEdge(E toInsert, E point_src, E point_tgt) {
    *          the target point of an edge
    */
   public void insertOnEdge(List<E> toInsert, E point_src, E point_tgt) {
-    innerChain.insertOnEdge(toInsert, point_src, point_tgt);
+    insertOnEdge((Collection<E>) toInsert, point_src, point_tgt);
```
StevenArzt: Do we really need this cast?

src/main/java/soot/PatchingChain.java
```diff
@@ -154,15 +294,17 @@ public void insertOnEdge(List<E> toInsert, E point_src, E point_tgt) {
    *          the target point of an edge
    */
   public void insertOnEdge(Chain<E> toInsert, E point_src, E point_tgt) {
-    innerChain.insertOnEdge(toInsert, point_src, point_tgt);
+    insertOnEdge((Collection<E>) toInsert, point_src, point_tgt);
```
StevenArzt: Do we really need this cast?

PR #1041, opened 2018-10-15
Add constant value tag even if the field is not final

MarcMil:


mbenz89:
Hi Marc.

There were several issues causing this. 
First of all, we bound the checkstyle plugin to Maven's `verify` phase at some point but missed to let the build jobs for the branches actually execute this phase. This is why the develop branch was not broken after merging the PR which introduced the style flaw (#1035).
Secondly, Steven seems to have missed allowing the checks to run for the PR in the first place, but merged it without the checks being able to report the violation. 

Anyway, thanks for reporting! It should be fixed now.

mbenz89:
retest this please

src/main/java/soot/toolkits/scalar/ConstantValueToInitializerTransformer.java
```diff
@@ -78,10 +78,10 @@ public void transformClass(SootClass sc) {
     Set<SootField> alreadyInitialized = new HashSet<SootField>();
 
     for (SootField sf : sc.getFields()) {
-      // We can only create an initializer for static final fields that
-      // have a constant value. We ignore non-static final fields as
+      // We can only create an initializer for static fields that
+      // have a constant value. We ignore non-static fields as
       // different constructors might assign different values.
```
mbenz89: Isn't this statement invalid now since basically any constructor or method could assign different values?

MarcMil: It's fine for a method to assign different values at runtime; in case of non-final fields it is just a default value.
In the previous implementation the initial value just got lost, since no tag was created albeit a default value was specified. Since the transformer puts the initialization at the beginning of the static initializer, it could even be overwritten within the static initializer itself without problems.

mbenz89: I see. Thanks for explaining.

Issue #1060, opened 2018-11-09
ThrowableSet.catchableAs(...) can be incorrect in case option no_bodies_for_excluded == true


jpstotz:
If you call `ThrowableSet.catchableAs(exceptionType)` using an exception that is not a basic class and not an application class the result will be incorrect.

Background: In case the Soot option `no_bodies_for_excluded == true` and `include_all == false` Soot will load most of the library classes (from rt.jar or android.jar) and mark them as phantom class if they come from a package that is excluded by default (see [Scene.determineExcludedPackages()](https://github.com/Sable/soot/blob/develop/src/main/java/soot/Scene.java#L106)).

The method `ThrowableSet.catchableAs(..)` checks if a class is phantom and if it is phantom it does not perform the type check which ends up in incorrect results. I assume the person who wrote the catchableAs method assumed that phantom classes do not have a valid type hierarchy and therefore can't be checked. 
However the "phantom library classes" do have a correct type hierarchy!

BTW: Alternatively there is the solution to say that the option `no_bodies_for_excluded` does not work correctly as it's name would suggest (or is there a documentation of this flag?) because it marks the whole class phantom instead of only marking the methods as phantom (which is what I would have expected by it's name). Also the default not-modifiable exclude list in `Scene.java` is IMHO not a very clean solution.

jpstotz:
> I am not quite sure how phantom classes are able to have a correct type hierarchy... How does Soot build the type hierarchy for classes with no class source?

For a real phantom class Soot can not build a class hierarchy (in such a case is `superClass` of the phanom class `null`?). However I assume the developer who have implemented this used it as a hack:

The following is partially an assumption based on the outcome I have seen and my general reverse engineering skill how developers usually think:

Situation: `no_bodies_for_excluded == true` and `include_all == false`

A library class that is excluded by the default list is loaded and resolved for the level `HIERARCHY`. To make sure Soot does not try to resolve the class for a higher like `SIGNATURES` or `BODIES` the class is marked phantom. Therefore we end up with classes that have hierarchy info but are marked as phantom.

mbenz89:
I see. I misread that this is about library classes... 

Could we just extend the check in `ThrowableSet.catchableAs(exceptionType)` to also check if the class has level `HIERARCHY` or better?

jpstotz:
That would be an option if a real phantom class would only have a resolved level of `DANGLING`. 

I am not that deep into the Soot phantom class generation to know if this is true or not.

[...]

PR #1073, opened 2018-11-26
fix: bugfix for #1060

jpstotz:
Instead of isPhantom() we now use !hasSuperclass() || "java.lang.Object".equals(sootClass.getName())

For further details please see issue #1060

### DIFF
--- a/src/main/java/soot/SootClass.java
+++ b/src/main/java/soot/SootClass.java
@@ -84,6 +84,8 @@ public class SootClass extends AbstractHost implements Numberable {
 
   public final static String INVOKEDYNAMIC_DUMMY_CLASS_NAME = "soot.dummy.InvokeDynamic";
 
+  public final static String JAVA_LANG_OBJECT = "java.lang.Object";
+  
   /**
    * Constructs an empty SootClass with the given name and modifiers.
    */
--- a/src/main/java/soot/toolkits/exceptions/ThrowableSet.java
+++ b/src/main/java/soot/toolkits/exceptions/ThrowableSet.java
@@ -40,6 +40,7 @@ import soot.RefLikeType;
 import soot.RefType;
 import soot.Scene;
 import soot.Singletons;
+import soot.SootClass;
 import soot.Unit;
 import soot.options.Options;
 
@@ -456,11 +457,12 @@ public class ThrowableSet {
       }
     }
     FastHierarchy hierarchy = Scene.v().getOrMakeFastHierarchy();
+    boolean eHasNoHierarchy = !(e.getSootClass().hasSuperclass() || //
+        SootClass.JAVA_LANG_OBJECT.equals(e.getSootClass().getName()));
 
     for (AnySubType excludedType : exceptionsExcluded) {
       RefType exclusionBase = excludedType.getBase();
-      if ((e.getSootClass().isPhantom() && exclusionBase.equals(e))
-          || (!e.getSootClass().isPhantom() && hierarchy.canStoreType(e, exclusionBase))) {
+      if ((eHasNoHierarchy && exclusionBase.equals(e)) || (!eHasNoHierarchy && hierarchy.canStoreType(e, exclusionBase))) {
         throw new AlreadyHasExclusionsException("ThrowableSet.add(RefType): adding" + e.toString() + " to the set [ "
             + this.toString() + "] where " + exclusionBase.toString() + " is excluded.");
       }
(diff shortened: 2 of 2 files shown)


src/main/java/soot/SootClass.java
```diff
@@ -84,6 +84,8 @@
 
   public final static String INVOKEDYNAMIC_DUMMY_CLASS_NAME = "soot.dummy.InvokeDynamic";
 
+  public final static String JAVA_LANG_OBJECT = "java.lang.Object";
```
mbenz89: There is a singleton for the Object type in the Scene which you could use: `Scene.v().getObjectType()`

src/main/java/soot/toolkits/exceptions/ThrowableSet.java
```diff
@@ -456,19 +457,20 @@ public ThrowableSet add(RefType e) throws ThrowableSet.AlreadyHasExclusionsExcep
       }
     }
     FastHierarchy hierarchy = Scene.v().getOrMakeFastHierarchy();
+    boolean eHasNoHierarchy = !(e.getSootClass().hasSuperclass() || //
+        SootClass.JAVA_LANG_OBJECT.equals(e.getSootClass().getName()));
```
mbenz89: Using the 'Object' type from Scene, you could cache the `SootClass` which you can derived from it and then do an object comparison `e.getSootClass() == OBJECT_CLASS`. This should be much faster then doing this string comparison and since `SootClass`should be managed by the `Scene`, you are guaranteed to only have one instance of the `Object` class

mbenz89: Btw. there is a `//` at the end of this line!

src/main/java/soot/toolkits/exceptions/ThrowableSet.java
```diff
@@ -543,6 +545,8 @@ public ThrowableSet add(AnySubType e) throws ThrowableSet.AlreadyHasExclusionsEx
 
     FastHierarchy hierarchy = Scene.v().getOrMakeFastHierarchy();
     RefType newBase = e.getBase();
+    boolean newBaseHasNoHierarchy = !(newBase.getSootClass().hasSuperclass() || //
+        SootClass.JAVA_LANG_OBJECT.equals(newBase.getSootClass().getName()));
```
mbenz89: same here

src/main/java/soot/toolkits/exceptions/ThrowableSet.java
```diff
@@ -830,13 +836,22 @@ public boolean catchableAs(RefType catcher) {
...
     FastHierarchy h = Scene.v().getOrMakeFastHierarchy();
+    /**
+     * Originally this implementation had checked if the catcher.getSootClass() is a phantom class. However this makes
+     * problems in case the soot option no_bodies_for_excluded==true because certain library classes will be marked as
+     * phantom classes even if they have a hierarchy. The workaround for this problem is to check for the suerClass. As every
+     * class except java.lang.Object have a superClass (even interfaces have!) only real phantom classes can be identified
+     * using this method.
+     */
+    boolean catcherHasNoHierarchy = !(catcher.getSootClass().hasSuperclass() || //
+        SootClass.JAVA_LANG_OBJECT.equals(catcher.getSootClass().getName()));
```
mbenz89: and here

src/main/java/soot/toolkits/exceptions/ThrowableSet.java
```diff
@@ -913,12 +928,14 @@ public Pair whichCatchableAs(RefType catcher) {
     if (INSTRUMENTING) {
       Manager.v().removesFromSearch++;
     }
+    boolean catcherHasNoHierarchy = !(catcher.getSootClass().hasSuperclass() || //
+        SootClass.JAVA_LANG_OBJECT.equals(catcher.getSootClass().getName()));
```
mbenz89: ..

mbenz89:
Looks fine to me except for the comments. Thanks for your efforts!

mbenz89:
retest this please

src/main/java/soot/toolkits/exceptions/ThrowableSet.java
mbenz89: You're welcome! 
I think it would make sense to cache the result of `Scene.v().getObjectType().getSootClass()` since the call still involves some unecessary overhead due to map lookup and sanity checks.
I guess it's debatable which version is more readable. Though, according to our styling conventions, operators should be placed after the line break in such cases: https://google.github.io/styleguide/javaguide.html#s4.5.1-line-wrapping-where-to-break

mbenz89:
Nevermind, this just instructs our CI to run the tests again. Sorry for the confusion!

PR #1075, opened 2018-11-30
Call graph generation optimized

jpstotz:
Note: This is a follow-up PR of #987 therefore I used the same branch. Originally I thought that it is possible to re-open the old PR but it looks GitHub does not work this way. Therefore I had to create this new pull request.

Again I had to modify the `FastHierarchy.canStoreClass(..)` because of serious performance issues:

For one app (com.google.android.googlequicksearchbox) the call graph generation took more than 4 hours (using the parallelization introduced by commit 98a8f9d of this PR on a server with 12 cores).

The problem is that the Google quicksearchbox app has one interface which is implemented ~8800 times (may be an protobuf related interface used by all protobuf generated classes?).
Having to check 8800 classes if their interval match a certain other class as implemented in `FastHierarchy.canStoreClass(..)` is a pretty CPU intensive task.

Therefore I implemented an different algorithm in the method `FastHierarchy.canStoreClassClasic(..)`. It uses the classic/intuitive approach: 

If you want to know if a certain child class is type compatible to a "parent" class just check the child class and all it's super classes if it matches the child class. If parent is an interface also check the interfaces (and their super interfaces).

Using this implementation the CHA call graph generation time of Google quicksearchbox reduces from 4 hours to 11 minutes on my test system. And of course the CPU utilization is much lower as it does not execute anything in parallel.

As before the new approach is only used in case we encounter an interface that has been implemented more than 1000 times. Therefore regular apps should not be affected in a negative way by this new implementation.

The correctness of the `FastHierarchy.canStoreClassClasic(..)` has been checked by "symbolic in-brain execution" as well as loading various apps using the old parallel approach and the new approach in parallel, comparing their results.


src/main/java/soot/FastHierarchy.java
```diff
@@ -346,6 +347,50 @@ public boolean canStoreClass(SootClass child, SootClass parent) {
...
+    SootClass sc = child;
+    final boolean parentIsInterface = parent.isInterface();
+    while (sc != null) {
+      if (sc.getName().equals(parent.getName())) {
+        // We finally found the correct class/interface
+        return true;
+      }
+      if (parentIsInterface) {
+        // Interfaces can only extend other interfaces - therefore we only have to consider the
+        // interfaces of a the child class if parent is an interface.
```
mbenz89: There is a typo 

> ...interfaces of **a** the child...

src/main/java/soot/FastHierarchy.java
```diff
@@ -346,6 +347,50 @@ public boolean canStoreClass(SootClass child, SootClass parent) {
...
+   * 
+   * @param child
+   * @param parent
+   * @return
+   */
+  protected boolean canStoreClassClassic(final SootClass child, final SootClass parent) {
+    SootClass sc = child;
+    final boolean parentIsInterface = parent.isInterface();
+    while (sc != null) {
+      if (sc.getName().equals(parent.getName())) {
```
mbenz89: Why are you comparing the names instead of using object equality on the `SootClasses` directly?
This maybe a bit safer but it's also quite more time consuming to use the String equality here. 
All classes have to be managed by the Scene and are thus guaranteed to be unique. IMHO, if this assumption does not hold, there's an issue elsewhere and should be fixed there. This would  BTW also break many other components of Soot.

jpstotz: I took over the class comparison via class name from `RefType.equals()`. Another reason may be hat as an experienced Java developer I encountered dozens of bugs introduced by the use of `==` (on Strings of course which act totally different than a SootClass).

I will change it to an equality and do some more tests.

jpstotz:
Another example is the Facebook Pages app. The overall processing time of this app changes from ~20 minutes to ~10 minutes with this PR. I have not tested any more apps regarding their speed-up.

src/main/java/soot/FastHierarchy.java
mbenz89: I'm also a huge fan of defensive programming and would, in most cases, also rather tend to use the safer variant of comparing the class names (I guess this is what happens in `RefType.equals()`). However, I think that we are rather going for performance here and should be able to rely on assumptions made across the whole Soot codebase. 
In the end, I can also live with using the safer variant if we the performance penalty compares to the gained safety.

jpstotz: I changed the class comparison as you suggested. I did not notice any negative side effects...

mbenz89: Thanks!

PR #1077, opened 2018-12-05
soot.FastHierarchy was not thread safe

jpstotz:
I noticed that the two methods `getAllImplementersOfInterface()` and `getAllSubinterfaces()` in soot.FastHierarchy are not safe for multithreading as they store intermediate results in their caches.

Especially the problem of getAllSubinterfaces() was significant so that I could trigger it using a unit test (and adding a small delay). AFAIK both methods are indirectly called while building e.g. a Jimple body of a method (which is used concurrently) thread safety is required.

I changed the implementations so that they now should work thread safe.


mbenz89:
Thanks for your contribution!

PR #1081, opened 2018-12-19
jbco improvements

pnest:
MethodRenamer
---
This transformer was handling _indirect_ inheritance partially: transformer was not able to find interfaces that were extending interfaces from library classes (it is interface A in javadoc example).

FieldRenamer
---
I hid new name generation algorithm and removed storage of renamed fields. Since now you don't have to provide old name (if there were no one) or save field that were renamed.

SootMethodRefImpl
---
I ran into a bug when soot was resolving methods after applying MethodRenamer. Since signature was changed soot was falling with error ResolutionFailedException. Now method is searching using `soot.SootClass#getMethodUnsafe(java.lang.String, java.util.List<soot.Type>, soot.Type)` instead of `soot.SootClass.getMethodUnsafe(soot.util.NumberedString)`.


pnest:
Resolved merge conflicts

src/main/java/soot/PackManager.java
```diff
@@ -657,14 +657,8 @@ private void runBodyPacks(final Iterator<SootClass> classes) {
...
-      executor.execute(new Runnable() {
-
-        @Override
-        public void run() {
-          runBodyPacks(c);
-        }
-
-      });
+      executor.execute(() -> runBodyPacks(c));
+      // runBodyPacks(c);
```
mbenz89: Please remove the comment

mbenz89:
Thanks for your contribution!

PR #1105, opened 2019-01-31
 Fix array set when initialized with length 0

MarcMil:
When initialized with an length of 0, it did not grow, but threw an index out of bounds since 2*0 = 0

### DIFF
--- a/src/main/java/soot/LambdaMetaFactory.java
+++ b/src/main/java/soot/LambdaMetaFactory.java
@@ -232,10 +232,10 @@ public final class LambdaMetaFactory {
     if (enclosingClass.isApplicationClass()) {
       tclass.setApplicationClass();
     }
-    
+
     for (SootMethod m : tclass.getMethods()) {
       // There is no reason not to load the bodies directly. After all,
-      // we are introducing new classes while loading bodies. 
+      // we are introducing new classes while loading bodies.
       m.retrieveActiveBody();
     }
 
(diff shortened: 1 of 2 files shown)


DvonHolten:
Hi there,

 

i just had a very brief look at the diff of that checkin.

Are you really sure the second toArray() - method is correct?

In Java-world, the incoming typed array is just a template for the array-element-class.

You cannot rely on the size to be suitable to hold the elements of the collection.

HTH

DvH

 

 

Von: Steven Arzt [mailto:notifications@github.com] 
Gesendet: Donnerstag, 31. Januar 2019 18:38
An: Sable/soot <soot@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
Betreff: Re: [Sable/soot] Fix array set when initialized with length 0 (#1105)

 

Merged #1105 <https://github.com/Sable/soot/pull/1105>  into develop.

StevenArzt:
Which line do you reference here? The diff doesn't introduce any accesses to the size of the incoming array,  but rather uses the ```numElements``` field, if I'm not mistaken. In fact, the ```toArray()``` methods weren't changed at all.

DvonHolten:
lines 206, 207, 208 - maybe not changed - but looks wrong. As i said, in the corresponding method in JRE, the incoming array is just a template for the type. I routinely pass arrays of size 0.

StevenArzt:
I see the problem. Thanks a lot for pointing this out. 123ede4df8ca02312ce056c02c0396e05dcdda73 should fix the problem.

PR #1115, opened 2019-02-13
Make the fresh-local-variable separator configurable.

gfour:
Having "_" as a delimiter for fresh locals created internally by Soot may be confusing when trying to connect Shimple locals with original Java sources. For example, if Soot creates "x_3" for some "x" and for some reason "x_3" was also a source variable, whose name was not preserved, the fresh local may be mistaken for the source local.

This pull request makes the `"_"` separator configurable, so that it can be set to something that cannot appear in the sources (e.g., `'_$_'`).


mbenz89:
add to whitelist

mbenz89:
Thanks for your contribution!

Issue #1083, opened 2018-12-21
How to reproduce synchronization issues mentioned in #1078


jonmathews-ensoft:
> Would be great if you tell me how to reproduce the stated exceptions still.

_Originally posted by @mbenz89 in https://github.com/Sable/soot/pull/1078#issuecomment-448972688_

Here are instructions for reproducing the synchronization issues mentioned in the recently merged #1078.  The issues were mitigated by adding 'synchronized' to LambdaMetaFactory.makeLambdaHelper().  I suspect that several data structures in Scene need to be synchronized, at which point makeLambdaHelper probably does not need to be synchronized itself.

To reproduce, start by removing 'synchronized' from makeLambdaHelper.

Then, convert the class files in the attached Eclipse project to Jimple text.  The source code is included for reference.  In order to increase the time in the critical section I had to make copies of two classes.  It still does not throw exceptions every time, but if you run it ten times you should see both kinds of stack traces (also attached).

I included the Eclipse launch profile, which has the following arguments:

--java-version 1.8 --f J --p jb stabilize-local-names:true --d ${project_loc:lmf-concurrency}/j1 --cp ${project_loc:lmf-concurrency}/bin:${soot_jdk8_classpath} --process-dir ${project_loc:lmf-concurrency}/bin

Note that ${soot_jdk8_classpath} is the list of jdk8 jars for your machine.  You can either replace the variable or define it.

[lmf-concurrency.zip](https://github.com/Sable/soot/files/2703398/lmf-concurrency.zip)
[stack1.txt](https://github.com/Sable/soot/files/2703403/stack1.txt)
[stack2.txt](https://github.com/Sable/soot/files/2703404/stack2.txt)

mbenz89:
fixed by https://github.com/Sable/soot/pull/1191

Issue #1067, opened 2018-11-15
ERROR heros.solver.CountingThreadPoolExecutor - Worker thread execution failed


zhou-shulin:
I am using soot to do some static analysis on Java program. I wrote an analysis transformer in 'wjtp" phase that only read information from jimple code, nothing is changed. But when I ran the transformer on HBase-2.1.1 (including hbase-common-2.1.1.jar, hbase-server-2.1.1.jar and hbase-client-2.1.1.jar), here comes a problem as follows (The detailed stack information is in [attachment-1.txt](https://github.com/Sable/soot/files/2584106/attachment-1.txt)):
```
[Thread-6] ERROR heros.solver.CountingThreadPoolExecutor - Worker thread execution failed: Trying to create interface invoke expression for non-interface type: org.apache.hadoop.hbase.regionserver.MetricsRegionServerSourceFactory Use JVirtualInvokeExpr or JSpecialInvokeExpr instead!
java.lang.RuntimeException: Trying to create interface invoke expression for non-interface type: org.apache.hadoop.hbase.regionserver.MetricsRegionServerSourceFactory Use JVirtualInvokeExpr or JSpecialInvokeExpr instead!
	at soot.jimple.internal.JInterfaceInvokeExpr.<init>(JInterfaceInvokeExpr.java:51)
	at soot.jimple.internal.JInterfaceInvokeExpr.clone(JInterfaceInvokeExpr.java:67)
	at soot.jimple.Jimple.cloneIfNecessary(Jimple.java:144)
	at soot.jimple.internal.JAssignStmt.clone(JAssignStmt.java:221)
	at soot.Body.importBodyContentsFrom(Body.java:165)
	at soot.jimple.JimpleBody.clone(JimpleBody.java:87)
	at soot.PackManager.convertJimpleBodyToBaf(PackManager.java:1030)
	at soot.PackManager.runBodyPacks(PackManager.java:991)
	at soot.PackManager.access$000(PackManager.java:112)
	at soot.PackManager$1.run(PackManager.java:620)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

[...]

PR #1116, opened 2019-02-15
ConcurrentModificationException in SootClass.getMethodUnsafe

jpstotz:
Using the latest Soot version I often encountered with certain apps a ConcurrentModificationException in 
`public SootMethod getMethodUnsafe(String name, List<Type> parameterTypes, Type returnType)`

This pull request fixes that problem.
Additionally it includes a small non-related fix for a missing exception message in `soot.jimple.toolkits.typing.integer.InternalTypingException` which was always null instead of the actual message.

### DIFF
--- a/src/main/java/soot/SootClass.java
+++ b/src/main/java/soot/SootClass.java
@@ -548,7 +548,7 @@ public class SootClass extends AbstractHost implements Numberable {
       return null;
     }
 
-    for (SootMethod method : methodList) {
+    for (SootMethod method : new ArrayList<>(methodList)) {
       if (method.getName().equals(name) && parameterTypes.equals(method.getParameterTypes())
           && returnType.equals(method.getReturnType())) {
         return method;
--- a/src/main/java/soot/jimple/toolkits/typing/integer/ClassHierarchy.java
+++ b/src/main/java/soot/jimple/toolkits/typing/integer/ClassHierarchy.java
@@ -149,7 +149,7 @@ public class ClassHierarchy {
     TypeNode typeNode = typeNodeMap.get(type);
 
     if (typeNode == null) {
-      throw new InternalTypingException();
+      throw new InternalTypingException(type);
     }
 
     return typeNode;
--- a/src/main/java/soot/jimple/toolkits/typing/integer/InternalTypingException.java
+++ b/src/main/java/soot/jimple/toolkits/typing/integer/InternalTypingException.java
@@ -44,8 +44,9 @@ class InternalTypingException extends RuntimeException {
   }
 
   @Override
-  public String toString() {
-    return "Unexpected type " + this.unexpectedType;
+  public String getMessage() {
+    return String.format("Unexpected type %s (%s)", unexpectedType,
+        unexpectedType == null ? "-" : unexpectedType.getClass().getSimpleName());
   }
 
 }


src/main/java/soot/SootClass.java
```diff
@@ -548,7 +548,7 @@ public SootMethod getMethodUnsafe(String name, List<Type> parameterTypes, Type r
       return null;
     }
 
-    for (SootMethod method : methodList) {
+    for (SootMethod method : new ArrayList<>(methodList)) {
```
StevenArzt: This is not a real fix, because the copy constructor of the ArrayList also iterates over the contents of the original list and copies over all elements. Therefore, if the list is concurrently modified by another thread, this issue might as well strike while the copy constructor is running. The much better question would be why the method is concurrently modified in the first place

jpstotz: Have you ever encountered a ConcurrentModificationException in the new `ArrayList(Collection)` constructor? Or are you referring to that the ´methodList` variable is not synchronized?

I have not investigated why the method is concurrently modified, however I assume that this happens because of necessary phantom method references being added as they are used by the current SootClass instance or any other SootClass that is processed at that time in a different thread. I don't see a chance to synchronize the whole process.

StevenArzt: Apparently, the copy constructor in newer JDKs doesn't iterate, but uses ```System.arrayCopy``` deep down. So you're technically right that you shouldn't get a ```ConcurrentModificationException``` anymore, but you might end up with all sorts of funny effects if a resizing happens while you are copying, because ArrayLists have a distinct field for their size, which they use when accessing the internal array. I think it's safe to say that your fix greatly reduces the risk of concurrency issues, but it's still a bit shady. I'm still not sure why aother thread manipulates methods concurrently here. Do you have a stack trace?

jpstotz: This is one example stack trace I found:
```
java.util.ConcurrentModificationException: null
	at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1042) ~[na:na]
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:996) ~[na:na]
	at soot.SootClass.getMethodUnsafe(SootClass.java:553) 
	at soot.SootMethodRefImpl.tryResolve(SootMethodRefImpl.java:210) 
	at soot.SootMethodRefImpl.tryResolve(SootMethodRefImpl.java:188) 
	at soot.toolkits.exceptions.UnitThrowAnalysis.mightThrow(UnitThrowAnalysis.java:297) 
	at soot.toolkits.exceptions.UnitThrowAnalysis$ValueSwitch.caseInstanceInvokeExpr(UnitThrowAnalysis.java:1120) 
	at soot.toolkits.exceptions.UnitThrowAnalysis$ValueSwitch.caseVirtualInvokeExpr(UnitThrowAnalysis.java:970) 
	at soot.jimple.internal.AbstractVirtualInvokeExpr.apply(AbstractVirtualInvokeExpr.java:79) 
	at soot.toolkits.exceptions.UnitThrowAnalysis.mightThrow(UnitThrowAnalysis.java:289) 
	at soot.toolkits.exceptions.UnitThrowAnalysis$UnitSwitch.caseAssignStmt(UnitThrowAnalysis.java:751) 
	at soot.jimple.internal.JAssignStmt.apply(JAssignStmt.java:242) 
	at soot.toolkits.exceptions.UnitThrowAnalysis.mightThrow(UnitThrowAnalysis.java:275) 
	at soot.toolkits.graph.ExceptionalUnitGraph.buildExceptionDests(ExceptionalUnitGraph.java:287) 
	at soot.toolkits.graph.ExceptionalUnitGraph.initialize(ExceptionalUnitGraph.java:229) 
	at soot.toolkits.graph.ExceptionalUnitGraph.<init>(ExceptionalUnitGraph.java:135) 
	at soot.toolkits.graph.ExceptionalUnitGraph.<init>(ExceptionalUnitGraph.java:163) 
	at soot.toolkits.scalar.LocalDefs$Factory.newLocalDefs(LocalDefs.java:65) 
	at soot.toolkits.scalar.LocalDefs$Factory.newLocalDefs(LocalDefs.java:51) 
	at soot.jimple.toolkits.scalar.DeadAssignmentEliminator.internalTransform(DeadAssignmentEliminator.java:239) 
	at soot.BodyTransformer.transform(BodyTransformer.java:55) 
	at soot.BodyTransformer.transform(BodyTransformer.java:59) 
	at soot.dexpler.DexBody.jimplify(DexBody.java:644) 
	at soot.dexpler.DexMethod$1.getBody(DexMethod.java:119) 
	at soot.SootMethod.retrieveActiveBody(SootMethod.java:402) 
```
It occurred when processing `<cjh: void <init>(cja)>` of `com.starfinanz.smob.android.bwmobilbanking` (one of several apps that caused this error).

I think more interesting would be where the call that modified the methodList came from, but that is not that simple to get.

StevenArzt: Where do you call ```retrieveActiveBody()```? Does that call stem from user code or from Soot code? Is that call issued in threaded code? I'm still trying to figure out where the multiple threads come from, i.e., why the method list gets changed while someone iterates over it.

jpstotz: The call comes from multi-threaded user code. After loading an app in soot we process every method of every class. According to some info I found (can't find it now) this was allowed in parallel.

StevenArzt: That's the issue then. Using Soot in any sort of multi-threaded environment is not supported to the best of my knowledge.

If you want to have a parallel version of Soot, you need to change quite a few things. I'd generally not recommend it. We have worked on the matter, but we needed to introduce semantics that are non-standard. Good enough for what we needed to do, but by no means a drop-in replacement for standard Soot.

mbenz89:
Might be related to #1083

mbenz89:
Might be related to #1067

src/main/java/soot/SootClass.java
mbenz89: How about synchronizing the `methodList` when creating a copy and using Jan's approach as a best-effort solution for now? This should prevent the resize while copy issue in `System.arrayCopy` stated by Steven and circumvent the witnessed `ConcurrentModificationException`.

mbenz89: @jpstotz @StevenArzt Can we come to a conclusion how to (not) handle this issue for now? 
What do you think about my proposal?

StevenArzt: The question is whether we want to give any concurrency guarantees or not. I'd rather not, because there are tons of other places that are likely to fail as well. As I said, we created a concurrent version of Soot, but had to change the semantics of many operations. The only alternative would be to synchronize all accesses to internal collections as suggested by @mbenz89 , but that would pretty much lead to a single-threaded version again.

mbenz89: I just fixed another issue that is caused by running the jimple body creation in parallel. There, the fast-hierarchy was not proper synchronized during jimple body creation.
@jpstotz is right here, [PackManager](https://github.com/Sable/soot/blob/master/src/main/java/soot/PackManager.java#L1262) does the exact parallel execution as he proposes.

I agree that synchronizing every field will ultimately lead to a kind-of single-threaded version again. However, the above-mentioned error seems to rise rather frequently, as it was reported by others as well. 

I guess we have to options here:
1. Disable the parallel jimplification at all, which will lead to huge performance decreases.
2. At least synchronize the method list and see how this plays out, i.e. see which other race conditions we are witnessing.

StevenArzt: Then let's do it like that, i.e., synchronize the method. It seems that we have no better option at the moment.

mbenz89: Like using this copy-iteration approach proposed by Jan? And synchronizing the method list during the copy operation? Or are you saying we should synchronize the whole `methodList`?

BTW one of my students just found another synchronization issue, where multiple instances of the same class could be added to the Scene and have different resolving levels (causing #1078 and more). This can be omitted if we synchronize all operations that add/remove/check classes in Scene -.- . It's still far from single-threaded then, but it will definitely worsen performance quite a bit during jimplification, i guess.

mbenz89:
Might be related to #1162

xzf1234:
I found a similar error when processing the apk with soot
[theakki.synctool.zip](https://github.com/soot-oss/soot/files/7660808/theakki.synctool.zip)

the call stack is:
```
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:911)
	at java.util.ArrayList$Itr.next(ArrayList.java:861)
	at soot.FastHierarchy.getSignaturePolymorphicMethod(FastHierarchy.java:921)
	at soot.FastHierarchy.resolveMethod(FastHierarchy.java:822)
	at soot.FastHierarchy.resolveMethod(FastHierarchy.java:774)
	at soot.SootMethodRefImpl.tryResolve(SootMethodRefImpl.java:228)
	at soot.SootMethodRefImpl.tryResolve(SootMethodRefImpl.java:214)
	at soot.toolkits.exceptions.UnitThrowAnalysis.mightThrow(UnitThrowAnalysis.java:304)
	at soot.toolkits.exceptions.UnitThrowAnalysis$ValueSwitch.caseInstanceInvokeExpr(UnitThrowAnalysis.java:1181)
	at soot.toolkits.exceptions.UnitThrowAnalysis$ValueSwitch.caseInterfaceInvokeExpr(UnitThrowAnalysis.java:998)
	at soot.jimple.internal.AbstractInterfaceInvokeExpr.apply(AbstractInterfaceInvokeExpr.java:121)
	at soot.toolkits.exceptions.UnitThrowAnalysis.mightThrow(UnitThrowAnalysis.java:296)
	at soot.dexpler.DalvikThrowAnalysis.access$600(DalvikThrowAnalysis.java:155)
	at soot.dexpler.DalvikThrowAnalysis$1.caseAssignStmt(DalvikThrowAnalysis.java:228)
	at soot.jimple.internal.JAssignStmt.apply(JAssignStmt.java:209)
	at soot.toolkits.exceptions.UnitThrowAnalysis.mightThrow(UnitThrowAnalysis.java:280)
	at soot.toolkits.exceptions.UnitThrowAnalysis.mightThrow(UnitThrowAnalysis.java:275)
	at soot.toolkits.graph.ExceptionalUnitGraph.buildExceptionDests(ExceptionalUnitGraph.java:277)
	at soot.toolkits.graph.ExceptionalUnitGraph.initialize(ExceptionalUnitGraph.java:220)
	at soot.toolkits.graph.ExceptionalUnitGraph.<init>(ExceptionalUnitGraph.java:127)
	at soot.jimple.toolkits.scalar.UnreachableCodeEliminator.internalTransform(UnreachableCodeEliminator.java:86)
	at soot.BodyTransformer.transform(BodyTransformer.java:52)
	at soot.BodyTransformer.transform(BodyTransformer.java:56)
	at soot.dexpler.DexBody.jimplify(DexBody.java:697)
	at soot.dexpler.DexMethod$1.getBody(DexMethod.java:120)
	at soot.PackManager.lambda$retrieveAllBodies$2(PackManager.java:1252)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```
I fixed the bug by adding new ArrayList<> in soot.FastHierarchy.getSignaturePolymorphicMethod like:
```
@@ -929,7 +918,7 @@
   private SootMethod getSignaturePolymorphicMethod(SootClass concreteType, String name, List<Type> parameterTypes,
       Type returnType) {
     SootMethod candidate = null;
-    for (SootMethod method : concreteType.getMethods()) {
+    for (SootMethod method : new ArrayList<>( concreteType.getMethods())) {
```

Issue #1057, opened 2018-11-06
java.lang.IndexOutOfBoundsException when Instrumenting 


yxycsz:
Hi,

Thanks for providing this amazing framework. I am trying to develop an instrumenting tool of android apps. 
However, I got java.lang.IndexOutOfBoundsException when processing an apk.
Here are the steps to reproduce this error.

soot: soot-3.1.0-20181031.115041-85-jar-with-dependencies.jar

target: wechat-6-7-3.apk
please download form:
https://wechat.en.uptodown.com/android
or
https://www.dropbox.com/s/57tpj2grro5ksd8/wechat-6-7-3.apk?dl=0

Command line:
```
java -jar soot-3.1.0-20181031.115041-85-jar-with-dependencies.jar -src-prec apk --allow-phantom-refs -output-format dex -force-android-jar /…/Android/Sdk/platforms/android-28/android.jar --process-dir /…/wechat-6-7-3.apk -process-multiple-dex
```

Stack trace:
```
[Thread-17] ERROR heros.solver.CountingThreadPoolExecutor - Worker thread execution failed: Index: 2140, Size: 2140
java.lang.IndexOutOfBoundsException: Index: 2140, Size: 2140
	at java.util.ArrayList.rangeCheck(ArrayList.java:653)
	at java.util.ArrayList.get(ArrayList.java:429)
	at soot.toDex.StmtVisitor.reduceInstructions(StmtVisitor.java:285)
	at soot.toDex.StmtVisitor.finalizeInstructions(StmtVisitor.java:228)
	at soot.toDex.DexPrinter.toInstructions(DexPrinter.java:1513)
	at soot.toDex.DexPrinter.toMethodImplementation(DexPrinter.java:1174)
	at soot.toDex.DexPrinter.toMethods(DexPrinter.java:1083)
	at soot.toDex.DexPrinter.addAsClassDefItem(DexPrinter.java:656)
	at soot.toDex.DexPrinter.add(DexPrinter.java:1646)
	at soot.PackManager.writeClass(PackManager.java:1110)
	at soot.PackManager$2.run(PackManager.java:710)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
[...]

PR #1135, opened 2019-03-26
Fix index out of bounds

MarcMil:


mbenz89:
Are you sure this is the correct fix?

I'm not a 100% sure what exactly is mapped here but I find it a bit strange that it is first checked that nextIndex does not point to the last instruction:

```
// Do not remove the last instruction in the body as we need 
 // remap
 // jump targets to the successor
if (nextIndex == this.insns.size() - 1) {
   continue;
 }
```

... and afterwards, one instruction is removed from the instructions list at `nextIndex`, which effectively sets the following element to `nextIndex`, and then the length of the instruction list is checked again so the element after the following one can be mapped.

```
 if (nextStmt == null || (!isJumpTarget(nextStmt) && !trapReferences.contains(nextStmt))) {
          insns.remove(nextIndex);

          if (nextStmt != null) {
            if (nextIndex == this.insns.size() - 1) {
              // we are now at the end of the list
              continue;
            }
            Insn nextInst = this.insns.get(nextIndex + 1);
            insnStmtMap.remove(nextInsn);
            insnStmtMap.put(nextInst, nextStmt);
          }
        }
      }
````

I thought, it might be correct to have `Insn nextInst = this.insns.get(nextIndex);` instead of `Insn nextInst = this.insns.get(nextIndex + 1);`. This would also prevent the indexOutOfBound and, also, it would actually map the __next__ instruction to `nextStmt` and not the one __after the next__.

Again, I am not sure I get the logic of this code. But it seems strange to me to call `remove` first and then still work with the incremented index.

Let me know if I'm wrong or that was too unclear..

mbenz89:
Solution for #1057

PR #1155, opened 2019-05-15
Merge defaul Soot and Java 9 Branch

anddann:
In the long run, we should merge the java 9 branch into the default Soot develop branch to handle JDK9, 10, and 11 in the default Soot implementation.

The purpose of this PR is to discuss improvements for the Java9 branch before merging it into develop


README.md
```diff
@@ -1,62 +1,123 @@
...
+        for (String module : map.keySet()) {
+            for (String klass : map.get(module)) {
+                logger.info("Loaded Class: " + klass + "\n");
+                loadClass(klass, false, module);
+
+            }
+    }
+
+
+    //this must be called after all classes are resolved
```
mbenz89: Do you mean "loaded" instead of "resolved"? Pleease fix if so, it might get confusing with the resolving process otherwies.

StevenArzt:
@mbenz89 Is there anything that's holding us back from merging this PR? If this PR doesn't make anything worse, I'd strongly suggest to merge it even in case it's not a fully-featured silver bullet solution yet. We're seeing more and more Java 9+ code out there.

mbenz89:
Nothing other than me not having found time reviewing it. I've scheduled some time next week. Depending on Andrea's availability, we will be able to finally merge it next week.

Jenkinsfile
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+
+
+
+
+		stage('Deploy'){
+		    when {
+			    branch 'master'
+			}
```
mbenz89: We still have to add teh deployment process to maven central for master and develop as well!

Also, shouldn't this file also handle the license and style check? If yes, we should add it as well.
Btw. I want to change the style procedure from checkstyle to just using the maven plugin for auto format and the check to be consistent here and don't go through the additional configuration hustle...

anddann:
Yes you are right Manuel.
However, honestly I don't have a clue how to inject the credentials for deployment into the Jenkins file. Nevertheless, I know Stefan tried sth. Like that, but I don't know if he succeed or failed. 
Would be awesome if you can ask him. I guess Stefan would be supper happy to contribute to Soot.

src/main/java/soot/asm/SootModuleInfoBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+  public SootModuleInfoBuilder(String name, SootModuleInfo klass, SootClassBuilder scb) {
+    super(Opcodes.ASM7);
+    this.klass = klass;
+    this.name = name;
+    this.scb = scb;
+  }
+
+  @Override
+  public void visitRequire(String module, int access, String version) {
+    // SootClass moduleInfo = SootResolver.v().resolveClass("module-info", SootClass.SIGNATURES, Optional.of(module));
```
mbenz89: Remove

StevenArzt: I just removed that line.

StevenArzt:
The only thing left to do seems to be the deployment descriptor. Can we do this quickly? I would like to merge this PR, because Java 9+ is now getting used in a lot of places. I'd even go without that, and re-use whatever method of deployment we have right now, because already have some means to get our artifacts out, and that discussion is unrelated to Java 9.

mbenz89:
Well, I'm not a fan of merging this before we got proper code quality/integration. Looking at past changes to Soot, this will probably never get sanitized after we merge this. IMHO Soot's codebase got kind of dirty due to that. 

Anyway, I see your demand and I'm thus fine with merging this. Using the current deployment mechanism will, unfortunately, not work here. Andreas uses multiple Docker containers with different Java versions to run the tests for the different Java versions. We need to either propagate build status, i.e. if the tests have passed, to our current build job and trigger the deployment thereby, or, better, integrate the deployment process into the Jenkinsfile.

We are currently out of time to implement that. If one of you guys could have a look and add stages for style/license check and deployment, we would certainly get this done faster.

Btw. why can't you use the artifacts from our nexus repo? There should be one for this branch online.

StevenArzt:
We don't have free capacity to help out with the integration either.

We would have one build that is being tested on one Java version if we merge the build right now and do nothing about integration, right? We (SIT) have a build server that's currently running Java 11 and that's running a Soot nightly build (including style and license check etc.). That would be one test. You are also running a CI pipeline, likely with a different Java version, which would give us a second test. That wouldn't be any worse than what we have today. Of course, testing Soot with multiple Java versions would be nice, but we don't have that right now either. Therefore, we're not making anything worse by going ahead.

So what's the plan now? I could also switch to the Java 9 branch with our projects, effectively making that branch productive for these projects. We would then end up with two current development branches, once we contribute code / fix bugs.

StevenArzt:
I have now had a deeper look at the code and I know why you are hesistant. As unfortunate as it is, I take back my suggestion to merge this branch. It definitely needs work. I'll see how I can continue my projects without Java 9 support for now, because that code is definitely not up to a reasonable standard. I was hoping for a drop-in replacement that is able to run on Java 9/11/13 with the basic classes inside modules and the user code in normal classes, but ran into quise some issues with that combination.

anddann:
Do you mind giving more insights into the issues?
Since the combination Java base classes in modules and user code on the  classpath works at least in a few different projects?

StevenArzt:
I have a Java 11 runtime with the basic classes inside the ``java.base.jmod`` file. The process-dir references normal Java JARs. This will first lead to an exception because of the missing ``rt.jar``. If you comment out this exception, the basic classes are not found. You can reproduce the issue by simply running any test case, e.g., ``MethodHandleTest`` on Java 11.

StevenArzt:
Speaking of code standard: We have classes that extend other classes and contain copies (with no changes) of methods from the base class, which should be removed (see class ``ModulePathSourceLocator``). Some implementations contain copy&paste code along with changes, that could be removed by refactoring (put extra code into separate method and call super method for the shared stuff). Further, the module path locator uses some field ``next`` as a loop iteration variable for some local operation, which is a fairly non-standard design choice. There's quite a bit of that in the code.

anddann:
I see that you have some points there. Yes, there some refactoring needed.
And you are absolutely right, the ```next``` variable is ugly. I guess I accidently implemented it to simulate the official code http://hg.openjdk.java.net/jdk9/jdk9/jdk/file/73113c19a5df/src/java.base/share/classes/jdk/internal/module/ModulePath.java closely.
Nevertheless, we should change it for Soot.

StevenArzt:
For the sake of completeness, here's the stack trace of the issue with the test case:

```
java.lang.RuntimeException: Error: cannot find rt.jar.
	at soot.Scene.defaultJavaClassPath(Scene.java:735)
	at soot.Scene.defaultClassPath(Scene.java:586)
	at soot.Scene.getSootClassPath(Scene.java:313)
	at soot.SourceLocator.getClassSource(SourceLocator.java:180)
	at soot.Scene.tryLoadClass(Scene.java:924)
	at soot.Scene.loadBasicClasses(Scene.java:1680)
	at soot.Scene.loadNecessaryClasses(Scene.java:1779)
	at soot.Main.run(Main.java:241)
	at soot.Main.main(Main.java:141)
	at soot.jimple.MethodHandleTest.testConstant(MethodHandleTest.java:85)
```

Run on OpenJDK 11 / Ubuntu x64.

AnakinSklavenwalker:
I have encountered the the same and other issues and i just discussed with Andreas the options. 

One of the main problems is that Soot at the moment requires both the executing environment and the target code to be on the same Java Version (or at least being together either on Java <9 XOR >= 9). 

Missmatching versions result in that soot is a) unable to find a proper Java installation that matches the target's version and b) in the case the soot environment is Java < 9 AND the target is >= 9, Soot is unable to handle the "jrt://" requests because it can not mount the \lib\modules as a filesystem and thus cannot read its content. 

The solution we came up with however requires quite a lot of additional work, as we need to mount the jrt-FS on demand. 
The idea was to use the `Options.v().set_soot_classpath("...")` and support passing a custom Java installation location here. Internally soot then shall decide how to correctly handle Java version management.

StevenArzt:
I am totally fine with only supporting Java 9+ classes when running on a Java 9+ VM. However, that needs to work seamlessly, i.e., have correct default classpaths for finding the basic classes on Java 9+, having clean code, etc.

I'm not really sure why the current implementation has a dedicated "module mode" either as a module is just another file format that is supported for Soot classpath and process-dir. Even when running on Java 9+, you have the basic classes in modules, and other code as JARs. When running on Java 8, you have no modules on the CP, and all of the code in JARs / class files.

I therefore suggest the following:

* Put modules on the normal classpath / process-dir. Do not use a separate module-path.
* Detect whether we're on Java 9+. In that case, put the base modules on the default CP. Otherwise, put rt.jar on the default CP. That should make Soot work on either platform without the user specifying anything (and would fix the failing test case)
* Generally merge the module-based stuff the normal code. In many cases, there is no need to switching between implementations, it just makes things more difficult.
* Have an option for class name compatibility: module-style or plain-style. In moduke-style use ``module:classname``, in plain style use the class name as before. That should be the only place where we really need an option for backwards compatibility, because we're making a breaking change.

anddann:
I see your point, however modules are a bit more than just another file format.
The class resolution process differs between the classpath and the modulepath on the JVM:
For instance, you are looking for a classs ```org.mypackage.MyClass```
in the classpath you just look on the classpath and return the first found class, if you use modules the process is slightly different.
1. You check, who/where the class is located that requested ```MyClass```
2. You check, which other modules the requesting class/module can actually read in the module graph
3. You look into the modules, that the requesting module can actually read and search there fore the class ```org.mypackage.MyClass```
Thus, in prinicipal a request to the class ```MyClass``` could return two different classes originating in two different modules, depending on the module graph.

Moreover, JARs on the module path have different visibility constraints (c.f. automatic or anonymous modules).

AnakinSklavenwalker:
> I am totally fine with only supporting Java 9+ classes when running on a Java 9+ VM.
If you mean, making this branch ready to merge first, i would be ok with that.
However totally discarding the possibility to perform corss-java version scenarios is something i think that has to be supported at some time. 

Still I also agree that the User Interface must be kept as simple as possible. Soot should do most of the stuff automatically if possible.

anddann:
thx a lot for your feedback Steven. In the next days, I will minimize the distance between this branch and the develop branch as much as possible and address your comments.

StevenArzt:
@anddann I think this is mostly a merge problem, rather than individual issues, with old code being dragged in for some reason.

anddann:
Thanks for the hint @StevenArzt. I could identify which merge went wrong and will update the branch tomorrow.

ericbodden:
@StevenArzt I spoke to @anddann about this merge request. As far as I understood, from Andreas' side, this request can be merged. Do you still see any open issues? If so, could you please let us know about specifically? Thanks!

StevenArzt:
@ericbodden @anddann The build still fails. On your Jenkins instance, you have a style check issue. On our instance, we have a failing test case (Java 11):

```
[ERROR] Failures: 
[ERROR]   InstanceOfCastsTest>AbstractASMBackendTest.runTestAndCompareOutput:133 Expected line CHECKCAST [Lsoot/asm/backend/targets/Measurable;, but got ARETURN in line 22 for class soot.asm.backend.targets.InstanceOfCasts
```

Can someone please have a look at that? Once we have a working build, I can throw some large Soot-based projects at it (with their own test cases) and see how that works.

ericbodden:
Thanks @StevenArzt. I am afraid our resources for this are a bit limited at the moment, so this may be quicker if you could take a look at those issues on your end. But if that's not possible then I will see what we can do. BTW is there something at odds with the StyleCheck integration? When I click the Details link above, I get "502 Bad Gateway".

mbenz89:
@ericbodden for some reason our Jenkins instance went down. Should be up and running again.

mbenz89:
Btw. everyone, merging this would still break our deployment functionality due to running maven inside of docker. Someone has to investigate a proper solution for this as well before merging...

ericbodden:
> The build still fails. On your Jenkins instance, you have a style check issue. 

@StevenArzt I fixed that one.

> On our instance, we have a failing test case (Java 11):
> 
> ```
> [ERROR] Failures: 
> [ERROR]   InstanceOfCastsTest>AbstractASMBackendTest.runTestAndCompareOutput:133 Expected line CHECKCAST [Lsoot/asm/backend/targets/Measurable;, but got ARETURN in line 22 for class soot.asm.backend.targets.InstanceOfCasts
> ```

@StevenArzt @mbenz89 
Why are we not running those test cases here? That would help with debugging.

> Btw. everyone, merging this would still break our deployment functionality due to running maven inside of docker. Someone has to investigate a proper solution for this as well before merging...

@mbenz89 Can you elaborate? How is this related to Java9? Why do the branches differ that way?

anddann:
I'll fix the failing test case at the weekend

anddann:
The test case ``InstanceOfCastsTest>AbstractASMBackendTest.runTestAndCompareOutput`` failes because the Javac 1.8 and Javac >=1.9 produce different byte code.

```
	public Measurable[] convertMeasurableArray(Object[] o) {
		if (o instanceof Measurable[]) {
			return (Measurable[]) o;
		}
		return null;
	}


```

Javac 1.8


```
  public soot.asm.backend.targets.Measurable[] convertMeasurableArray(java.lang.Object[]);
    descriptor: ([Ljava/lang/Object;)[Lsoot/asm/backend/targets/Measurable;
    Code:
       0: aload_1
       1: instanceof    #3                  // class "[Lsoot/asm/backend/targets/Measurable;"
       4: ifeq          15
       7: aload_1
       8: checkcast     #3                  // class "[Lsoot/asm/backend/targets/Measurable;"
      11: checkcast     #3                  // class "[Lsoot/asm/backend/targets/Measurable;"
      14: areturn
      15: aconst_null
      16: areturn
}

```

Java 1.9

```
  public soot.asm.backend.targets.Measurable[] convertMeasurableArray(java.lang.Object[]);
    descriptor: ([Ljava/lang/Object;)[Lsoot/asm/backend/targets/Measurable;
    Code:
       0: aload_1
       1: instanceof    #3                  // class "[Lsoot/asm/backend/targets/Measurable;"
       4: ifeq          12
       7: aload_1
       8: checkcast     #3                  // class "[Lsoot/asm/backend/targets/Measurable;"
      11: areturn
      12: aconst_null
      13: areturn
}
``` 
In fact, Java 1.8 generates a second ``checkCast`` instructionction for `` return (Measurable[]) o;`` whereas Java > 1.9 omits this second one.

anddann:
I have:
- addressed Steven's comment
- fix the test cases
- make all test cases pass on JDK8, JDK9, and JDK11
- adapted the Readme.md
- fixed the Jenkinspipline

mbenz89:
Thanks for the hard work @anddann ! 
Is this ready to merge from your side? I'll merge and prepare a release at the weekend then...

anddann:
Yes, from my side it is ready to  be merged. Nevertheless, there is still space for improvements see `SourceLocator.java`

anddann:
Are there still issues that prevent this PR from being merged?
I would be happy to fix them if you let me know.

mbenz89:
It's fine from my side.

What's left to do is to adapt the build jobs to propose the necessary environment to enable the deployment to maven central and check how the branch-plugin of Jenkins works together with the merge-request builder.
 
I won't be able to do this before about the 28. of this month sorry. You are free to have a look at this yourself if you like too, however.

Have a nice Christmas time everybody!

mbenz89:
I merged your changes and adapted the build pipeline accordingly. 
Soot's 4.0.0 release is about to come online an about an hour as a little Christmas present to the community :).

Thanks for your contribution!

PR #1156, opened 2019-05-21
 Created test case for verification of Soot.GetType() return type

rakshitkr:


anddann:
Is there a reason for reverting my commit and changing the source code back to the old asm bytecode frontend?

PR #1170, opened 2019-06-28
Fix ClassConstant value passed to Jasmin

### LINKED ISSUE #1167: Wrong ClassConstant value when loading classfile built with Jasmin
Tested with Soot 3.3.0 and Java 1.8
Inside a bytecode we may have some Class constants but for some reasons i can generate an invalid classfile from a valid classfile using the Jasmin backend with soot.
Let me explain in details : 
- Test.java uses some Class<?> and is compiled with javac to get a valid Test.class (1)
- This Test.class (1) is then used as an input with Soot to get a valid Test.jimple (1)
- Then Test.jimple (1) is transformed into a Test.class (2) using the Jasmin backend. The resulting Test.class (2) works with java but the bytecode instructions differs from the ASM backend.
- Finally we load Test.class (2) inside Soot to get a Test.jimple (2) that is invalid. In fact class constants becomes LLjava/lang/...;; which is invalid.

As the bytecode generated by the Jasmin back-end works with java, it is most likely a parser issue than a back-end issue.

The main differences between Jasmin and ASM backend regarding Class constants is the use of ldc_w instead of ldc and different corresponding Utf8 string (Ljava/lang/...; instead of simply java/lang/...). Everything can be analysed with `javap -verbose Test.class`.

All the files described are joined with this issue.
[bug_class_constant.tar.gz](https://github.com/Sable/soot/files/3324443/bug_class_constant.tar.gz)

Aztorius:
Fixes #1167 
This is needed for Java 9 conformance with Jasmin.
Please review the ClassConstant.toInternalString(). Although it works, it seems not perfect.
In details, classes are represented as 'java/lang/Object' instead of 'Ljava/lang/Object;'.
The patch is inspired from AsmUtil https://github.com/Sable/soot/blob/master/src/main/java/soot/asm/AsmUtil.java#L72 .
Bytecode is now corresponding between ASM, javac and Jasmin (some corner cases may remains different).

### DIFF
--- a/src/main/java/soot/jimple/ClassConstant.java
+++ b/src/main/java/soot/jimple/ClassConstant.java
@@ -134,6 +134,27 @@ public class ClassConstant extends Constant {
     return numDimensions > 0 ? ArrayType.v(baseType, numDimensions) : baseType;
   }
 
+  /**
+   * Gets an internal representation of the class used in Java bytecode.
+   * The returned string is similar to the fully qualified name but with '/' instead of '.'.
+   * Example: "java/lang/Object".
+   * See https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.2.1
+   */
+  public String toInternalString() {
+    String internal = value;
+    while (internal.startsWith("[")) {
+      internal = internal.substring(1);
+    }
+    if (internal.endsWith(";")) {
+      internal = internal.substring(0, internal.length() - 1);
+      if (internal.startsWith("L")) {
+        internal = internal.substring(1);
+      }
+    }
+
+    return internal;
+  }
+
   // In this case, equals should be structural equality.
   @Override
   public boolean equals(Object c) {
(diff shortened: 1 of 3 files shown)


mbenz89:
ok to test

Aztorius:
@mbenz89 tests are passing.
Good to merge ?

mbenz89:
Sure, it just went out of my head.
Thanks for your contribution!

src/main/java/soot/jimple/ClassConstant.java
```diff
@@ -134,6 +134,24 @@ public Type toSootType() {
     return numDimensions > 0 ? ArrayType.v(baseType, numDimensions) : baseType;
   }
 
+  /**
+   * Gets an internal representation used in Java bytecode
+   */
+  public String toInternalString() {
```
mbenz89: Could you please add a bit more documentation on how the internal representation actually looks like? It gets a bit confusing on what to expect from all the "toString" methods here...

Aztorius: Fixed

PR #1175, opened 2019-07-03
Add class soot.util.WeakMapNumberer

### LINKED ISSUE #1174: High Memory Usage with Numberer in soot.Scene
While creating and deleting a huge amount of SootClass, SootMethod, SootField and SootLocal there is an increasing memory consumption. Using the VisualVM profiler it appears that soot.Scene contains some ArrayNumberer and Numberer that contains references to normaly deleted objects. Thus those objects are not deleted by the garbage collector.

I am thinking of a couple of solutions : 
- Add remove methods to soot.Scene for methodNumberer and others fields
- Change to public those fields to be able to remove objects inside it
- Set all of those fields to null with a method (name ex: disableNumbererSystem) and change all code that use them

I don't know which solution should be preferred. Let me know what you think and I may add a pull request.

Aztorius:
Work in progress
Useful for #1174 .

Inspired by ArrayNumberer and MapNumberer with weak references to reduce memory consumption.
For performance reasons we create 2 WeakHashMap each one is the reverse of the other.
Even if it is 2x more consuming memory it is definitively a win when removing a lot of objects during execution.

### DIFF
--- a/src/main/xml/options/soot_options.xml
+++ b/src/main/xml/options/soot_options.xml
@@ -405,6 +405,16 @@
                 and the original classes you loaded. Use this option at your own risk.
             </long_desc>
         </boolopt>
+        <boolopt>
+            <name>Weak Map Structures</name>
+            <alias>weak-map-structures</alias>
+            <short_desc>Use weak references in Scene to prevent memory leakage when removing many classes/methods/locals</short_desc>
+            <long_desc>
+                If this option is set, Soot will use Maps with weak references to lower memory usage when performing
+                many additions and deletions of classes/methods/locals. Without this option, all created objects
+                are kept in memory. This has a bit larger memory footprint if only a small amount of deletions are conducted.
+            </long_desc>
+        </boolopt>
     </section>
     <section>
         <name>Input Options</name>
--- a/doc/soot_options.htm
+++ b/doc/soot_options.htm
@@ -508,6 +508,10 @@
             <td><tt>-ignore-resolving-levels </tt><br></td>
             <td colspan="2">Ignore mismatching resolving levels</td>
          </tr>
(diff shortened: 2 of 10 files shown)


src/main/xml/options/soot_options.xml
```diff
@@ -405,6 +405,16 @@
                 and the original classes you loaded. Use this option at your own risk.
             </long_desc>
         </boolopt>
+        <boolopt>
+            <name>Weak Map Structures</name>
+            <alias>weak-map-structures</alias>
+            <short_desc>Replace some Arrays with Weak Map structures</short_desc>
```
mbenz89: Could we make the short_desc a bit more expressive? In the options html, only the short_desc is shown for general options, so this is the most important part.

Maybe something like: "Use weak references in Scene to prevent memory leakage when removing many classes/methods/locals".

src/main/xml/options/soot_options.xml
```diff
@@ -405,6 +405,16 @@
...
+            <long_desc>
+                If this option is set, Soot will use Weak Maps to lower memory usage. This is useful when performing
+                many additions and deletions of Locals or Methods. Without this option, all created Locals or Methods
+                are kept in memory.
```
mbenz89: And here a sentence regarding the drawback, e.g.: "This has a bit larger memory footprint if only a small amount of deletions are conducted."

src/main/xml/options/soot_options.xml
Aztorius: Fixed

src/main/xml/options/soot_options.xml
Aztorius: Fixed

PR #1191, opened 2019-08-09
Fix/name to class map

anddann:
Fix concurrent access issues for `Scene.nameToClass` map


mbenz89:
should fix https://github.com/Sable/soot/issues/1122 and the issues linked there. I think there might be some more issues caused by this in the issue tracker. Someone should have a look :)

PR #1204, opened 2019-09-19
Default Interface Test Case

pavanupb:


mbenz89:
@pavanupb Please fix and finish up the test suite asap. I like to test my fix for the default interfaces.

pavanupb:
> @pavanupb Please fix and finish up the test suite asap. I like to test my fix for the default interfaces.

@mbenz89 I am working today will probably complete test suite soon. I already have pushed the target code for all the corner cases. I went through the blog you provided and quite other articles on default interfaces. I could get three different cases which I have pushed yesterday.

src/systemTest/targets/soot/interfaceTesting/TestDerivedInterfaces.java
```diff
@@ -13,7 +13,7 @@ default void print() {
 	}
 }
 
-interface interfaceTestTwo extends interfaceOne{
+interface interfaceTestTwo extends interfaceTestOne{
```
mbenz89: In Java classes and Interfaces start with a capital letter.

pavanupb: Fixed it

mbenz89:
It seems that there are still checks missing for the abstract dispatch. When will you add those?

mbenz89:
Could you please place your classes into package "soot.defaultInterfaceMethods" in both java and targets dir and fix the signatures accordingly?

pavanupb:
> It seems that there are still checks missing for the abstract dispatch. When will you add those?

I have a query which I posted over the mail. Can you please confirm once. I have posted an example. After that I will be adding the assert statements.

pavanupb:
> Could you please place your classes into package "soot.defaultInterfaceMethods" in both java and targets dir and fix the signatures accordingly?
> 
> Also please do not let the targets start with "Test" and name them according to your testing methods, e.g. `"soot.testing.framework.DefaultInterfaceTest#interfaceWithSameSignatureTest`and `soot.interfaceTesting.TestInterfaceSameSignature` do not match.

Sure, I will be changing it.

mbenz89:
I just found a few issues in one of your test case regarding "resolveNonSpecial" and "ConcreteDispatch". Can you double check and fix the other test cases accordingly: https://github.com/Sable/soot/commit/64b93aa05a2d5b26a156b04f22452b4364c563d4

pavanupb:
> I just found a few issues in one of your test case regarding "resolveNonSpecial" and "ConcreteDispatch". Can you double check and fix the other test cases accordingly: [64b93aa](https://github.com/Sable/soot/commit/64b93aa05a2d5b26a156b04f22452b4364c563d4)

Could you let me know what is the issue?. I couldn't get from [64b93aa](https://github.com/Sable/soot/commit/64b93aa05a2d5b26a156b04f22452b4364c563d4)

mbenz89:
In line 157, 158, 169, 170 you specified that the dispatch should resolve to the interfaces instead to the class that overwrites the print methods.

pavanupb:
> In line 157, 158, 169, 170 you specified that the dispatch should resolve to the interfaces instead to the class that overwrites the print methods.

Oops my bad. Will check others and correct them.

mbenz89:
Btw. I just found that `SootMethodRefImpl#tryresolve` does not attempt to do a virtual dispatch, it just tries to find the first matching method and return it.

So  for `soot.testing.framework.DefaultInterfaceTest#derivedInterfacesTest`, it will resolve to `InterfaceTestOne#print` instead of `InterfaceTestTwo#print` because it goes trough the interfaces sequentially and takes the first matching one. 

Please double-check this in your cases and fix as well?

pavanupb:
> Btw. I just found that `SootMethodRefImpl#tryresolve` does not attempt to do a virtual dispatch, it just tries to find the first matching method and return it.
> 
> So for `soot.testing.framework.DefaultInterfaceTest#derivedInterfacesTest`, it will resolve to `InterfaceTestOne#print` instead of `InterfaceTestTwo#print` because it goes trough the interfaces sequentially and takes the first matching one.
> 
> Please double-check this in your cases and fix as well?

Okay

mbenz89:
Forget that. The implementation is incorrect. We have to consider order since java 8 so I have to fix this. 

The logic is basically take the first matching method, first searching the super class heirarchy and then the interface hierarchy. Take the most specific interface method that fits. The method can be abstract! (That is the main difference to virtual dispatch).

pavanupb:
In soot.testing.framework.DefaultInterfaceTest#derivedInterfacesTest, the print() of InterfaceTestTwo should win over InterfaceTestOne right?. According to Rule#2 of https://blog.codefx.org/java/everything-about-default-methods/#Conflict-Resolution

mbenz89:
> In soot.testing.framework.DefaultInterfaceTest#derivedInterfacesTest, the print() of InterfaceTestTwo should win over InterfaceTestOne right?. According to Rule#2 of https://blog.codefx.org/java/everything-about-default-methods/#Conflict-Resolution

Exactly.

PR #1220, opened 2019-10-22
Clarify semantics of Chain.follows

MarcMil:


mbenz89:
Wouldn't it make sense to fix the implementation? It sounds very counterintuitive as it currently is which is also represented by the required documentation.

MarcMil:
I'd like to do that as well, but what about existing implementations which may depend on the current implementation's behavior?

mbenz89:
Are you aware of any such implementations? I would rather guess that people using the API expect it to behave as the previous documentation stated.
There seem to be only two uses in the Soot codebase. Maybe you have a look at those and potential usages in FlowDroid.

Anyway, I don't think we should keep strange things like this for "backward compatibility". We will issue a new release soon and if people do not want to adapt to the changes, they are free to stick to an older version.

MarcMil:
I've changed the implementation. Should be consistent now.

mbenz89:
Thanks for that! Have you checked the usages in Soot for correctness?

MarcMil:
It should be fine, as no different units from different methods should be part of the follows call at these two call sites.

mbenz89:
Perfect, thanks!

Issue #1062, opened 2018-11-13
Implement "default" interface methods introduced in Java 8


mbenz89:
Java 8 introduces _default_ methods for interfaces that contain a default implementation of an interface method. (https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html)

Soot does currently not handle this construct and has to be extended for it.

Related to #1054
Maybe related to #956

mbenz89:
Solution: soot.SootMethodRefImpl#tryResolve(java.lang.StringBuffer) has to deal with interfaces as well.

amishra123:
I would be interested to work on this.

anddann:
@amishra123 Thanks!
We welcome your participation?
Do you want to do it alone or should we implement this feature together?

mbenz89:
@amishra123 Great to hear! Let us know if you have questions.

amishra123:
> @amishra123 Thanks!
> We welcome your participation?
> Do you want to do it alone or should we implement this feature together?

Either ways is fine with me. Let me come up with an implementation and I'll let you know if I need help. Thanks.

anddann:
This is a duplicate of  issue #1054.
@amishra123 Did you make any progress or need support?

mbenz89:
done with https://github.com/soot-oss/soot/pull/1221

PR #1221, opened 2019-10-22
Feature/default methods

### LINKED ISSUE #1054: Soot does not find default methods reachable
Consider the example:

```java
public class Main implements Default {
  public static void main(String[] args) {
    new Main().target();
  }
}
interface Default {
  default void target() {
    System.out.println("Hello, World!");
  }
}
```

When trying to find reachable methods from main, soot does not find Default.target. Even though
the program print "Hello, World!", when run.

I followed the same setup as in #1045.

## Versions:
Soot version 3.1.0

```
$ java -version
openjdk version "1.8.0_181"
OpenJDK Runtime Environment (build 1.8.0_181-13)
OpenJDK 64-Bit Server VM (build 25.181-b13, mixed mode)
```

mbenz89:
Implements virtual dispatch of default methods for Java > 7. (fixes #1054, #1062)

Tries to unify dispatch behavior that was duplicated among the classes `VirtualCalls`, `FastHierarchy` and `SootMethodRefImpl#tryResolve`, to be able to maintain one feature-complete implementation at all locations.


mbenz89:
@anddann Could you have a look at especially `FastHierarchy#concreteDispatch` and `FastHierarchy#abstractDispatch` if that makes sense to you?

src/main/generated/options/soot/AntTask.java
```diff
@@ -366,6 +370,12 @@ public void setjava_version(String arg) {
                 || arg.equals( "8" )
                 || arg.equals( "1.9" )
                 || arg.equals( "9" )
+                || arg.equals( "1.10" )
+                || arg.equals( "10" )
+                || arg.equals( "1.11" )
+                || arg.equals( "11" )
+                || arg.equals( "1.12" )
+                || arg.equals( "12" )
```
anddann: if you want to support java 12, we should also update the ASM dependency

src/main/java/soot/SootMethod.java
```diff
@@ -586,7 +586,7 @@ public boolean isStaticInitializer() {
    * @return yes, if this is a class initializer or main function.
    */
   public boolean isEntryMethod() {
-    if (isStatic() && subsignature.equals(VirtualCalls.v().sigClinit)) {
```
anddann: we should delete `sigClinit` field completely

src/main/java/soot/asm/AsmUtil.java
```diff
@@ -297,7 +286,88 @@ public static String baseTypeName(String s) {
...
   }
 
-  private AsmUtil() {
+  private AsmUtil() {}
+
+  public static int byteCodeToJavaVersion(int bytecodeVersion) {
+    int javaVersion;
+
+    switch (bytecodeVersion) {
+      case (Opcodes.V1_5):
```
anddann: wouldn't it be easier to store the java_version directly in the Opcodes and then calling

 `OpCode opcode = OpCodes.valueOf(byteCodeVersion); 
return opcode.get_java_version();`

src/main/generated/options/soot/AntTask.java
mbenz89: Haven't you already done this with the Java9 branch? Otherwise, I'm happy to do that.

src/main/java/soot/asm/AsmUtil.java
mbenz89: I have to have a look at this. It will only work if the returned java version has the same format as the one defined in the options.  I wouldN#t be suprised if somewhere in soot the version is compared with some constant string which would not change with a format change of the constants in `Options`.

mbenz89:
Thanks for your review @anddann . There are still some issues with this feature that Pavan is investigating. We have to wait for his ok to merge this.

mbenz89:
TODO: We generate phantom methods in SootMethodRef implementation if we cannot dispatch a method. These phantom methods are then added to the call-graph as targets. In the old implementation this did not happen since we did not tried to dispatch a `SootMethod` (which has to first be generated by `SootMethodRefImpl`) but only worked with the signature. 

I think both is fine. Since the old implementation did not have those we should still make sure we also do not add such guessed dispatch methods to stay consistent to what clients might expect from the call-graph.

mbenz89:
Fixed issues from the last comment.

Note: We just noticed that the old implementation did not contain some edges to overwritten methods in abstract classes since the `typeToSubType` map in `VirtualCalls`  only considers `concrete` subtypes, while in this implementation we use `getSubtypesFor` in `FastHierarchy`, which catches also abstract sub-classes.

src/main/java/soot/SootMethod.java
mbenz89: @pavanupb could you please do this?

src/main/java/soot/asm/AsmUtil.java
mbenz89: @pavanupb Could you please check if the approach proposed by andreas can be used to replace my switch case?

src/main/generated/options/soot/AntTask.java
```diff
@@ -223,6 +223,10 @@ public Path createprocess_dir() {
             return process_dir.createPath();
         }
   
+        public void setderive_java_version(boolean arg) {
```
mbenz89, line 226: @pavanupb Here is the stuff? What are you missing?

Please just make sure that you do not remove anything I added in this mr with your merge with develop. All files in the `src/main/generated` folder will be generated when calling `mvn verify`

pavanupb: @mbenz89 I executed the maven command mvn clean verify but still the files are not getting generated. Am I missing something?

src/main/java/soot/SootMethod.java
pavanupb: @mbenz89 The sigClinit field in SootMethod class has been removed completely in my feature/default-methods branch. While merging I took the changes from develop, so after merging my PR this should be reflected.

src/main/java/soot/asm/AsmUtil.java
pavanupb: > @pavanupb Could you please check if the approach proposed by andreas can be used to replace my switch case?

@mbenz89 Again I took the changes from the develop here. So should be reflecting once my PR is merged.

src/main/generated/options/soot/AntTask.java
mbenz89: Are you sure you did not overwrite my changes with the ones from develop?

src/main/java/soot/asm/AsmUtil.java
mbenz89: Andreas proposed to change the functionality by using OpCodes class. Could you check if that makes sense, i.e. produces the same String values for the versions.

src/main/generated/options/soot/AntTask.java
pavanupb: Yes I took your changes as well as the new changes from develop

src/main/java/soot/asm/AsmUtil.java
pavanupb: Should I use the statement proposed by andreas in the constructor of AsmUtil class?.

mbenz89: No. I think you should replace the switch case here and in `javaToBytecodeVersion` with the code he proposed. Make sure that the result in the same strings first! Maybe write a small test case that  uses the switch cases and the Opcodes approach and compare the results for all the java versions

src/main/xml/options/soot_options.xml
```diff
@@ -542,6 +545,23 @@
                 </p>
             </long_desc>
         </listopt>
+        <boolopt>
+            <name>Derive Java version from input</name>
+            <alias>derive-java-version</alias>
```
mbenz89, line 550: @pavanupb This is the code I actually added manually. From this xml document the options.html and some Java classes relating to the option will be automatically created by Maven. if you merged my changes from options.xml into your branch, you should thus get all the Java code related to `derive-java-version`

src/main/java/soot/asm/AsmUtil.java
pavanupb: @mbenz89 Which package should I be using for Opcodes. I checked with org.jf.dexlib2.Opcode.valueof() function, but I am getting an exception thrown when I send the bytecodeVersion value as 49.

mbenz89: you need to use asm. dexlib is for android.

pavanupb: So it should be objectweb.asm correct?.

pavanupb: The Opcodes interface does not have a definition for valueOf. Am I using the right package?

mbenz89: Yes it shout be object web.sm I think. If there is no value of Andreas maybe mebt a different library. If you cannot get this done under 1h it's not worth the effort.

PR #1224, opened 2019-10-30
Fix the Lambda newinvokespecial exception

raintung:
Fix the Lambda newinvokespecial exception, Issue class:

```
import java.util.List;

public class testLambdaNew {
	class test {
		public test(String s) {
			System.out.println("Hello " + s);
		}
	}
	public void go(List<String> list){
		list.stream().forEach(test::new);
	}
}
```

```
Caused by: java.lang.RuntimeException: Box InvokeExprBox(null) cannot contain value: $r4(class soot.jimple.internal.JimpleLocal)
	at soot.AbstractValueBox.setValue(AbstractValueBox.java:41)
	at soot.jimple.internal.InvokeExprBox.<init>(InvokeExprBox.java:31)
	at soot.jimple.Jimple.newInvokeExprBox(Jimple.java:728)
	at soot.jimple.internal.JInvokeStmt.<init>(JInvokeStmt.java:46)
	at soot.jimple.Jimple.newInvokeStmt(Jimple.java:608)
	at soot.LambdaMetaFactory$ThunkMethodSource.invokeImplMethod(LambdaMetaFactory.java:672)
	at soot.LambdaMetaFactory$ThunkMethodSource.getInvokeBody(LambdaMetaFactory.java:486)
	at soot.LambdaMetaFactory$ThunkMethodSource.getBody(LambdaMetaFactory.java:337)
	at soot.SootMethod.retrieveActiveBody(SootMethod.java:402)
	at soot.LambdaMetaFactory.makeLambdaHelper(LambdaMetaFactory.java:239)
	at soot.asm.AsmMethodSource.convertInvokeDynamicInsn(AsmMethodSource.java:1420)
	at soot.asm.AsmMethodSource.convert(AsmMethodSource.java:1791)
	at soot.asm.AsmMethodSource.getBody(AsmMethodSource.java:2038)
```

### DIFF
--- a/src/main/java/soot/LambdaMetaFactory.java
+++ b/src/main/java/soot/LambdaMetaFactory.java
@@ -645,14 +645,15 @@ public final class LambdaMetaFactory {
     private void invokeImplMethod(JimpleBody jb, PatchingChain<Unit> us, LocalGenerator lc, List<Local> args) {
       Value value = _invokeImplMethod(jb, us, lc, args);
 
-      if (value instanceof InvokeExpr && soot.VoidType.v().equals(implMethod.getMethodRef().getReturnType())) {
-        // implementation method is void
-        us.add(Jimple.v().newInvokeStmt(value));
-        us.add(Jimple.v().newReturnVoidStmt());
-      } else if (soot.VoidType.v().equals(implMethodType.getReturnType())) {
-        // dispatch method is void
-        us.add(Jimple.v().newInvokeStmt(value));
-        us.add(Jimple.v().newReturnVoidStmt());
+      if ( soot.VoidType.v().equals(implMethodType.getReturnType()) || soot.VoidType.v().equals(implMethod.getMethodRef().getReturnType())) {
+          // implementation method is void
+      	if(value instanceof Local) {
+      		Local ret = lc.generateLocal(value.getType());
+              us.add(Jimple.v().newAssignStmt(ret, value));
+      	} else {
+      		us.add(Jimple.v().newInvokeStmt(value));
+      	}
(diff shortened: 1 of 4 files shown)


mbenz89:
Hi. Thanks for your contribution. The PR looks quite fine. Could you please also provide a test case that tests for the fixed issue, i.e., fails without your fix?

You can use https://github.com/Sable/soot/blob/master/src/systemTest/java/soot/lambdaMetaFactory/Issue1146Test.java and https://github.com/Sable/soot/blob/master/src/systemTest/targets/soot/lambdaMetaFactory/Issue1146.java as baseline for your test case.

mbenz89:
Any updates?

raintung:
It very different with your test case. The major problem is that Issue1146.java source is compiled by mvn jvm version then Issue1146Test.java analyze the Issue1146.class byte code after compiled.
For this bug, only some jdk 1.8 low version compile will meet this case. 
If provide the java source, it can't duplicate it in high jdk version that depend on running unit test jvm version.
At best provide the java byte code, unit test only analyze this byte code. I amn't sure the testing framework can implement it.

mbenz89:
Hi @raintung . 
In this case using the testing framework does not make much sense. It mainly exists to allow for analyzing java source code that is compiled to bytecode by maven during the build process.

If you require a specific .class file, you could add it to the `test/resources` directory and write a test case that runs soot only on that file. I would create a specific folder for the test-case in the resources directory and then point Soot towards this folder.

Let me know if you need help :).

raintung:
create the new pull request:
https://github.com/Sable/soot/pull/1295

PR #1250, opened 2019-12-10
Fix issue 618: NullPointerException when creating soot.toolkits.graph.pdg.EnhancedUnitGraph

ftc:
This pull request fixes the issue: https://github.com/Sable/soot/issues/618 . I added a unit test to load the example code from the linked issue and test that the exception traps are reachable.


mbenz89:
Great PR. Thanks for your contribution!

PR #1263, opened 2019-12-31
Improve error message in FieldRefValidator

knewbury01:
Previously in exception handling the error message
printed was incorrect (looks like copy paste code mistake to me). 
Previously, in the jimple field validation step, 
when a field reference in jimple was failed to be resolved, it
would print some error message that was unrelated to the cause of the problem.

Now FieldRefValidator reports the actual message
in the cases that an exception occurred.

This is an enhancement,

This change will improve the ability to debug programatically constructed jimple.


mbenz89:
Thanks for your contribution!

PR #1265, opened 2020-01-06
Update FastHierarchy.java

pavanupb:


src/main/java/soot/FastHierarchy.java
```diff
@@ -53,6 +55,8 @@
  */
 public class FastHierarchy {
   private static final Logger LOGGER = LoggerFactory.getLogger(FastHierarchy.class);
+  private final LargeNumberedMap<Type, SmallNumberedMap<SootMethod>> typeToVtbl =
```
mbenz89: Please change to Map<SootClass, Map<String,SootMethod>> and use HashMaps as implementation

src/main/java/soot/FastHierarchy.java
```diff
@@ -763,10 +770,21 @@ private SootMethod resolveMethod(
       Set<SootClass> ignoreList) {
     SootClass concreteType = baseType;
 
+    
+    SootMethod candidate = null;
```
mbenz89: Generate a signature with soot.SootMethod#getSubSignature(java.lang.String, java.util.List<soot.Type>, soot.Type) as key for the inner map of vtable

mbenz89: Use the declaring class as key for the outer map

src/main/java/soot/FastHierarchy.java
```diff
@@ -763,10 +770,21 @@ private SootMethod resolveMethod(
...
+    		return candidate;
+    	}
+    }
+    
     // When there is no proper dispatch found, we simply return null to let
     // the caller decide what to do
-    SootMethod candidate = null;
-
     while (concreteType != null && !ignoreList.contains(concreteType)) {
       ignoreList.add(concreteType);
```
mbenz89: Dont forgett to add the resolved method to the vtable if one was found (at the end of the method)

pavanupb:
@mbenz89 I have changed the vtbl. Please have a look and let me know for any changes

PR #1267, opened 2020-01-13
Fix the bug for "requires transitive"

raintung:
recycle check if exported packages is "requires transitive" case.
"requires transitive" module will transmit, need chain check until transitive finished.

Pull request in the same branch.


anddann:
Thanks for the PR @raintung ; I will take a look at the PR and @mbenz89 will merge it.

anddann:
Looks super from my side. It would be awesome if you could provide a test case because we want to make the Java Module features more stable.

anddann:
Thanks for your bug fix @raintung 
@mbenz89 with the test cases I've added; I'm fine with this PR. Do you see any issue with merging this PR?

PR #1268, opened 2020-01-15
Virtual call resolution takes default methods into account.

Lex999:
I've modified the virtual call resolution for call graph construction to take default methods into account. I've tested it with SPARK and some artificial examples. This should solve issue #1054 . Maybe, someone knows of a more elegant way to find the "maximally-specific superinterface methods" as defined in the JVM specification.

### DIFF
--- a/src/main/java/soot/jimple/toolkits/callgraph/CallGraph.java
+++ b/src/main/java/soot/jimple/toolkits/callgraph/CallGraph.java
@@ -122,9 +122,11 @@ public class CallGraph implements Iterable<Edge> {
     boolean hasSwapped = false;
     for (Iterator<Edge> edgeRdr = edgesOutOf(out); edgeRdr.hasNext();) {
       Edge e = edgeRdr.next();
+      MethodOrMethodContext src = e.getSrc();
+      MethodOrMethodContext tgt = e.getTgt();
       removeEdge(e);
       e.remove();
-      addEdge(new Edge(e.getSrc(), in, e.getTgt()));
+      addEdge(new Edge(src, in, tgt));
       hasSwapped = true;
     }
     return hasSwapped;
--- a/src/main/java/soot/jimple/toolkits/callgraph/ClinitElimTransformer.java
+++ b/src/main/java/soot/jimple/toolkits/callgraph/ClinitElimTransformer.java
@@ -1,7 +1,5 @@
 package soot.jimple.toolkits.callgraph;
 
-import java.util.HashSet;
-
 /*-
  * #%L
  * Soot - a J*va Optimization Framework
@@ -26,7 +24,6 @@ import java.util.HashSet;
 
 import java.util.Iterator;
 import java.util.Map;
-import java.util.Set;
 
 import soot.Body;
 import soot.BodyTransformer;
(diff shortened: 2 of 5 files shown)


mbenz89:
Hi. I've already implemented that feature in https://github.com/Sable/soot/pull/1221 . We are still testing the branch due to some more intrusive changes. 

I did not had the time to compare with your implementation. Maybe you want to have a look yourself as well. The main implementation is now located in `FastHierarchy' and `VirtualCalls`class does only delegate to the hierarchy. There was a lot of duplicated code which is why I merged those. 
It also adds subsignature polymorphism. I have to check but your approach for finding the "maximally-specific superinterface methods"  might be a bit more performant than mine.

Lex999:
Even better. Your implementation looks more generic. Performance is difficult to guess from just looking at the code. As your pull request is not referenced in issue #1054, I didn't notice that someone is working on this.

mbenz89:
I'm sorry I did not mention the fix so you worked on it as well. I'll compare our implementations of finding the maximally specific interface and take the more performant one.
Thanks for doing this work, anyways!

src/main/java/soot/jimple/toolkits/callgraph/VirtualCalls.java
```diff
@@ -106,19 +108,67 @@ public SootMethod resolveNonSpecial(RefType t, NumberedString subSig, boolean ap
...
+   * Collects all interfaces which are directly or indirectly implemented by the given class.
+   * Additionally, stores the max inheritance distance between the given class and each interface.
+   *
+   * @param cls                  The class to collect the interfaces for.
+   * @param inheritanceDistance  The map which receives the interfaces together with the max inheritance distance.
+   * @param distance             The current distance from the class of interest.
+   */
+  private void collectInterfaces(SootClass cls, HashMap<SootClass,Integer> inheritanceDistance, int distance) {
+    for (SootClass interf : cls.getInterfaces()) {
+      inheritanceDistance.merge(interf, distance, Integer::max);
```
mbenz89: Why do you store the `max` distance? Wouldn' be the `min` distance the interesting part if you want to find the most specific interface? 

Also, you are invoking this with 0 all the time. How can the distance ever be different to 0?

mbenz89: You always store the maximum distance. Shouldn't you store the minimum distance to find the most specific interface?
If a class inherits from the same interface over two different inheritance chains, you would take the farthest away implementation as most specific.

Lex999: The list of all interfaces is sorted in ascending order (line 164). So the "nearest" interface will be at the beginning of the list and will be searched first for an implementation. For one interface, always the maximum distance is stored to make sure it comes in the list after all implemented child interfaces.

Lex999: The correct line number is 146.

mbenz89: I see. Thanks for the explanation.

PR #1269, opened 2020-01-15
Primitive casts are not considered essential.

Lex999:
During execution of dead assignment elimination, primitive casts are not essential because they cannot throw any exceptions.


mbenz89:
Could you elaborate this fix a bit more, please? An example where the old code breaks would be very helpful.

Lex999:
Consider the following Jimple code:
```
public static int myMethod(int, int)
{
   int a, b, c;
   double d;
   a = @parameter0: int;
   b = @parameter1: int;
   c = a + b;
   d = (double) a;
   return c;
}
```
The old code would not eliminate the assignment to d because "double" is no ReferenceType. As primitive type casts cannot throw any exceptions, they can be removed without modifying behaviour when the result is not used. The new code will remove the assignment to d.

After thinking about it, "RefType" in the modified code line should probably replaced by "RefLikeType" to catch casts to an ArrayType.

mbenz89:
I see. Thanks for your explanation!
Could you add a test case for the fix? This would greatly help to understand the feature and keep it robust against future changes which.

> After thinking about it, "RefType" in the modified code line should probably replaced by "RefLikeType" to catch casts to an ArrayType.

Good point. Could you change that as well?

ericbodden:
@Lex999 are you planning any updates on this? We'd be happy to merge it once completed.

Lex999:
I'm going to write a test and make the type check correct. It will take a week or two.

mbenz89:
Great! Thank you for your efforts!

Lex999 <notifications@github.com> schrieb am Sa., 29. Feb. 2020, 16:26:

Lex999:
OK. Now, it should be ready for merge.

PR #1275, opened 2020-01-21
Fix Jenkinsfile due to jenkins bug

anddann:
docker agent should reuse the node otherwise a deadlock may occur
https://issues.jenkins-ci.org/browse/JENKINS-60701?attachmentViewMode=gallery


mbenz89:
Already fixed in develop. Thanks anyways.

PR #1276, opened 2020-01-22
Removes uses of unordered data structures.

michael-emmi:


mbenz89:
Great improvement, thanks for your contribution!

PR #1281, opened 2020-01-23
Deterministic generation of local variable names.

michael-emmi:


mbenz89:
Some time ago we introduced the `LocalGenerator` which is encapsulating this naming behavior. Could you please refactor your fix to use the generator instead?

michael-emmi:
Ahh thanks for the tip! I'll update this PR shortly.

anddann:
do you have any test code available on which you experienced the issue?
We would be happy to include it as a test case

michael-emmi:
Hmm well I’m not sure how to write a test that will reliably fail due to nondeterminism.

anddann:
Okay, I see. A test that specifies just the desiered outcome would be sufficient

michael-emmi:
@mbenz89 using the `LocalGenerator` now.

@anddann I’m at a loss at how to write a test for this one. What I want to ensure is that a given Jimple assignment is given the same variable name in every run, *for a given Soot version, from the same Java byte code*. At the same time, fixing any particular name would make the test fragile — any future naming-related update could break it. That seems to leave us with running Soot twice and accessing the name of one particular unit’s assignment in each run. But I don’t actually know how to write a program for which such a name can vary across runs for two reasons. First, I have only observed this in very large programs whose sources I cannot reveal. Second, the variance seems to rely on JVM implementation details, e.g., how default hash-codes are generated.

As I do not know how to write a test for this that could ever fail, and you probably don’t want vacuous or fragile tests, could this PR could be admitted without an additional test?

Anyhow a much better way to prevent regressions to this sort of thing is probably to ensure that (a) default hash-code values are never used, and (b) iteration does not occur over unordered data structures, i.e., by banning certain classes and method calls in the source code. This seems clearly beyond the scope of this PR.

anddann:
Thanks @michael-emmi , sry my mistake. I totally see your point. From my side this PR can be merged.
Thanks a lot for your contribution

michael-emmi:
@mbenz89 are you happy with this PR now?

src/main/java/soot/jimple/toolkits/typing/fast/TypeResolver.java
```diff
@@ -585,9 +584,8 @@ private void split_new() {
                   deflist = defs.getDefsOfAt((Local) assign.getRightOp(), assign);
                   continue;
                 } else if (assign.getRightOp() instanceof NewExpr) {
-                  Local newlocal = jimple.newLocal("tmp", null);
```
mbenz89: This local was initialized without a type. You changed that. Any idea why that was the case?

mbenz89:
Beyond my question, I'm happy with this pr. You are not only making Soot more deterministic but also cleaning up some old mess here. Thanks for contributing!

src/main/java/soot/jimple/toolkits/typing/fast/TypeResolver.java
michael-emmi: Yes, it’s because the local generator does not accept null values for that parameter — it would throw an exception.

mbenz89: Which is a good thing imho. It might not be needed in this case but users of the api should be able to expect and get a type for a local variable.

PR #1287, opened 2020-02-03
Improve readability of local variable names when reading dex 

zxy-sourcebrella:
Hello everyone,

This PR uses the variable names recorded in the debug information in `.dex` files for registers when reading `.dex` files into jimple, making the generated code a little bit more readable.

Along with the original names for registers, the debug information also carries the original Java type and signature for each register, so it might be useful for improving the type inference when jimplification. 

Best,
Zhixuan

### DIFF
--- a/src/main/java/soot/dexpler/DexBody.java
+++ b/src/main/java/soot/dexpler/DexBody.java
@@ -29,10 +29,10 @@ package soot.dexpler;
 
 import static soot.dexpler.instructions.InstructionFactory.fromInstruction;
 
+import com.google.common.collect.ArrayListMultimap;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -181,7 +181,7 @@ public class DexBody {
     }
   }
 
-  private final Map<Integer, LinkedList<RegDbgEntry>> localDebugs;
+  private final ArrayListMultimap<Integer, RegDbgEntry> localDebugs;
 
   // detect array/instructions overlapping obfuscation
   protected List<PseudoInstruction> pseudoInstructionData = new ArrayList<PseudoInstruction>();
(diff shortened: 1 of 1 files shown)


src/main/java/soot/dexpler/DexBody.java
```diff
@@ -420,6 +488,10 @@ public Body jimplify(Body b, SootMethod m) {
       int thisRegister = numRegisters - numParameterRegisters - 1;
 
       Local thisLocal = jimple.newLocal("$u" + thisRegister, unknownType); // generateLocal(UnknownType.v());
+      //thisLocal.setType(jBody.getMethod().getDeclaringClass().getType());
+      if (localDebugs.containsKey(thisRegister)) {
+        thisLocal.setName("this");
```
StevenArzt: We should be careful to avoid duplicate local names in case of obfuscated code with deliberately broken debug info

src/main/java/soot/dexpler/DexBody.java
```diff
@@ -204,6 +233,8 @@ protected DexBody(DexEntry<? extends DexFile> dexFile, Method method, RefType de
 
     instructions = new ArrayList<DexlibAbstractInstruction>();
     instructionAtAddress = new HashMap<Integer, DexlibAbstractInstruction>();
+    localDebugs = new HashMap<Integer, LinkedList<RegDbgEntry>>();
```
StevenArzt: That should be some sort of LinkedMultiMap (maybe create such a class if we don't have it already) for simplifying the code here and not having to deal with putting lists into maps

src/main/java/soot/dexpler/DexBody.java
```diff
@@ -458,6 +530,9 @@ public Body jimplify(Body b, SootMethod m) {
         }
 
         Local gen = jimple.newLocal(localName, localType);
+        if (localDebugs.containsKey(parameterRegister)) {
+          gen.setName(localDebugs.get(parameterRegister).getFirst().name);
```
StevenArzt: again, make sure to avoid duplicates on obfuscated apps

src/main/java/soot/dexpler/DexBody.java
zxy-sourcebrella: Thanks! It is fixed now by generating a fresh name every time we create a local.

src/main/java/soot/dexpler/DexBody.java
zxy-sourcebrella: Fixed with `ArrayListMultimap` from guava.

mbenz89:
@StevenArzt are you happy with the PR now?

StevenArzt:
I am. We just need to check why the tests are failing.

mbenz89:
Thanks for contributing to the Soot project!

PR #1288, opened 2020-02-03
Making SCC generic and sort of iterative

MarcMil:


StevenArzt:
Just as additional explanation: We had to abolish the recursion, because we had StackOverflowErrors with huge method bodies

PR #1293, opened 2020-02-17
Feature/default methods

pavanupb:
A proposed fix for checking subclasses of the resolved methods as well.

The Test cases are from the previous commit. The test cases can be removed as the issue is already fixed.


mbenz89:
I think we should keep the test case if it checks correct behavior (even if we already fixed the issue and reached correct behavior). Please make a semantical copy of the test that is not a copy of the original puppycrawl code. We might run into licensing issues otherwise.

pavanupb:
> I think we should keep the test case if it checks correct behavior (even if we already fixed the issue and reached correct behavior). Please make a semantical copy of the test that is not a copy of the original puppycrawl code. We might run into licensing issues otherwise.

Okay. Yeah sure I will modify the test case

pavanupb:
Closing this pull request. A new pull request #1375 has been opened with changes.

PR #1300, opened 2020-02-25
Add gitpod config

ericbodden:
this commit adds support for Gitpod.io, a free automated
dev environment that makes contributing and generally working on GitHub
projects much easier. It allows anyone to start a ready-to-code dev
environment for any branch, issue and pull request with a single click.


ericbodden:
@anddann @mbenz89 do you see why this build fails? is this a flaky test maybe?

mbenz89:
The parallel build execution makes the tests non-deterministically fail due to shared workspaces. 
We already fixed this in the develop branch and will merge it back with the next release.

Generally, though, please do not merge or push directly to the master branch. Builds of the master branch will be published directly to Maven Central which does not allow re-deployment of artifacts. Therefore, either the version of our artifact has to change or the build will fail.

Of course, we can ignore the failed build but then the badge turns red indicating a build failure and Soot master is deviating from the actual release artifact. 

Eventually, we might want to not trigger the build if only the README or such changed. For now, please introduce changes to the develop branch first so we can merge them back with the next release.

ericbodden:
Hmm, I don't quite understand why you are writing about merges into the master branch. It was never my intention to merge into master, this pull request is intended to be merged into develop. Had I misconfigured this?

mbenz89:
It was targeted towards master before I changed it. No problem at all. Just wanted to make sure we are on the same page :).

.gitpod.yml
```diff
@@ -0,0 +1,2 @@
+tasks:
+  - init: mvn install -DskipTests=false
```
mbenz89: Why are you enabling the test cases? I would have expected the opposite since these take quite a while when starting up the IDE.

ericbodden: This was auto-generated. I agree it makes sense to set this to true.

PR #1301, opened 2020-02-26
Feature: Support Local Variable Annotation Tag

raintung:
Add the local variable annotation tag in the method to enhance analyze the annotation.

### DIFF
--- a/src/main/java/soot/asm/MethodBuilder.java
+++ b/src/main/java/soot/asm/MethodBuilder.java
@@ -24,11 +24,15 @@ package soot.asm;
 
 import com.google.common.base.Optional;
 
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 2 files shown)


mbenz89:
Looks good to me. 

Could you please fix the styling issues? Here is some documentation on how to do that: https://github.com/Sable/soot/blob/develop/CONTRIBUTING.md

I think just using the google code formatter plugin does the job since some time

raintung:
> Looks good to me.
> 
> Could you please fix the styling issues? Here is some documentation on how to do that: https://github.com/Sable/soot/blob/develop/CONTRIBUTING.md
> 
> I think just using the google code formatter plugin does the job since some time

ok, I will format it.

mbenz89:
Great! Thanks for contributing to Soot!

PR #1304, opened 2020-03-01
Fix long type signature in ClassConstant

michalbednarski:
"Lpackage/Class;" is type signature of class
"J" is type signature of long


mbenz89:
Wow, that's gone unnoticed for quite some time. Thanks for your fix!

PR #1308, opened 2020-03-08
Set the Method is Phantom while instructions is null

raintung:
Some classes's method instructions is empty, set the method is phantom status. Otherwise call AsmMethodSource convert the method body will throw nullpointexception.

### DIFF
--- a/src/main/java/soot/asm/MethodBuilder.java
+++ b/src/main/java/soot/asm/MethodBuilder.java
@@ -24,11 +24,15 @@ package soot.asm;
 
 import com.google.common.base.Optional;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import org.objectweb.asm.AnnotationVisitor;
 import org.objectweb.asm.Attribute;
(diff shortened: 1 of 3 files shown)


mbenz89:
Is this empty instruction set caused by an empty source code method or what would be the reason for having an empty instruction set?

If so, I would rather like to fix how the null pointer handling in when AsmMethodSource wants to convert the method since the method would not be phantom but rather empty in reality.

raintung:
> Is this empty instruction set caused by an empty source code method or what would be the reason for having an empty instruction set?

For some reason, will build empty instructions method that isn't abstract method or an empty source code(at least will has "return" opcode).
Those classes only work for compiling not running that I don't think it is reality method/classes. It is abnormal classes, so mark it is phantom. BTW, I really don't know howto make empty instructions method , it just appear in my testing envoriement.

linghuiluo:
see https://github.com/soot-oss/soot/pull/1618

PR #1309, opened 2020-03-09
Use linked variants of hash sets and maps.

michael-emmi:
Overall this patch is straightforward: replacing nearly all instances of three data structures — JDK’s HashMap, HashSet, Hashtable — with their linked variations — LinkedHashMap and LinkedHashSet.

Aside from that there are only two small changes to keep unit tests passing:
1. Replacing `contains` with `containsInAnyOrder` to remove dependence on container iteration order.
2. Adding two classes from `java.lang.invoke` to a list of distinguished “basic” classes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,12 +5,29 @@ name: Soot CI with Maven
 
 on:
   push:
-    branches: [ develop ]
   pull_request:
-    branches: [ develop ]
 
 jobs:
-  buildAndTest:
+  Codequality:
+    name: Check LGPL V2.1 headers´
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Use Java 8
+        uses: actions/setup-java@v1
+        with:
+          java-version: '8'
+      - name: Stylecheck
+        if: always()
+        run: |
+          mvn clean checkstyle:check -Dcheckstyle.failOnViolation=true
+      - name: Licensecheck
+        if: always()
+        run: |
+          mvn clean license:check-file-header -Dlicence-check.failOnMissingHeader=true
+    
+  BuildAndTest:
     name: Build and Test with java ${{ matrix.java_version }}
     runs-on: ubuntu-latest
     strategy:
@@ -18,12 +35,12 @@ jobs:
         java_version: ['8', '9', '11']
 
     steps:
-    - uses: actions/checkout@v1
+    - name: Checkout
+      uses: actions/checkout@v1
     - name: Use Java ${{ matrix.java_version }}
       uses: actions/setup-java@v1
       with:
         java-version: ${{ matrix.java_version }}
-
     - name: Build and test Java ${{ matrix.java_version }}
       run: |
         mvn clean test -PJava${{ matrix.java_version }}


mbenz89:
Hi. Thanks for your efforts in making Soot more deterministic. I'm struggling a bit with this patch since these linked data structures generally require a bit more space than the non-ordered one. In Soot's case, memory is an issue already, which is why I dislike making this the default. 

What do you think about introducing a factory for these datatypes and make it optional to use the ordered collections, so that the user can decide for her use-case if determinism is more important than space requirements?

michael-emmi:
Hey @mbenz89!

I think that introducing such a factory is technically feasible, and I am happy to update this PR to do exactly that. I suspect the intent might be to cover previously-patched hash maps and sets as well (?), and I could also include those.

Given that the motivation for this would come an existing issue about memory usage, I wonder whether it might be worth looking at the effect this patch would have on memory usage?

You probably wouldn’t want to introduce this extra factory unless it were really necessary, since that would be another user-level option (complicating Soot usage), and another burden on internal consistency (will the next would-be user of hash maps know to use this factory?)

So given that memory usage is already an issue: is there an issue tracking this? Are there any diagnostics or regressions? How does one figure out whether this PR makes memory usage in Soot worse for the benchmarks with which you’re concerned?

I know these questions may be unanswerable, and I am happy to just add the factory if you wish. Thought I should ask just in case :-)

michael-emmi:
Also, according to [Java Performance Tuning Guide](http://java-performance.info/memory-consumption-of-java-data-types-2/), JDK’s linked hash set consumes exactly the same amount of memory as its hash set, and its linked hash map consumes only 8 bytes more memory per element.

So, for instance, assuming one thousand elements were stored in hash maps through Soot on a given execution, that would only add an overhead of 8KB; one million elements would incur 8MB overhead; etc.

ericbodden:
Hi all.

> Also, according to [Java Performance Tuning Guide](http://java-performance.info/memory-consumption-of-java-data-types-2/), JDK’s linked hash set consumes exactly the same amount of memory as its hash set, and its linked hash map consumes only 8 bytes more memory per element.
> 
> So, for instance, assuming one thousand elements were stored in hash maps through Soot on a given execution, that would only add an overhead of 8KB; one million elements would incur 8MB overhead; etc.

If that is so then I think the impact should be rather benign. In any case I agree that if anything we should measure the impact before we go with factories.

michael-emmi:
Are there any particular tests one could target for such profiling?

mbenz89:
> I think that introducing such a factory is technically feasible, and I am happy to update this PR to do exactly that. I suspect the intent might be to cover previously-patched hash maps and sets as well (?), and I could also include those.

Yes, that would be perfect!

> Given that the motivation for this would come an existing issue about memory usage, I wonder whether it might be worth looking at the effect this patch would have on memory usage?

I agree that we should measure the impact of that change first. 

> Also, according to Java Performance Tuning Guide, JDK’s linked hash set consumes exactly the same amount of memory as its hash set, and its linked hash map consumes only 8 bytes more memory per element.

I profiled Soot some time ago and if I do not mix up things here, a huge chunk of memory was consumed due to map structures. Considering your source, it really doesn't sound like using ordered structures would also have a huge impact on memory usage. 

> You probably wouldn’t want to introduce this extra factory unless it were really necessary, since that would be another user-level option (complicating Soot usage), and another burden on internal consistency (will the next would-be user of hash maps know to use this factory?)

I agree that this introduces maintenance effort. We would have to take care of contributors using the factory and put the burden to decide if it makes sense to use it in their specific use-case.

> Are there any particular tests one could target for such profiling?

We observed memory issues with Soot especially in the context of commercial software projects that outscale any open-source project I know by far. We are already thinking about how we could lower Soot's memory footprint which is why I hesitate a bit on such changes. The impact, however, may not be as huge as I initially feared.

Unfortunately, we do not have any test cases that would make sense for such a benchmakr. Analyzing the JDK itself is always a good starting point. Other than that, I guess you could just take a library from maven central and analyze it together with all its transitive dependencies. That should also lead to a rather huge amount of classes and Jimple Code.

michael-emmi:
@mbenz89 given the lack of adequate performance benchmarks, I have implemented your suggestion to use factory methods.

michael-emmi:
@mbenz89 I’m having trouble rebasing this one due to the following merge, which reindented an entire file (Scene.java).

https://github.com/Sable/soot/commit/b6a74bd9c44984949e03ccf19881c6c12de83cf6

Even if I actually go through the tedious work of rebasing against these commits (which I have done locally), then my build reports 1461 check style violations due to the new indentation.

Should I just ignore these warnings and push my rebased changes anyhow?

mbenz89:
Hi @michael-emmi, thanks for going through that effort! 

It seems that a push from a colleague has put the Scene into a state that violates our style guidelines. Therefore, the problem is not on your side. Sorry you already investigated the work for applying this style. I will fix the styling issues now and you can target your changes against the correct style. I guess you can just use the version of your change before you rebasing against the stated commit.

mbenz89:
I fixed the formatting in 187c5f879a82873cd987397396ebdbc254a7a04d

michael-emmi:
@mbenz89 thanks!

I’ve just rebased and everything seems fine now, pending your approval.

michael-emmi:
@ericbodden @mbenz89 any idea when this might be merged?

mbenz89:
I think we all agreed that we should measure the performance impact of ordered collections before going with a factory. I can run some tests when I find time for it. Since I'm currently lacking time a lot and there are quite some other topics left on my agenda, this might still take some time.

Would you have time to investigate the memory footprint for some larger projects? This won't be perfect but we could try to jimplify the JDK and build a callgraph and then before/after gc measure the memory footprint for both cases to get some idea. If the JDK is not large enough, I would just throw some larger libraries in the mix.

michael-emmi:
@mbenz89 I may have misunderstood the conversation above, but my take-aways are:
1. The concern for potential overhead with linked hash maps is *memory* overhead, not time.
2. We collectively have no way to determine the memory-overhead impact for clients of Soot.
3. The factory-based solution you suggest nullifies that concern, since using linked hash maps becomes a Soot option.

(I assume we are not concerned with the runtime performance overhead in the factory code itself, i.e., deciding whether to instantiate a linked or regular hash map. However, in case that is a concern, I will update the factory to make that decision at factory-instantiation time instead.)

So doesn’t the factory-based solution obviate the need for more rigorous performance testing?

ericbodden:
Michael I am not at all a fan of the factory-based solution. It would be horrible to maintain in the future because we would need to somehow enforce that factories are, in fact, being used.

For that reason I would like to see some memory measurements to see if the footprint is, in fact, negligible. Then we could go without factories.

StevenArzt:
I am usually a big supporter of factories (in fact, it was my team that introduced factories for SootMethod, etc.). In this special case, however, I doubt that it makes sense, for the following reasons.

- We might easily introduce new code that accidentially misses the factory and directly creates an instance. For something as basic as a set, people usually don't assume there is a factory.
- Some algorithms use particular collections for a reason, such as having ordering contraints (or none of them), performance requirements on particular access patterns, etc. A generic factory cannot really help here.

Unless there is a very good reason to make this change, I don't think we should do it.

@ericbodden agreed.

mbenz89:
I totally see your point, maintaining this is prone to error and would be a major effort.

I think we should measure the impact of these ordered structures to get at least a coarse-grained view on memory requirements.

> Some algorithms use particular collections for a reason, such as having ordering contraints (or none of them), performance requirements on particular access patterns, etc. A generic factory cannot really help here.

This problem would eliminate the possibility of such a bulk update of used collections altogether (independent of if we use a factory or not). I think there are a few algorithms that require ordered-collections. Do you know of opposite cases, where unordered collections are required @StevenArzt ?

StevenArzt:
> This problem would eliminate the possibility of such a bulk update of used collections altogether (independent of if we use a factory or not). I think there are a few algorithms that require ordered-collections. Do you know of opposite cases, where unordered collections are required @StevenArzt ?

Not exactly for ordering. However, not requiring an ordering can help with reducing memory consumption. Sometimes you rather take an array, or bitsets, or a HashSet, and would waste memory with a linked set.

michael-emmi:
Wanting to eschew the factory here makes perfect sense, and aligns with my initial argument against factories (see above).

But in that case I do not understand how to proceed with this PR, since:
* The criterion of “acceptable performance overhead” in this case is not clear to me; e.g., what might be the required set of benchmarks, and threshold (?).
* At the moment I do not have the resources nor expertise to conduct benchmarking.

I would greatly appreciate your guidance on next steps!

ericbodden:
@michael-emmi Manuel wrote:

> Would you have time to investigate the memory footprint for some larger projects? This won't be perfect but we could try to jimplify the JDK and build a callgraph and then before/after gc measure the memory footprint for both cases to get some idea. If the JDK is not large enough, I would just throw some larger libraries in the mix.

Is this something you could do? Or would you not find the time?

michael-emmi:
> @michael-emmi Manuel wrote:
> 
> > Would you have time to investigate the memory footprint for some larger projects? This won't be perfect but we could try to jimplify the JDK and build a callgraph and then before/after gc measure the memory footprint for both cases to get some idea. If the JDK is not large enough, I would just throw some larger libraries in the mix.
> 
> Is this something you could do? Or would you not find the time?

At the moment I do not have the resources nor expertise to do this.

mbenz89:
@michael-emmi Hi! Is there a reason you deleted the PR? In general, I would be happy to have this kind of determinism in Soot, since non-determinism is something a lot of people already had their problems with during experimental evaluations. Unfortunately, we just have a lack of manpower here to get this quickly tested.

@pavanupb would it make sense to use the test cases we used to test the performance impact of the synchronized maps in `Scene` to also test memory footprints among different versions of Soot?

pavanupb:
> @michael-emmi Hi! Is there a reason you deleted the PR? In general, I would be happy to have this kind of determinism in Soot, since non-determinism is something a lot of people already had their problems with during experimental evaluations. Unfortunately, we just have a lack of manpower here to get this quickly tested.
> 
> @pavanupb would it make sense to use the test cases we used to test the performance impact of the synchronized maps in `Scene` to also test memory footprints among different versions of Soot?

@mbenz89 Yeah we could do that.

michael-emmi:
@mbenz89 I did not mean to delete/close this PR — it may have been automatically closed when I revoked the changes to my fork’s branch. Apologies for that!

The reasons I revoked those changes are:
1. Despite this PR’s intention to further reduce nondeterminism, a long-running experiment revealed that this PR was not quite effective in doing that.
2. As you also mentioned above, the factory-based solution is not ideal. If this PR were to move forward, it would be better to do it without the CollectionFactory.

As a side-effect of the aforementioned experiment, I noticed no significant impact on memory overhead using with the linked- variants of hash sets and maps over several dozen runs on sizable codes (for which loading classes alone can take several minutes).

mbenz89:
> Despite this PR’s intention to further reduce nondeterminism, a long-running experiment revealed that this PR was not quite effective in doing that.

That's unfortunate. We should effectively eliminate the non-determinism. Otherwise, I do not see a reason to merge this. However, I think this is a step in the correct direction!

> As you also mentioned above, the factory-based solution is not ideal. If this PR were to move forward, it would be better to do it without the CollectionFactory.

I agree. We should further test the impact on memory usage. As you theoretically and practically testified, the impact should be neglectible. If this stays true, and we fully eliminate non-determinism with it, I do not see a reason against using ordered collections all over the place.

> Not exactly for ordering. However, not requiring an ordering can help with reducing memory consumption. Sometimes you rather take an array, or bitsets, or a HashSet, and would waste memory with a linked set.

@StevenArzt that's why I am hesitating to go on with this change. We need to investigate the effective impact on Soot's memory footprint. If we, however, notice that the difference between ordered and non-ordered collections is indeed neglectable, would you see any other points that are against introducing ordered collections to gain a more deterministic version of Soot?

Issue #1313, opened 2020-03-13
java.lang.RuntimeException when using CFGViewer command line option


TealTulips:
Steps to reproduce:
1.) javac -d ./target/classes -cp ./soot-3.3.0.jar ./src/main/java/staticAnalysis/FindPrimeFactor.java
2.) java -cp /Users/xyz/eclipse-workspace/staticAnalysis/sootclasses_j9-trunk-jar-with-dependencies.jar soot.tools.CFGViewer -cp "." -pp --graph=BriefUnitGraph --ir=grimp -p jb use-original-names:true FindPrimeFactor

Files used to reproduce: 

import java.util.Arrays;
import java.util.List;

public class FindPrimeFactor {
   	public static void main(String []args) {
		
	    List<Integer>a= Arrays.asList(12, 25, 18, 8); //happy path
	    List<Integer> b = Arrays.asList(6, 2, 3, 2);
	   System.out.println("INDEX "+findPrimeFactor(a,b));

	}
	

    public static int findPrimeFactor(List<Integer> a, List<Integer> b) throws IllegalArgumentException{
    	   	
    	try {
    	    if (a == null) 
    	        throw new NullPointerException();
    	    }    	
    	catch(NullPointerException e) {
    		throw new IllegalArgumentException("List a cannot be null");
    	}
    	try {
    		if (b == null) 
    	        throw new NullPointerException();    	}
    	catch(NullPointerException e) {
    		throw new IllegalArgumentException("List b cannot be null");
    	}
        for(int i = 0; i < b.size() && i < a.size(); i++) {
        	if(a.get(i)==null||b.get(i)==null|| b.get(i).equals(0))
        	 i++;
        	if(isPrime(b.get(i)) == true && a.get(i) % b.get(i) == 0) {
                return i;
            }
        }
		throw new IllegalArgumentException("List b doesn't have prime factor of any element in list a.");

    }
   
    public static boolean isPrime(int num) {
	    int m,n,test=1;  
    	if(num==1){
		test=2;
		testFunc(test);
    		return false;
	}
	if(num<5){

		 m =6;
		 //return true;
	}
	else
	{	 n = 7;
	test =5;
		testFunc(test);
	
    	for(int i = 2; i < num; i++) {
            if(num % i == 0) {
                return false;
            }
        }}
        return true;
    }

    public static void testFunc(int test){
	    System.out.println("TEST HERE "+test);

    }
}

Soot version: <pre>sootclasses_j9-trunk</pre>

[...]

PR #1314, opened 2020-03-13
Fix/1313

anddann:
Fixes issue #1313


mbenz89:
Thanks for the fix, Andreas! The PR was targeted towards master. After changing the target to develop, some merge issues remain. The easiest thing might be to make a new pr targeted towards develop where you make your change in develop to omit merge issues.

mbenz89:
outdated by #1315

PR #1333, opened 2020-04-04
Feature/default methods

pavanupb:


src/main/java/soot/SootMethodRefImpl.java
```diff
@@ -190,68 +199,46 @@ public SootMethod tryResolve() {
...
-      // using soot.SootClass.getMethodUnsafe(java.lang.String, java.util.List<soot.Type>, soot.Type)
-      // instead of soot.SootClass.getMethodUnsafe(soot.util.NumberedString)
-      // to avoid false positive on method/class renaming
-      final SootMethod method = selectedClass.getMethodUnsafe(name, parameterTypes, returnType);
-      if (method != null) {
-        checkStatic(method);
-        return method;
-      }
-
-      if (Scene.v().allowsPhantomRefs() && selectedClass.isPhantom()) {
```
pavanupb, line 216: @mbenz89 Why "(Scene.v().allowsPhantomRefs() && selectedClass.isPhantom())" is being changed to "(Scene.v().allowsPhantomRefs() && Options.v().ignore_resolution_errors())" in default?

mbenz89: If you look more closely, you'll see, that the change is rather the removal of the first condition `Scene.v().allowsPhantomRefs() && selectedClass.isPhantom()` from develop but not a replacement by `Scene.v().allowsPhantomRefs() && Options.v().ignore_resolution_errors()`. In develop, the exact same code resides: https://github.com/Sable/soot/blob/develop/src/main/java/soot/SootMethodRefImpl.java#L251

However,  in the default branch, I replaced the first half of the `tryResolve` method with a call to `FastHierarchy.resolve`. The `FastHierarchy.resolve` call now cares about dispatch and method resolution. It does not take care of phantom classes if I remember corrrectly. Can you please double check if the code in [this block](https://github.com/Sable/soot/blob/develop/src/main/java/soot/SootMethodRefImpl.java#L216) was moved to `FastHierarchy.resolve`? If not, we might net to model this behavior there as well (only with if a parameter indicating this behavior is set, though). So that `SootMethodRefImpl.tryResolve` can opt to add methods to phantom classes and the usual dispatch behavior does not.

pavanupb: @mbenz89 The check to phantom has not been moved to resolveMethod in FastHierarchy, which is the reason I could see a lot of edges in default that is not present in develop.

mbenz89: Could you explain some examples where this makes a difference?

Could you implement the phantom behavior optionally and only for the call from ´SootMethodRefImpl´ and see if that solves your problem?

pavanupb:
Closing this pull request as it was opened only for discussion

PR #1334, opened 2020-04-06
Develop

pavanupb:


src/main/java/soot/jimple/toolkits/callgraph/VirtualCalls.java
```diff
@@ -108,6 +108,7 @@ public SootMethod resolveNonSpecial(RefType t, NumberedString subSig, boolean ap
 
     SootMethod m = cls.getMethodUnsafe(subSig);
     if (m != null) {
+    	// Why are we not adding to targets for abstract methods
       if (!m.isAbstract()) {
         ret = m;
```
pavanupb, line 113: @mbenz89 Can there be an edge from one method to another method within the same abstract class?. I am not sure why are we not adding to targets if a method is abstract.

mbenz89:
This is not a real pr, i.e., should not be merged, right?

src/main/java/soot/jimple/toolkits/callgraph/VirtualCalls.java
mbenz89: Because an edges always has to point to a method that contains actual code! An non-default interface method or an abstract method, however, cannot contain any code that is executed. Wenn dispatching, we want to find the concrete method that is called at a given callside. This can never be an abstract method?

Important: The condition checks if the _method!_ is abstract not the class. We can of course have edges to methods in abstract classes as long as they contain an implementation. Not if the method is abstract still.

pavanupb: @mbenz89 Thanks for the clarification. The default implementation is allowing edges between two abstract methods. So I guess a check needs to be added in FastHierarchy to check if the method is abstract before adding to targets.

pavanupb:
> This is not a real pr, i.e., should not be merged, right?

Yes, this is not a real PR.

src/main/java/soot/jimple/toolkits/callgraph/VirtualCalls.java
mbenz89: Oh, that should not be possible. Can you show me the specific code? I thought that would only be possible if `allowAbstract` is set to true. This should only be the case if `FastHierarchy.resolveMethod` is called from `SootMethodRefImpl`. Isn't that correct? Is this consistent between default and develop?

pavanupb: @mbenz89 There is a check to abstract method in FastHierarchy at https://github.com/Sable/soot/blob/feature/default-methods/src/main/java/soot/FastHierarchy.java#L859. But still we are returning the candidate at FastHierarchy which is the reason getting added to targets at https://github.com/Sable/soot/blob/feature/default-methods/src/main/java/soot/FastHierarchy.java#L884.

mbenz89: How is this different to https://github.com/Sable/soot/blob/fcc380a506258291d6fe6954425586a4ae4968ea/src/main/java/soot/jimple/toolkits/callgraph/VirtualCalls.java#L112 and https://github.com/Sable/soot/blob/fcc380a506258291d6fe6954425586a4ae4968ea/src/main/java/soot/jimple/toolkits/callgraph/VirtualCalls.java#L122?

pavanupb: In https://github.com/Sable/soot/blob/fcc380a506258291d6fe6954425586a4ae4968ea/src/main/java/soot/jimple/toolkits/callgraph/VirtualCalls.java#L112, if a method is resolved by calling getMethodUnsafe(), and if the method is abstract we are not assigning it the method to ret. So ret would be null.
Whereas that's not the case in default branch.

mbenz89: In default, we would look into the other interfaces for an implementation instead. Isn't that correct? After all, the interafcae implementatin hierarchy is kind of a tree where the method to resolve is the root. We traverse all paths from the root to find the closest (most specific) implementation of the method. If the method is abstract (not-implemented) in one interace, it might still be in another. Do I miss something here?

pavanupb: Yeah, that's correct. But what if we did not find the implementation when traversed through the hierarchy. We are still returning the candidate which is an abstract method.

mbenz89: Ahhh. Now I understand! Can you please fix this in your already opened merge request for the other fix?

mbenz89: Wait, no. How can candidate ever be set to an abstract method?

pavanupb: Its been assigned here at https://github.com/Sable/soot/blob/feature/default-methods/src/main/java/soot/FastHierarchy.java#L818

mbenz89: Perfect! Please set `candidate` to `null` here: https://github.com/Sable/soot/blob/feature/default-methods/src/main/java/soot/FastHierarchy.java#L823

pavanupb: Okay.

src/main/java/soot/jimple/toolkits/callgraph/VirtualCalls.java
```diff
@@ -239,13 +240,19 @@ protected void resolveAnySubType(Type declaredType, Type sigType, NumberedString
...
+    Set<Type> newSubTypes = new HashSet<>();    
+    
 
     LinkedList<SootClass> worklist = new LinkedList<SootClass>();
     HashSet<SootClass> workset = new HashSet<SootClass>();
     FastHierarchy fh = fastHierachy;
     SootClass cl = base.getSootClass();
+    
+    //Do not add abstract classes and interfaces as they do not contain implementations
+    if(!cl.isAbstract() && !cl.isInterface())
```
pavanupb, line 252: @mbenz89 Added a condition to check if the class is not abstract and not an interface. After adding the check the number of edge difference got reduced from 429 to 143.

I am thinking we should still be allowing abstract classes to get added as they may contain an implementation because an abstract class can contain non abstract methods as well right?

mbenz89: You are right, it depends on if the method is abstract or not. But for this case, please just change it back as in the original develop branch. We want to compare to that state.

pavanupb: Okay. So I should not be making any changes to develop but just check why those differences are present right?

mbenz89: Exactly. We want to know if there are issues in the pr. We would only fix something if it is not fixed by the pr anyway.

mbenz89:
Would you be able to write some test cases for the issues we already fixed? So we can make sure we do not break them again in the future?

How long would this take?

How long will it take to review the remaining edges?

pavanupb:
@mbenz89 Why java.lang.Enum is missing in addSootBasicClasses() in Default branch at https://github.com/Sable/soot/blob/961d90ca52bee5fdd7b41ab69a1e82786098ced0/src/main/java/soot/Scene.java#L1601, While it is present in Develop Branch at https://github.com/Sable/soot/blob/2d5d2ed3e6f2c957774d1a0178a9d73d971d4656/src/main/java/soot/Scene.java#L1678?. This is causing around 20 -30 edge differences between default and develop branch

mbenz89:
uhhh. migh it be that your dev branch version is newer than the default branch? i guess it was added after i forked for the patch. btw. this is why we should get it merged fast. I will have to merge quite some stuff the longer we wait :)

pavanupb:
> uhhh. migh it be that your dev branch version is newer than the default branch? i guess it was added after i forked for the patch. btw. this is why we should get it merged fast. I will have to merge quite some stuff the longer we wait :)

:) :) I guess we are almost at the end of getting the branch merged. Maybe I will manually fix and re-run soot to find other edge differences

pavanupb:
> > uhhh. migh it be that your dev branch version is newer than the default branch? i guess it was added after i forked for the patch. btw. this is why we should get it merged fast. I will have to merge quite some stuff the longer we wait :)
> 
> :) :) I guess we are almost at the end of getting the branch merged. Maybe I will manually fix and re-run soot to find other edge differences


After adding the java.lang.Enum to basicClasses, the number of edge differences got reduced from 152 to 30 :)

mbenz89:
Awesome!

Pavan Gurkhi Bhimesh <notifications@github.com> schrieb am Mo., 18. Mai
2020, 21:32:

pavanupb:
Closing this pull request as it was opened only for discussion

Issue #1336, opened 2020-04-09
UnconditionalBranchFolder not thread-safe although used by multiple threads


ingo-budde:
We encountered a `NullPointerException` that appears sporadically and we wanted to share our bugfix and therefore also opened [a merge request](https://github.com/Sable/soot/pull/1337). The exception is:

```java
java.lang.NullPointerException
	at soot.jimple.toolkits.scalar.UnconditionalBranchFolder.internalTransform(UnconditionalBranchFolder.java:74)
	at soot.BodyTransformer.transform(BodyTransformer.java:51)
	at soot.Transform.apply(Transform.java:120)
	at soot.BodyPack.internalApply(BodyPack.java:50)
	at soot.Pack.apply(Pack.java:125)
	at soot.PackManager.runBodyPacks(PackManager.java:1009)
	at soot.PackManager.lambda$runBodyPacks$0(PackManager.java:660)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

### Problem
In our understanding (which might be wrong), it could happen that the same instance of the class `UnconditionalBranchFolder` is used by multiple threads. It is this assumption that led us to the following bug explanation:

The class `PackManager` spawns multiple threads that perform body transformations on the JIMPLE Code and the `UnconditionalBranchFolder` is one class that is part of such a body transformation.

The `UnconditionalBranchFolder` class creates two integer arrays `numFound` and `numFixed`, which it stores as attributes. This is done on-demand, meaning that the arrays are allocated as soon as the first thread requires access to it [1].

The NullPointerException can happen if one thread already allocated the `numFound` array in line 68, but not yet the `numFixed` array in line 69:
```java
    // allocate counters once only
    if (numFound == null) {
      numFound = new int[JUMPOPT_TYPES + 1];
      numFixed = new int[JUMPOPT_TYPES + 1];
    }
```

Then, another thread skips the allocation, because the if-statement in line 67 evaluates to `false` and as `numFixed` is still `null`, it runs into a null pointer access in line 74.

There are other race-conditions possible, but they currently stay unobserved. For example, it could be that one thread started to store values into such array in the for loop, but in the meantime the array is reassigned by another thread, so that the calculated values until that point are lost and the analysis continues on invalid data.

[...]

PR #1337, opened 2020-04-09
Make the UnconditionalBranchFolder class stateless to avoid concurrency issues

ingo-budde:
In issue #1336 we reported a `NullPointerException` that appears sporadically as part of a race condition and that we have a created a bugfix (Pull Request) for:

```java
java.lang.NullPointerException
	at soot.jimple.toolkits.scalar.UnconditionalBranchFolder.internalTransform(UnconditionalBranchFolder.java:74)
	at soot.BodyTransformer.transform(BodyTransformer.java:51)
	at soot.Transform.apply(Transform.java:120)
	at soot.BodyPack.internalApply(BodyPack.java:50)
	at soot.Pack.apply(Pack.java:125)
	at soot.PackManager.runBodyPacks(PackManager.java:1009)
	at soot.PackManager.lambda$runBodyPacks$0(PackManager.java:660)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

### Problem
In our understanding (which might be wrong), it could happen that the same instance of the class `UnconditionalBranchFolder` is used by multiple threads. It is this assumption that led us to the following bug explanation:

The class `PackManager` spawns multiple threads that perform body transformations on the JIMPLE Code and the `UnconditionalBranchFolder` is one class that is part of such a body transformation.

The `UnconditionalBranchFolder` class creates two integer arrays `numFound` and `numFixed`, which it stores as attributes. This is done on-demand, meaning that the arrays are allocated as soon as the first thread requires access to it [1].

The NullPointerException can happen if one thread already allocated the `numFound` array in line 68, but not yet the `numFixed` array in line 69:
```java
    // allocate counters once only
    if (numFound == null) {
      numFound = new int[JUMPOPT_TYPES + 1];
      numFixed = new int[JUMPOPT_TYPES + 1];
    }
```

Then, another thread skips the allocation, because the if-statement in line 67 evaluates to `false` and as `numFixed` is still `null`, it runs into a null pointer access in line 74.

There are other race-conditions possible, but they currently stay unobserved. For example, it could be that one thread started to store values into such array in the for loop, but in the meantime the array is reassigned by another thread, so that the calculated values until that point are lost and the analysis continues on invalid data.

Therefore we decided to not use attributes in this case, but to have each execution of the `internalTransform()` method to operate on its own datastructure.

[1] https://github.com/Sable/soot/blob/master/src/main/java/soot/jimple/toolkits/scalar/UnconditionalBranchFolder.java#L66

### Our solution
Instead of storing the "state" in instance variables, the state is stored in local variables and passed to other methods via parameters methods as needed.

### DIFF
--- a/src/main/java/soot/jimple/toolkits/scalar/UnconditionalBranchFolder.java
+++ b/src/main/java/soot/jimple/toolkits/scalar/UnconditionalBranchFolder.java
@@ -52,9 +52,6 @@ public class UnconditionalBranchFolder extends BodyTransformer {
   }
 
   static final int JUMPOPT_TYPES = 6;
-  int numFound[], numFixed[];
-
-  HashMap<Stmt, Stmt> stmtMap;
 
   protected void internalTransform(Body b, String phaseName, Map<String, String> options) {
     StmtBody body = (StmtBody) b;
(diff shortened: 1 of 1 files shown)


mbenz89:
Hi Ingo. 

Thanks for the great issue report and the pull request, it is highly appreciated!

Generally speaking, Soot is full of such race condition issues since the jimplification was not always parallelized. Some time ago it got parallelized, but not all the code was adapted accordingly. We still run into such race conditions from time to time. 

Your fix makes total sense to me. Indeed, since `numFixed ` and `numFound ` are only used for body-based reporting, sharing them over multiple threads does not make sense at all.

marcus-h:
1337... what a great number for a PR:) *SCNR*

ingo-budde:
Thanks also go to Marcus Hüwe for creating the bugfix.

PR #1341, opened 2020-04-24
Fix/innerclass

anddann:
Currently the handling of inner classes is bugy, and inner classes are not properly detected.
This PR provides a fix and test case


anddann:
adapted to outerclass computation, to the one in the jimple reader

PR #1368, opened 2020-06-11
Fix 1367 constructor references

### LINKED ISSUE #1367: LambdaMetaFactory generating invalid jimple bodies for constructor references
Hello!

This issue is possibly related to #1292.

I am using Soot 4.1.0 and openjdk 1.8.0_222, and I am trying to perform some analysis on Java code that uses constructor references:

```
public class A {

    public Supplier<Object> constructorReference() {
        return Object::new;
    }

}
```

Soot recognizes the resulting `invokedynamic` instruction and creates a Supplier instance via LambdaMetaFactory. The following jimple body is created for the supplier `get()` method:

```
    public java.lang.Object get()
    {
        A$init__1 $r0;
        java.lang.Object $r1;

        $r0 := @this: A$init__1;

        $r1 = new java.lang.Object;

        specialinvoke $r1.<java.lang.Object: void <init>()>();

        return;
    }
```

I would expect that the method should return $r1 (the newly created object), as per the method's signature, but a void return statement is inserted instead.  The [comments in LambdaMetaFactory](https://github.com/Sable/soot/blob/3b5bcb09f536e2676e9f781405f852f0f7cde249/src/main/java/soot/LambdaMetaFactory.java#L783-L785) indicate that the void return statement may be intentional in the specific case of a constructor reference.  This seems strange, since the method signature and return type do not match.

I've noticed that the Jimple validators allow for this sort of disagreement between method signature and return type, so I imagine there must be some cases where it's desirable.  For my use case, I am performing some analysis on the jimple representation and then generating transformed class files, so it is causing some issues for me with invalid bytecode.

Any idea if this is intentional behaviour or whether it's something I could fix?

dseekatz:
Fixes #1367 

- Added simple check to `ReturnStatementsValidator` that requires methods with non-void return type to have a non-void return statement and methods with void return type to have a void return statement
- Properly handle constructor references as a special case in `LambdaMetaFactory`
- Added a new test case

### DIFF
--- a/src/main/java/soot/LambdaMetaFactory.java
+++ b/src/main/java/soot/LambdaMetaFactory.java
@@ -773,28 +773,31 @@ public final class LambdaMetaFactory {
         JimpleBody jb, PatchingChain<Unit> us, LocalGenerator lc, List<Local> args) {
       Value value = _invokeImplMethod(jb, us, lc, args);
 
-      if (soot.VoidType.v().equals(implMethod.getMethodRef().getReturnType())) {
+      if (soot.VoidType.v().equals(implMethodType.getReturnType())) {
         // dispatch method is void
         if (value instanceof InvokeExpr) {
           us.add(Jimple.v().newInvokeStmt(value));
         }
 
+        us.add(Jimple.v().newReturnVoidStmt());
+
+      } else {
+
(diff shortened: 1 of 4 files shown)


mbenz89:
Thanks for investigating and proposing a fix!

PR #1391, opened 2020-07-23
Fix typo

CirQ:


ericbodden:
Thanks! Sorry, we were too slow merging this, due to vacation. If it's not too much trouble, would you kindly update this to the current branch?

CirQ:
Sure, I've updated it (by clicking the button under the code checking list), any thing else should I do?

PR #1394, opened 2020-07-30
Make getThrowAnalysis() public, format some comments

Waldo1637:
Aside from general usage, this method must be public to allow a subclass outside of the current package to define a constructor that has an ExceptionalUnitGraph parameter and to obtain its ThrowAnalysis in order to store that in its own field:
```
public class SubclassExceptionalUnitGraph extends ExceptionalUnitGraph {
    public SubclassExceptionalUnitGraph(ExceptionalUnitGraph graph) {
        ...
        this.throwAnalysis = graph.getThrowAnalysis();
    }
}
```


Waldo1637:
Will update and resubmit.

PR #1396, opened 2020-07-30
BUGFIX: Cleanup and modernize graph classes and fix clone bug

Waldo1637:
Add override annotations, add generics, remove unnecessary casts, convert StringBuffer to StringBuilder, convert size()==0 checks to isEmpty() checks, use foreach loops, fix typos, etc.


Waldo1637:
Will update and resubmit.

PR #1398, opened 2020-07-31
BUGFIX: fix several bugs in Validator classes

tim-hoffman:
CheckTypesValidator was appending ```in 'method-sig'``` twice to most validation lines and the NewValidator was creating and filling an extra list that appears to be left over from a prior refactoring.


Waldo1637:
Also removed an immediate return from the InvokeValidator. It looks like it's been this way as far back as I looked in source control but I couldn't see any reason why it was like this. It's not currently used anywhere in the Soot project so this shouldn't immediately break anything but it does allow the validator to actually be used if desired.

Waldo1637:
Another thing, several validators directly threw their exceptions rather than adding them to the exception list as expected. Changed them to append the list like a validator should.

Waldo1637:
It turns out ```InvokeValidator``` had multiple bugs. I guess that's why the immediate return was there before! However, I believe I have fixed them and it seems to be working correctly now.

PR #1399, opened 2020-07-31
Move Android code to reduce runtime dependencies

Waldo1637:
Move the code using pxb.android.axml.* into the AndroidVersionInfo class so that the AXML jar is not necessary when running Soot on non-Android inputs (i.e. the AndroidVersionInfo  class will not be loaded thus the references in that package will not be seen).
Also some code modernizations (foreach loops and try-with-resources) and optimizations (use StringBuilder, switch over String, isEmpty() and other String operations, etc.)


Waldo1637:
Will resolve conflicts and resubmit.

PR #1401, opened 2020-07-31
BUGFIX: Cleanup and modernize Body and all subclasses

tim-hoffman:


Waldo1637:
I'm not sure why the Java 11 tests failed when I added that last commit just now (checks were successful after the first 2 commits). The log shows ```ConcurrentModificationException``` via ```FastHierarchy.getSignaturePolymorphicMethod(...)``` but there is no way the minor changes in that commit could cause it.

tim-hoffman:
Will update and resubmit.

PR #1403, opened 2020-08-03
Cleanup, modernize, and optimize some classes in soot.*

Waldo1637:
use foreach loop, switch instead of chained if statements, String.isEmpty(), etc.


Waldo1637:
Closing due to test failure.

PR #1412, opened 2020-08-07
BUGFIX: Cleanup shimple classes, and fix a couple bugs

tim-hoffman:
In several classes, removed the fields caching objects from the ```ShimpleFactory``` because in a scenario where the body was modified and the factory was reset, these fields would contain stale data (and there was little benefit to caching them anyway). Also removed unused fields from ```DefaultShimpleFactory``` and the unused ```GuaranteedDefs``` from ```PhiNodeManager```.


Waldo1637:
I believe more can be done in the managers and builder but I didn't want to mess with their public API. However, I do think that public API is too large. Is it really necessary for methods such as ```PiNodeManager.insertPiNodes(Local,Block)``` and ```ShimpleBodyBuilder.renameLocalsSearch(Block)``` to be public? These (and others) seem like more of internal helper methods to me but I thought best to have a discussion before removing them.

PR #1419, opened 2020-08-24
Cleanup, modernize, and optimize AbstractJasminClass and subclasses

tim-hoffman:


tim-hoffman:
will update and resubmit

PR #1433, opened 2020-08-26
Code cleanup in some files in soot.toolkits.graph

Waldo1637:
Cleanup code in remaining files (i.e. those not touched by any other recent PR) in soot.toolkits.graph


ericbodden:
@mbenz89 do you see why the status of the license check remains at "Expected — Waiting for status to be reported"? Is something broken?

Waldo1637:
I see this happening on all pull requests and figured there must be some configuration issue.

Waldo1637:
It seems some dependencies are failing to download for some reason. For example:
```
[ERROR] Failed to execute goal on project soot: Could not resolve dependencies for project org.soot-oss:soot:jar:4.3.0-SNAPSHOT: Failed to collect dependencies at de.upb.cs.swt:axml:jar:2.1.0-SNAPSHOT: Failed to read artifact descriptor for de.upb.cs.swt:axml:jar:2.1.0-SNAPSHOT: Could not transfer artifact de.upb.cs.swt:axml:pom:2.1.0-SNAPSHOT from/to sonatype-snapshot (https://oss.sonatype.org/content/repositories/snapshots): Transfer failed for https://oss.sonatype.org/content/repositories/snapshots/de/upb/cs/swt/axml/2.1.0-SNAPSHOT/axml-2.1.0-SNAPSHOT.pom 503 Service Unavailable
```

src/main/java/soot/toolkits/graph/CytronDominanceFrontier.java
```diff
@@ -109,35 +110,28 @@ protected void bottomUpDispatch(DominatorNode<N> node) {
    *              if (idom(Y)!=X) then DF(X) <- DF(X) U Y
    *        end
    *      end
+   * </code>
    * </pre>
-   **/
+   */
   protected void processNode(DominatorNode<N> node) {
-    List<DominatorNode<N>> dominanceFrontier = new ArrayList<DominatorNode<N>>();
+    HashSet<DominatorNode<N>> dominanceFrontier = new HashSet<DominatorNode<N>>();
```
StevenArzt, line 117: Why a set?

Waldo1637: The Cytron algorithm defines the dominance frontier as an unordered set. The issue with using a List in the implementation is that some nodes may be inserted more than once. Since the tree is traversed bottom-up, using the results from lower nodes in the tree to compute the result for higher nodes, the duplication is especially disastrous when it occurs lower in the tree because it can be duplicated again for each level up the tree thus exponentially increasing the size of the List, with each entry being duplicated many times.

Waldo1637: I ran into this issue while processing part of Oracle JDK 8 (I don't remember which class/method specifically) and replacing the List with a Set significantly improved performance while still giving the correct outcome in code that uses the dominance frontiers.

PR #1437, opened 2020-08-28
cleanup and modernize classes in soot.jimple.toolkits.typing package

tim-hoffman:


tim-hoffman:
will update and resubmit

PR #1448, opened 2020-09-02
Cleanup and optimize LocalGenerator

tim-hoffman:


src/main/java/soot/javaToJimple/LocalGenerator.java
```diff
@@ -23,203 +23,158 @@
...
+    this.locals = b.getLocals();
   }
 
   /** generates a new soot local given the type */
-  public soot.Local generateLocal(soot.Type type) {
+  public Local generateLocal(Type type) {
 
     // store local names for enhanced performance
-    initLocalNames();
+    HashSet<String> localNames = new HashSet<String>();
```
StevenArzt, line 71: Why did you move the local name cache inside the method, instead of keeping it as a class member? When generating many locals, this might impact performance.

Waldo1637: The previous version actually had the line ```localNames = null;``` at the end of the ```generateLocal(Type)``` method so the set was rebuilt at for each call to the method anyway. Thus, logically equivalent to my updated code. I simply removed the field since it was essentially useless.

I would guess it was rebuilt for each call to the method because the set can easily become out of sync with the Chain of locals. I do have an approach that can address this issue and achieve the performance from caching that you are wanting: go back to storing the Set in a field and add an additional field to store the expected modification count of the Chain; instead of setting the Set to ```null``` at the end of each call to the method the start of the method should compare the actual modification count of the chain to the expected modification count and only rebuild if necessary. I can update the PR to use this improved approach.

StevenArzt: Your approach sounds good, that should improve performance quite a bit. I agree, the previous version did not have a cache, and that was actually bad right from the start. Could you update your PR?

Waldo1637: I will make the changes mentioned and update the PR.

PR #1453, opened 2020-09-02
Make ArrayBoundsChecker a whole-program analysis to fix a concurrency bug

### LINKED ISSUE #1354: Exception when using ArrayBoundsChecker
Hi,

When enabling the ArrayBounds checker optimization it sometimes throws a NullPointerException. It happens randomly, sometimes the program runs fine.

`[Thread-6] ERROR heros.solver.CountingThreadPoolExecutor - Worker thread execution failed: null
java.lang.NullPointerException
	at soot.jimple.toolkits.annotation.arraycheck.ArrayBoundsCheckerAnalysis.doAnalysis(ArrayBoundsCheckerAnalysis.java:372)
	at soot.jimple.toolkits.annotation.arraycheck.ArrayBoundsCheckerAnalysis.<init>(ArrayBoundsCheckerAnalysis.java:170)
	at soot.jimple.toolkits.annotation.arraycheck.ArrayBoundsChecker.internalTransform(ArrayBoundsChecker.java:104)
	at soot.BodyTransformer.transform(BodyTransformer.java:51)
	at soot.Transform.apply(Transform.java:120)
	at soot.BodyPack.internalApply(BodyPack.java:50)
	at soot.Pack.apply(Pack.java:125)
	at soot.PackManager.runBodyPacks(PackManager.java:1010)
	at soot.PackManager.lambda$runBodyPacks$0(PackManager.java:660)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Exception in thread "main" java.lang.NullPointerException
	at soot.jimple.toolkits.annotation.arraycheck.ArrayBoundsCheckerAnalysis.doAnalysis(ArrayBoundsCheckerAnalysis.java:372)
	at soot.jimple.toolkits.annotation.arraycheck.ArrayBoundsCheckerAnalysis.<init>(ArrayBoundsCheckerAnalysis.java:170)
	at soot.jimple.toolkits.annotation.arraycheck.ArrayBoundsChecker.internalTransform(ArrayBoundsChecker.java:104)
	at soot.BodyTransformer.transform(BodyTransformer.java:51)
	at soot.Transform.apply(Transform.java:120)
	at soot.BodyPack.internalApply(BodyPack.java:50)
	at soot.Pack.apply(Pack.java:125)
	at soot.PackManager.runBodyPacks(PackManager.java:1010)
	at soot.PackManager.lambda$runBodyPacks$0(PackManager.java:660)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Exception in thread "Thread-6" java.lang.NullPointerException
	at soot.jimple.toolkits.annotation.arraycheck.ArrayBoundsCheckerAnalysis.doAnalysis(ArrayBoundsCheckerAnalysis.java:372)
	at soot.jimple.toolkits.annotation.arraycheck.ArrayBoundsCheckerAnalysis.<init>(ArrayBoundsCheckerAnalysis.java:170)
	at soot.jimple.toolkits.annotation.arraycheck.ArrayBoundsChecker.internalTransform(ArrayBoundsChecker.java:104)
	at soot.BodyTransformer.transform(BodyTransformer.java:51)
	at soot.Transform.apply(Transform.java:120)
	at soot.BodyPack.internalApply(BodyPack.java:50)
	at soot.Pack.apply(Pack.java:125)
	at soot.PackManager.runBodyPacks(PackManager.java:1010)
	at soot.PackManager.lambda$runBodyPacks$0(PackManager.java:660)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)`

Flags used:
`        PhaseOptions.v().setPhaseOption("jap.npc", "enabled:true")
        PhaseOptions.v().setPhaseOption("jap.abc", "enabled:true")
        PhaseOptions.v().setPhaseOption("jap.abc", "with-cse:true")
        PhaseOptions.v().setPhaseOption("jap.abc", "with-classfield:true")
`
When disabling with-classfield and with-cse it still happens.

The bug started to occur after I added this custom transformation:
```
        PackManager.v().getPack("wjtp").add(Transform("wjtp.setup", object: SceneTransformer() {
            override fun internalTransform(p0: String, p1: MutableMap<String, String>) {
                Scene.v().reachableMethods.listener().asSequence().map {it.method().declaringClass}.distinct().forEach {
                    it.setApplicationClass()
                }
            }

        }))
```

The reason I added that optimization is, I'm using Soot with a custom (java) runtime to generate C code. Without setting all classes (which I generate into source code, including the runtime) to Application Classes, optimizations (BodyTransformer's) only get applied to the Main class and not the runtime classes, which I also need to be optimized. I can only know using the callgraph which classes end up being part of the application ((in)directly used by the Main class). I added this transformation to do that in the phase when the callgraph is available. Did I do something wrong? Is there a better way to do it? Before I would just my (custom) optimizations after all the packs were finished and modify the bodies, but I suspect the jop/wjop packs don't optimize the runtime classes since they're not application classes.

Thanks!

I'm also using Soot version 4.1.0.

tiziw:
This fixes #1354. It's caused by the fact that multiple running ArrayBoundsChecker transformations can run certain analyses (SimpleLocalDefs) on the same method, resulting into data being overwritten (in this case, the local.getNumber field). I suspect this happened, because the transformation was designed before (2000) Soot parallelized BodyTransformation's or even supported SceneTransformation's.

I'm not sure if I modified all the documentation correctly. I added a -w flag to the annot-arraybounds, since now its a whole-program transformation. As for some ui code (PhaseOptionsDialog) I didn't test it.

I've ran some basic tests and it works now in whole-program mode. Hope this is of any use.

### DIFF
--- a/eclipse/ca.mcgill.sable.soot/src/ca/mcgill/sable/soot/testing/TestOptionsDialogHandler.java
+++ b/eclipse/ca.mcgill.sable.soot/src/ca/mcgill/sable/soot/testing/TestOptionsDialogHandler.java
@@ -1483,7 +1483,103 @@ public class TestOptionsDialogHandler {
 			cmd.append(COLON);
 			cmd.append(path);
 			cmd.append(SPACE);
-		}	
+		}
+
+		key = "p"+" "+"wjap.abc"+" "+"disabled";
+		value = settings.getBoolean(key.trim());
+
+		boolDefault = true;
+
+
+		if (value != boolDefault ) {
+			cmd.append(DASH);
+			cmd.append(key.trim());
+			cmd.append(SPACE);
+		}
+
+		key = "p"+" "+"wjap.abc"+" "+"with-all";
+		value = settings.getBoolean(key.trim());
+
+		boolDefault = false;
+
+
+		if (value != boolDefault ) {
+			cmd.append(DASH);
+			cmd.append(key.trim());
+			cmd.append(SPACE);
+		}
+
+		key = "p"+" "+"wjap.abc"+" "+"with-fieldref";
+		value = settings.getBoolean(key.trim());
+
+		boolDefault = false;
+
+
+		if (value != boolDefault ) {
+			cmd.append(DASH);
+			cmd.append(key.trim());
+			cmd.append(SPACE);
+		}
+
+		key = "p"+" "+"wjap.abc"+" "+"with-arrayref";
+		value = settings.getBoolean(key.trim());
+
+		boolDefault = false;
+
+
+		if (value != boolDefault ) {
+			cmd.append(DASH);
+			cmd.append(key.trim());
+			cmd.append(SPACE);
+		}
+
+		key = "p"+" "+"wjap.abc"+" "+"with-cse";
+		value = settings.getBoolean(key.trim());
+
+		boolDefault = false;
+
+
+		if (value != boolDefault ) {
+			cmd.append(DASH);
+			cmd.append(key.trim());
+			cmd.append(SPACE);
+		}
+
+		key = "p"+" "+"wjap.abc"+" "+"with-classfield";
+		value = settings.getBoolean(key.trim());
+
+		boolDefault = false;
+
+
+		if (value != boolDefault ) {
+			cmd.append(DASH);
+			cmd.append(key.trim());
+			cmd.append(SPACE);
+		}
+
+		key = "p"+" "+"wjap.abc"+" "+"with-rectarray";
+		value = settings.getBoolean(key.trim());
+
+		boolDefault = false;
+
+
+		if (value != boolDefault ) {
+			cmd.append(DASH);
+			cmd.append(key.trim());
+			cmd.append(SPACE);
+		}
+
+		key = "p"+" "+"wjap.abc"+" "+"profiling";
+		value = settings.getBoolean(key.trim());
+
+		boolDefault = false;
+
+
+		if (value != boolDefault ) {
+			cmd.append(DASH);
+			cmd.append(key.trim());
+			cmd.append(SPACE);
+		}
 		
 		key = "p"+" "+"wjtp2"+" "+"disabled";
 		value = settings.getBoolean(key.trim());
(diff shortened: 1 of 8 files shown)


tiziw:
On second note, I made ArrayBoundsCheck run for all reachable methods, as a whole-program transformation. Maybe a config option should be added incase someone wants to run the transformation only on application classes?

MarcMil:
This is not a good idea, it is a workaround, not a fix. I acknowledge that running SimpleLocalDefs running on the same method concurrently may lead to problems, but it would be interesting to see _why_ that happens in the first place. Normally, Soot loads in the methods in parallel, but with one method assigned to one thread. It should not happen that these kind of analyses are requested to run concurrently **on the same method**.

We could somewhat solve this by synchronizing on the body (@StevenArzt: Your opinion?). But still, when there are multiple threads performing modifying changes on the same body, this will lead into other problems.

StevenArzt:
> This is not a good idea, it is a workaround, not a fix. I acknowledge that running SimpleLocalDefs running on the same method concurrently may lead to problems, but it would be interesting to see _why_ that happens in the first place. Normally, Soot loads in the methods in parallel, but with one method assigned to one thread. It should not happen that these kind of analyses are requested to run concurrently **on the same method**.
> 
> We could somewhat solve this by synchronizing on the body (@StevenArzt: Your opinion?). But still, when there are multiple threads performing modifying changes on the same body, this will lead into other problems.

This is a fundamental issue because bodies are shared between all analyses. If transformations are performed concurrently, the output is undefined. This is not a problem for read-only analyses.

Inside the Soot framework, we should normally load each body only once, i.e., if there is already a thread that loads a body, and some other thread requests the same body, this second thread should be blocked until the first thread has completed loading the body. This approach should take care of all transformations that need to happen during initial loading of a body.

From Soot's perspective, everything that happens afterward is under the responsibility of the user. If you want to have multiple threads, you need to make sure that they don't modify the body or that they are properly synchronized. There is no guarantee that a built-in transformer can safely be used inside your own code when your client analysis is multi-threaded.

MarcMil:
After further inspection I've found out that apparently that ```ArrayBoundsChecker``` uses ```ClassFieldAnalysis```, which tries to figure out values of fields and thus needs access to all initializer methods. This is obviously a problem, and as you already stated in the beginning, setting the particular analysis  to whole program mode solves the problem in this case.
To be fair, I've haven't seen this code before, and actually, my life was happier until I've seen this mess.
I'll tidy some of it up and set it to whole program mode without merging this, because we have some merge conflicts here unfortunately. Thanks for the contribution!

PS: Given the code of the analysis, I wouldn't bet on the results being completely trustworthy, so use this analyzer with great caution. There's a reason no-one else has  triggered or reported this bug yet: It has been ill tested and thus might not deliver the results you want.

StevenArzt:
I still have a bunch of concerns with making this a whole-program analysis and using it during body resolving. We load the bodies in parallel and a whole program analysis can, by definition, only be executed when all bodies (at least the ones on the process-dir) have been loaded. You would trigger a whole program analysis, and then continue loading bodies, which can potentially invalidate the results of the whole program analysis as the latter can rightfully expect the relevant bodies to all be there (or be phantom).

MarcMil:
My [PR](https://github.com/soot-oss/soot/pull/2257) made the array bounds checker a whole program analysis. I've made sure that the analysis doesn't load bodies by itself. This can be closed @StevenArzt.

PR #1456, opened 2020-09-03
Fix bugs, cleanup, and modernize Scene and ModuleScene

tim-hoffman:
1. Move android code to an inner class to reduce runtime jar dependencies when running with non-Android inputs
2. Fix bug: methods getting class name from signature must unescape the class name in case it contains reserved words because the Scene will contain the original/unescaped version of the name
3. optimize signature parsing methods to avoid repeat work, etc.
4. organize fields more clearly
5. make dynamicClasses and doneResolving fields protected so ModuleScene doesn't need its own version or to use the methods to get/set (also allowed removing a few override methods from ModuleScene)
6. use logger instead of G.v().out in ModuleScene
7. Additional cleanup, optimization steps throughout the 2 files


Waldo1637:
Further comment on point 1:
Move the code using pxb.android.axml.* into the AndroidVersionInfo class so that the AXML jar is not necessary when running Soot on non-Android inputs (i.e. the AndroidVersionInfo class will not be loaded thus the references in that package will not be seen).

Waldo1637:
Failing due to a dependency error (unrelated to code changes):
```Failed to read artifact descriptor for de.upb.cs.swt:axml:jar:2.1.0-SNAPSHOT: Could not transfer artifact de.upb.cs.swt:axml:pom:2.1.0-20200813.103355-3 from/to sonatype-snapshot (https://oss.sonatype.org/content/repositories/snapshots): Transfer failed for https://oss.sonatype.org/content/repositories/snapshots/de/upb/cs/swt/axml/2.1.0-SNAPSHOT/axml-2.1.0-20200813.103355-3.pom 502 Bad Gateway```

tim-hoffman:
will update and resubmit

PR #1473, opened 2020-10-19
Cleanup/modernize Constant and subclasses

Waldo1637:


src/main/java/soot/jimple/ClassConstant.java
```diff
@@ -92,66 +92,78 @@ private static String sootTypeToString(Type tp) {
...
+    String tmp = this.value;
+    return tmp.charAt(0) == 'L' && tmp.charAt(tmp.length() - 1) == ';';
   }
 
   public Type toSootType() {
     int numDimensions = 0;
-    String tmp = value;
-    while (tmp.startsWith("[")) {
+    String tmp = this.value;
+    while (tmp.charAt(0) == '[') {
```
StevenArzt, line 102: not equivalent to the original code if the string is empty

src/main/java/soot/jimple/ClassConstant.java
```diff
@@ -92,66 +92,78 @@ private static String sootTypeToString(Type tp) {
...
-   * Example: "java/lang/Object".
-   * See https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.2.1
+   * Gets an internal representation of the class used in Java bytecode. The returned string is similar to the fully
+   * qualified name but with '/' instead of '.'. Example: "java/lang/Object". See
+   * https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.2.1
    */
   public String toInternalString() {
     String internal = value;
-    while (internal.startsWith("[")) {
+    while (internal.charAt(0) == '[') {
```
StevenArzt, line 157: same problem as above with empty strings

PR #1481, opened 2020-10-27
 ignore two special cases when constructing HashMutablePDG;

Marvinmw:
ignore two special cases when constructing HashMutablePDG; 1 Block may point to itself.2. PDGNode may contain Block node not Region node

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,3 @@
+# These are supported funding model platforms
+
+github: soot-oss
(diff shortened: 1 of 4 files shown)


StevenArzt:
That looks good content-wise, but please modify the MR to point to our ``develop`` branch instead of ``master``.

Marvinmw:
> That looks good content-wise, but please modify the MR to point to our `develop` branch instead of `master`.

hi, I changed it.

Issue #1064, opened 2018-11-14
Add extension point to OnFlyCallGraphBuilder for handling special cases


mbenz89:
When generating a call graph, SPARK has to handle many special cases for concurrency helper classes like `java.lang.Thread` or `android.os.AsyncTask`.  These classes usually have control flow properties _hidden_ to the call graph analysis, e.g., the `Thread`'s `Thread.run` method will be invoked asynchronously after calling `Thread.start` or `AsyncTask.onPostExecute` will be executed after `AsyncTask.doInBackground` has finished. SPARK handles these cases by building an artifical short circuit between these framework managed calls, i.e., it adds edges between the start, computation and epilogue methods of these concurrecy helper classes to the call graph.

Currently, these special cases are hard coded into the [`OnFlyCallGraphBuilder`](https://github.com/Sable/soot/blob/develop/src/main/java/soot/jimple/toolkits/callgraph/OnFlyCallGraphBuilder.java) class.

`OnFlyCallGraphBuilder` needs an extension point which is capable of handling such special cases. Handling of concurrency helper classes and other special cases should then be removed from `OnFlyCallGraphBuilder` and encapsulated in implementations of the extension point.

StevenArzt:
Have you looked at ``virtualedges.xml``, which is our declarative specification for such callbacks?

We can generate / extend this file using dynamic analysis. We had a paper on ECOOP on that: https://drops.dagstuhl.de/entities/document/10.4230/LIPIcs.ECOOP.2024.4 There's nothing wrong with manually adding entries, though.

Danish-git123:
Thanks for pointing out this approach

PR #1485, opened 2020-10-28
Improvements to OnFlyCallgraphBuilder

JuliusNmn:
Addressed issue #1064, OnFlyCallgraphBuilder now loads library descriptions from a configuration file in `src/main/resources/virtualedges.xml`. Added missing java stdlib functions to such as java.io.Files.listFiles(). Additionally added callback summaries for several libraries. Additionally, allow specifying "Indirect Callbacks" to describe behavior such as the following:
```
Request r = new Request(" http ://www. example .com");
ResponseHandler handler = new ResponseHandler(){
    @Override
    void onResponse(String[] data, int httpStatus)
    ...
    }
};
r.setResponseHandler(handler);
requestQueue.add(r);
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -570,15 +570,7 @@
             <groupId>org.glassfish.jaxb</groupId>
             <artifactId>jaxb-runtime</artifactId>
             <version>2.4.0-b180830.0438</version>
-        </dependency>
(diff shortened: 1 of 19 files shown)


StevenArzt:
@JuliusNmn The tests on Java 11 are failing. Could you please double-check? I guess that's a problem with all JDKs newer than 9, as Java 9 introduced quite a few breaking changes.

PR #1490, opened 2020-11-12
Detect the object-level redundancy issue

anonymousemberxxx:
Hello,

We ran soot with our tool and detected the object-level redundancy issues (object-level redundancies that happen across objects sharing the same calling context). Our tool reports two objects that suffer from object-level redundancy, which are allocated at line 339 in method orAndAndNot and accessed at line 359 and line 360 in method orAndAndNot of class BitVector. The optimized code is in this pull request.

### DIFF
--- a/src/main/java/soot/util/BitVector.java
+++ b/src/main/java/soot/util/BitVector.java
@@ -29,6 +29,8 @@ package soot.util;
  *
  * @author Ondrej Lhotak
  */
+import java.util.*;
+
 public class BitVector {
   public BitVector() {
     this(64);
@@ -310,6 +312,9 @@ public class BitVector {
    * @param set
    *          a bit set.
    */
+  public static long[] tmp_b;
+  public static long[] tmp_e;
+  public static boolean tmp_ret;
   public boolean orAndAndNot(BitVector orset, BitVector andset, BitVector andnotset) {
     boolean ret = false;
     long[] a = null, b = null, c = null, d = null, e = null;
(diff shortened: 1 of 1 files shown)


src/main/java/soot/util/BitVector.java
```diff
@@ -310,6 +312,9 @@ public String toString() {
    * @param set
    *          a bit set.
    */
+  public static long[] tmp_b;
```
StevenArzt, line 315: Why is is this static? You can have more than instance of BitVector.

bolunli11: Yes, the program can has more than instance of BitVector, but after profiling with our tool, we found that sometimes the computation process generates some same variable (e.g., b, e, ret). So if the BitVector contents are the same with the BitVector from last iteration, we can avoid some computations here.

MarcMil: I am pretty sure this would break multi threaded applications in a spectacular way. And given that analyses are often run multi threaded I'm afraid this is not a good idea.

StevenArzt: I agree with @MarcMil. Your proposed change would make it impossible to have two threads that each use their own BitVector, as these BitVectors may suddenly interfere with each other. Further, I haven't yet seen any convincing data that this redundance is actually a problem in practice. I would be quite surprised if this were a major cause of Soot's memory consumption. Even if so, we would need a more complex solution that can deal with multi-threading.

In total, I think that not a lot is to gain here, but quite a lot to break. I'll therefore close this PR.

PR #1504, opened 2020-12-16
SourceLocators cleanup, modernization, and cache tuning

tim-hoffman:


src/main/java/soot/ModulePathSourceLocator.java
```diff
@@ -656,66 +621,62 @@ protected void setupClassProviders() {
...
+        // Note that some FileSystem implementations used by Path use even on Windows
+        // "/" as path separator. Therefore we have to use the file-system specific path
+        //  separator instead of the system specific one (File.separatorChar).
+        String fileName = aPath.relativize(file).toString().replace(file.getFileSystem().getSeparator(), ".");
+        if (fileName.endsWith(".class")) {
+          classes.add(fileName.substring(0, fileName.lastIndexOf(".class")));
+        } else if (fileName.endsWith(".jimple")) {
+          classes.add(fileName.substring(0, fileName.lastIndexOf(".jimple")));
+        } else if (fileName.endsWith(".java")) {
+          classes.add(fileName.substring(0, fileName.lastIndexOf(".java")));
```
swissiety, line 648: if the condition is true we can calculate the end position of the substring to avoid additional search costs of O( suffix.length() ) so we can replace `fileName.lastIndexOf( suffix )` by filename.length()-suffix.length()
```suggestion
        if (fileName.endsWith(".class")) {
          classes.add(fileName.substring(0, fileName.length()-6));
        } else if (fileName.endsWith(".jimple")) {
          classes.add(fileName.substring(0, fileName.length()-7));
        } else if (fileName.endsWith(".java")) {
          classes.add(fileName.substring(0, fileName.length()-5));
```

patricklam: That makes the code harder to understand, do we know this to be a bottleneck?

swissiety: I'll agree.. suggestion updated.

michael-emmi:
Are there any blockers on merging this PR?I’m seeing performance regressions with JDK11 runtime

PR #1506, opened 2020-12-16
OnFlyCallGraphBuilder bugfix and cleanup

Waldo1637:
When the flag was not static, rebuilding the callgraph would cause the guards pack to be added again which causes an exception.


mbenz89:
Thanks for all those great cleanups, modernizations, and bug fixes!

PR #1519, opened 2021-01-05
Update FieldRef classes

Waldo1637:
- formatting and cleanup
- caching of the resolved SootField in AbstractSootFieldRef because it can be an expensive operation

### DIFF
--- a/src/main/java/soot/AbstractSootFieldRef.java
+++ b/src/main/java/soot/AbstractSootFieldRef.java
@@ -34,7 +34,6 @@ import soot.options.Options;
  * actually exist; the actual target of the reference is determined according to the resolution procedure in the Java Virtual
  * Machine Specification, 2nd ed, section 5.4.3.2.
  */
-
 public class AbstractSootFieldRef implements SootFieldRef {
   private static final Logger logger = LoggerFactory.getLogger(AbstractSootFieldRef.class);
 
@@ -42,6 +41,8 @@ public class AbstractSootFieldRef implements SootFieldRef {
   private final String name;
   private final Type type;
   private final boolean isStatic;
+
+  private SootField resolveCache = null;
 
   public AbstractSootFieldRef(SootClass declaringClass, String name, Type type, boolean isStatic) {
     if (declaringClass == null) {
@@ -102,7 +103,12 @@ public class AbstractSootFieldRef implements SootFieldRef {
 
   @Override
   public SootField resolve() {
-    return resolve(null);
+    SootField cached = this.resolveCache;
+    if (cached == null) { // Use the cached SootField if available
+      cached = resolve(null);
+      this.resolveCache = cached;
+    }
+    return cached;
   }
 
   private SootField checkStatic(SootField ret) {
--- a/src/main/java/soot/AbstractSootFieldRef.java
+++ b/src/main/java/soot/AbstractSootFieldRef.java
@@ -38,11 +38,12 @@ import soot.options.Options;
 public class AbstractSootFieldRef implements SootFieldRef {
   private static final Logger logger = LoggerFactory.getLogger(AbstractSootFieldRef.class);
 
+  private final SootClass declaringClass;
+  private final String name;
+  private final Type type;
+  private final boolean isStatic;
+
   public AbstractSootFieldRef(SootClass declaringClass, String name, Type type, boolean isStatic) {
-    this.declaringClass = declaringClass;
-    this.name = name;
-    this.type = type;
-    this.isStatic = isStatic;
     if (declaringClass == null) {
       throw new RuntimeException("Attempt to create SootFieldRef with null class");
     }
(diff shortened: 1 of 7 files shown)


linghuiluo:
@swissiety can you review this ?

src/main/java/soot/AbstractSootFieldRef.java
```diff
@@ -34,15 +34,17 @@
...
-
 public class AbstractSootFieldRef implements SootFieldRef {
   private static final Logger logger = LoggerFactory.getLogger(AbstractSootFieldRef.class);
 
+  private final SootClass declaringClass;
+  private final String name;
+  private final Type type;
+  private final boolean isStatic;
+
+  private SootField resolveCache = null;
```
swissiety, line 45: as there can exist multiple instances of *FieldRef refering to the same SootField - maybe we can use a global cache for it?

tim-hoffman: That's true. I did see a performance improvement with this local caching and had not even considered global caching. I guess the same would apply to the SootMethodRefImpl caching in PR #1520.

linghuiluo:
@swissiety  Is this PR ready for merging?

swissiety:
@linghuiluo that depends on whether we want a SootField caching mechanism that works globally instead of locally per FieldRef as it is introduced in this PR

tim-hoffman:
Regardless, local caching does provide an improvement. Changing to global caching could be considered separately in the future.

src/main/java/soot/AbstractSootFieldRef.java
```diff
@@ -97,42 +95,51 @@ public FieldResolutionFailedException() {
...
+    SootField cached = this.resolveCache;
+    // Use the cached SootField if available and still valid
+    if (cached == null || !isValidResolve(cached)) {
+      cached = resolve(null);
+      this.resolveCache = cached;
+    }
+    return cached;
+  }
+
+  private boolean isValidResolve(SootField f) {
```
linghuiluo, line 115: Is there a case this function would return false for a cached SootField?

tim-hoffman: The ```isValidResolve(SootField)``` function can return false if the ```SootField``` instance was modified after it was cached here. The ```SootField``` class has methods such as ```setName(..)```, ```setType(..)```, etc. that allow it to be modified.

linghuiluo: @tim-hoffman this is probably bad. If someone doesn't take care and add more setters to SootField in future without modifying this function. IMHO, it is better to compare hashcode of objects or use an oberser pattern.

tim-hoffman: It's important to note that adding more setters to ```SootField``` in the future could only cause an issue if they overlap with concepts/values which have getters in the ```SootFieldRef``` interface. Witht that said, here are my thoughts on those two options as well as a third which I think provides the best solution.

Option 1 - Observer: ```SootField``` would need to keep a Set of ```AbstractSootFieldRef``` that have previously been resolved to that ```SootField``` and any time the ```SootField``` is modified (in a way that overlaps with the ```SootFieldRef``` interface), the Set must be cleared. ```AbstractSootFieldRef#resolve()``` would need to register itself with that Set in the respective resolved ```SootField```. Then the ```isValidResolve``` check would be replaced with checking if the cached ```SootField``` contains the current ```AbstractSootFieldRef``` instance in its Set. I don't think this is a great solution because it has additional storage overhead for the Set and if someone adds another setter to ```SootField```, it still has the potential of getting out of sync. There is also the potential of someone adding a setter to ```SootField``` and clearing the Set unnecessarily if that new setter does not overlap with the ```SootFieldRef``` interface (that person would have to remember that not all modifications to a ```SootField``` object make a ```SootFieldRef``` for it invalid).
Option 2 - hashCode: This would require adding an additional field alongside the cached ```SootField``` reference to store the hashcode of that reference. However, ```SootField``` does not implement ```#hashCode()``` (which may be intentional so I'd rather not change that). It does have ```#equivHashCode()``` but that does not check the declaring class field (and again, probably intentional so I'd rather not change it). Thus, we either need an ad-hoc solution that combines ```#equivHashCode()``` with the declaring class reference's hashcode or have to introduce another form of hashcode method in ```SootField```. Neither seems great, and again, there is additional memory overhead for the 'int' field to store alongside the cached ```SootField``` reference in ```AbstractSootFieldRef```.
Option 3 - Move ```#isValidResolve(..)``` to the ```SootField``` class and change its signature to ```isValidResolve(SootFieldRef)```. I think this is the best solution because it adds no additional memory overhead for the field/fieldref. I think placing that method in the ```SootField``` class can help with your concern of someone updating the ```SootField``` class by adding something else to it while still making it clear that only additions/changes that also overlap with the ```SootFieldRef``` interface need to used in the comparison.

linghuiluo: @tim-hoffman  Thanks for elaborating the options, option 3 looks good.

tim-hoffman: Okay, I will make that change and update.

StevenArzt:
Has been integrated separately.

PR #1521, opened 2021-01-05
Cleanup and format subclasses of soot.Type

tim-hoffman:


tim-hoffman:
will update and resubmit

PR #1523, opened 2021-01-05
cleanup and format other "Printer" classes

tim-hoffman:


tim-hoffman:
will update and resubmit

PR #1525, opened 2021-01-06
cleanup and format BodyPack and related classes

tim-hoffman:


tim-hoffman:
will update and resubmit

PR #1535, opened 2021-01-15
Log error and proceed in OnFlyCallgraph

MarcMil:


StevenArzt:
We have an open issue with the type assigner, which fails for certain casts between ``java.lang.Object`` and array types. Unfortunately, corner cases with the type assigner come up every now and then.

We need to fix the root cause in each case, but in large evaluations or productive environments, it's better to carry on with a potentially incomplete callgraph than fail altogether and never get to an evaluation because you're stuck in fixing stuff. We need to make this lenient behavior an option, though.

Issue #1562, opened 2021-02-24
Soot fails due to not found super class 


piskachev:
While Jimple is constructed there is a case where Soot fails at [this point](https://github.com/soot-oss/soot/blob/59931576784b910a7d38f81910b7313aa2feafea/src/main/java/soot/SootClass.java#L892).

The example trace is below. What is happening is that some classes get generated and we are missing them on the classpath so Soot does not find them. Even though the phantom option is enabled some of them as you see are not phantom and this exception occurs. 

How can we handle such cases? Is it possible to relax the condition when super class is not found? 
It is better to have incomplete hierarchy as long as Soot does not crashes.  

------------------------------------------------------
Caused by: java.lang.RuntimeException: Failed to apply jb to <some.Mod: java.util.List getAuthorizers(java.lang.String,java.lang.String)>
    at soot.asm.AsmMethodSource.getBody(AsmMethodSource.java:2191) 
    at soot.SootMethod.retrieveActiveBody(SootMethod.java:402) 
    at soot.jimple.toolkits.callgraph.OnFlyCallGraphBuilder.processNewMethod(OnFlyCallGraphBuilder.java:812) 
    at soot.jimple.toolkits.callgraph.OnFlyCallGraphBuilder.processReachables(OnFlyCallGraphBuilder.java:286) 
    at soot.jimple.toolkits.callgraph.CallGraphBuilder.build(CallGraphBuilder.java:111) 
    at soot.jimple.toolkits.callgraph.CHATransformer.internalTransform(CHATransformer.java:51) 
    at soot.SceneTransformer.transform(SceneTransformer.java:36) 
    at soot.Transform.apply(Transform.java:102) 
    at soot.RadioScenePack.internalApply(RadioScenePack.java:68) 
    at soot.jimple.toolkits.callgraph.CallGraphPack.internalApply(CallGraphPack.java:58) 
    at soot.Pack.apply(Pack.java:117) 
    at utils.SootRunner.run(SootRunner.java:137) 
    at some.Runner.runTransforms(Runner.java:111) 
    ... 6 more
Caused by: java.lang.RuntimeException: no superclass for some.Interface
    at soot.SootClass.getSuperclass(SootClass.java:897) 
    at soot.jimple.toolkits.typing.fast.BytecodeHierarchy.buildAncestryTree(BytecodeHierarchy.java:76) 
[...]

PR #1564, opened 2021-03-02
Fix for #1562

piskachev:
If the option ignore-resolution-errors is enabled Soot should not fail when super class does not exist. This can happen when some library generates classes that land in the Scene as non phantom. Hence, it can happen that some of them have incomplete hierarchy (Soot throws exception when super class is not found). 
This is fix for ticket #1562

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,13 @@ We are regularly applying for funding to help us maintain Soot. You can help us 
 
 Also many thanks to [![JProfiler](https://www.ej-technologies.com/images/product_banners/jprofiler_small.png)](https://www.ej-technologies.com/products/jprofiler/overview.html) for supporting Soot with a free-to-use open source license!
 
+# Thanks to our Sponsors...
+... for supporting the further Development of Soot!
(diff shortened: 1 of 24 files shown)


mbenz89:
@piskachev look fine to me! Have you tested if allowing `null` as superclass does not lead to errors elsewhere?

piskachev:
> @piskachev look fine to me! Have you tested if allowing `null` as superclass does not lead to errors elsewhere?

yes, tested! To make sure it does not fails, should I add the same condition (&& !Options.v().ignore_resolution_errors()) also in BytecodeHierarchy:75 ?

mbenz89:
>yes, tested! To make sure it does not fails, should I add the same condition (&& !Options.v().ignore_resolution_errors()) also in BytecodeHierarchy:75 ?

Yes. That would be good. Thereby, we at least do not break any assumption when having `!Options.v().ignore_resolution_errors()`

piskachev:
I added the change.

ericbodden (commented):
Looks good to me. Thanks!

PR #1569, opened 2021-03-13
Improve type assigner and introducing a flow-sensitive constant propagator

MarcMil:


StevenArzt:
@mbenz89 @ericbodden 

This PR is the output of a large-scale test of Soot's type assigner on Android apps from a 2021 PlayStore dump. The type assigner already has a ``UseChecker`` that verifies whether a typing is inconsistent, e.g., there are casts between types that are not cast-compatible. We found quite a few of these issues, with the respective method bodies being erroneuous accordingly. At the moment, Soot just ignores such issues, and unless your client analysis is very precise (e.g., checks typecasts), the problem remains hidden.

The root cause of these typing issues is usually local splitting. When loading a Jimple body from dex, Dexpler first converts each register into a local. Since a register can first be used as type A and later as type B, there is no valid typing at this point. Therefore, Dexpler invokes the local splitter to establish different code-locals for independent use/def clusters of the same register-local. Afterwards, the local types are consistent and the type assigner can do its work. So far so good. Now let's take the following simplified code:

``
u0 = 0
if (...) goto l0:
// use u0 as boolean
goto l1;
l0:
// use u0 as int
l1:
```

This code is perfectly fine in the Dalvik VM, because the intialization value 0 is both an integer and a boolean value, and once it has been used in one of the two contexts, the code is consistent with its runtime typing. Soot's local splitter will not split anything, because the two uses depend on the same definition. Therefore, type assignment fails. Local ``u0`` cannot be an intger and  boolean at the same type, and the two types do not have a common ancestor. There is no valid typing for ``u0``,  and we cannot resolve the issue by introducing typecasts.

The solution is to split the constant initialization into two statements:

``
u0 = 0
u1 = 0
if (...) goto l0:
// use u0 as boolean
goto l1;
l0:
// use u1 as int
l1:
```

Note that we identify exactly those cases where the only shared definition is a side effect-free constant assignment that we can safely duplicte. Also note that there is never a join point where we would need ``u0`` and ``u1`` to be the same local, since such code would semantially be wrong and be rejected by the runtime type checks of the Dalvik VM.

This brings us into the problem of detecting constant initializations, and that's why we had to improve the constant propagator quite a bit. We have tested the changes on a large set of apps and are quite confident that we didn't break any obvious stuff.

The Dexpler frontend already did quite a few transformations before this change, so I don't consider it a problem that we now even do more, i.e., the constant propagator will optimize away trivial operations.

Issue #1577, opened 2021-04-03
Crash in retriving body of method


FXTi:
Soot throws RuntimeException in retriving method body:

```java
java.lang.RuntimeException: Failed to convert <cn.com.chinatelecom.account.api.c.g: int h(android.content.Context)>
	at soot.asm.AsmMethodSource.getBody(AsmMethodSource.java:2163)
	at soot.SootMethod.retrieveActiveBody(SootMethod.java:402)
	at soot.jimple.toolkits.callgraph.OnFlyCallGraphBuilder.processNewMethod(OnFlyCallGraphBuilder.java:846)
	at soot.jimple.toolkits.callgraph.OnFlyCallGraphBuilder.processReachables(OnFlyCallGraphBuilder.java:313)
	at soot.jimple.toolkits.callgraph.CallGraphBuilder.build(CallGraphBuilder.java:111)
	at soot.jimple.toolkits.callgraph.CHATransformer.internalTransform(CHATransformer.java:51)
	at soot.SceneTransformer.transform(SceneTransformer.java:36)
	at soot.Transform.apply(Transform.java:102)
	at soot.RadioScenePack.internalApply(RadioScenePack.java:68)
	at soot.jimple.toolkits.callgraph.CallGraphPack.internalApply(CallGraphPack.java:58)
	at soot.Pack.apply(Pack.java:117)
	at soot.PackManager.runWholeProgramPacks(PackManager.java:611)
	at soot.PackManager.runPacksNormally(PackManager.java:495)
	at soot.PackManager.runPacks(PackManager.java:419)
```

This is sample file: [sample](https://github.com/soot-oss/soot/files/6252213/sample.zip)

PR #1579, opened 2021-04-05
Fix crashes in AsmMethodSource(#1577,#1580)

FXTi:
Fix: https://github.com/soot-oss/soot/issues/1577

* `mergeIn()` should resolve the right `stack_n := @caughtException` when `nop` for `InlineExceptionHandler` appears.
* `UnitContainer` may appear instead of `nop` for `InlineExceptionHandler`.

Fix: https://github.com/soot-oss/soot/issues/1580

* Replace `Edge.stack` with `Edge.stacks` and keep `stack`s in the same order as `edge`s in `worklist`.

### DIFF
--- a/src/main/java/soot/asm/AsmMethodSource.java
+++ b/src/main/java/soot/asm/AsmMethodSource.java
@@ -1902,6 +1902,16 @@ final class AsmMethodSource implements MethodSource {
     edges = null;
   }
 
+  Unit getUnitMaybeInlineExceptionHandler(AbstractInsnNode insn) {
+    Unit ret = getUnit(insn);
+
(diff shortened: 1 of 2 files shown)


linghuiluo:
This fix doesn't look so trivial. Can you provide unit tests?

FXTi:
No, but I recommend add some unit tests for this pr.

linghuiluo:
@FXTi Will you add the tests? If you add the tests, I can review this PR.

linghuiluo:
@swissiety Can you add the two test cases into soot and see if the bugs are still there? If it is the case, make a pull request with the tests and mark them as ignored at first.

PR #1588, opened 2021-04-15
Add a type validator for BafBody instances

tim-hoffman:
NOTE: OpStackCalculator is based on soot.jbco.bafTransformations.StackTypeHeightCalculator but with some modifications and decoupling from the jbco package.


src/main/java/soot/baf/AbstractInstSwitch.java
```diff
@@ -0,0 +1,392 @@
...
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ * #L%
+ */
+
+/**
+ * @author Timothy Hoffman
+ */
+public abstract class AbstractInstSwitch<T> implements InstSwitch {
+
```
linghuiluo, line 29: This class is not used anywhere in the PR.

src/main/java/soot/baf/toolkits/base/OpStackCalculator.java
```diff
@@ -0,0 +1,863 @@
...
+import soot.baf.XorInst;
+import soot.baf.internal.AbstractOpTypeInst;
+import soot.toolkits.graph.BriefUnitGraph;
+import soot.util.Chain;
+
+/**
+ * @author Michael Batchelder
+ * @author Timothy Hoffman
+ */
+public class OpStackCalculator {
```
linghuiluo, line 135: can you add some javadoc explaining what this class oes

src/main/java/soot/baf/toolkits/base/OpStackCalculator.java
```diff
@@ -0,0 +1,863 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(OpStackCalculator.class);
+
+  private static class StackEffectSwitch implements InstSwitch {
+
+    public boolean shouldThrow = true;
+    public Type[] remove_types = null;
+    public Type[] add_types = null;
+
```
linghuiluo, line 143: What are the purpose of using these two Type arrays.

src/main/java/soot/baf/toolkits/base/OpStackCalculator.java
```diff
@@ -0,0 +1,863 @@
...
+//      return true;
+//    }
+//    if (t1 instanceof DoubleType && t2 instanceof DoubleType) {
+//      return true;
+//    }
+//    if (t1 instanceof FloatType && t2 instanceof FloatType) {
+//      return true;
+//    }
+//    return false;
+//  }
```
linghuiluo, line 744: code comment should be removed

src/main/java/soot/baf/toolkits/base/OpStackCalculator.java
```diff
@@ -0,0 +1,863 @@
...
+      Map<Unit, Integer> indexes = new HashMap<Unit, Integer>();
+      {
+        int count = 0;
+        for (Iterator<Unit> it = units.snapshotIterator(); it.hasNext();) {
+          indexes.put(it.next(), count++);
+        }
+      }
+      for (Iterator<Unit> it = units.snapshotIterator(); it.hasNext();) {
+        Unit unit = it.next();
+        String s = "";
```
linghuiluo, line 768: might be better to use a StringBuilder here

src/main/java/soot/baf/toolkits/base/OpStackCalculator.java
```diff
@@ -0,0 +1,863 @@
...
+      return "J";
+    } else if (t instanceof RefLikeType) {
+      return "L" + t.toString();
+    } else {
+      return "U(" + t.getClass().toString() + ")";
+    }
+  }
+
+  public static void printUnits(PatchingChain<Unit> u, String msg) {
+    System.out.println("\r\r***********  " + msg);
```
linghuiluo, line 823: looks like this is just for your debugging purpose?
Are you aware of this class?
https://github.com/soot-oss/soot/blob/51acd8db18bb90b91f469c282ac49a62b9438c6b/src/main/java/soot/Printer.java

src/main/java/soot/baf/AbstractInstSwitch.java
tim-hoffman: I ended up not using it but decided to keep it in because all of the other *Switch interfaces have an Abstract implementation (which is what I based this on) and I thought it might be a useful framework class to have.

src/main/java/soot/baf/toolkits/base/OpStackCalculator.java
tim-hoffman: They are basically gen and kill sets for the types on the JVM stack that is being emulated.

src/main/java/soot/baf/toolkits/base/OpStackCalculator.java
tim-hoffman: Like I said earlier, OpStackCalculator is based on soot.jbco.bafTransformations.StackTypeHeightCalculator but with some modifications and decoupling from the jbco package so I really didn't put any effort into changing the debugging aspects.

src/main/java/soot/baf/toolkits/base/OpStackCalculator.java
tim-hoffman: Will update.

src/main/java/soot/baf/AbstractInstSwitch.java
linghuiluo: when it is not used, please remove it

src/main/java/soot/baf/toolkits/base/OpStackCalculator.java
linghuiluo: add comments next to them or to the javadoc the class

PR #1594, opened 2021-04-19
Allow more general graph type in the LiveLocals analysis

tim-hoffman:


ericbodden:
@mbenz89 could you please have a look why the license-header check does not terminate?

mbenz89:
> @mbenz89 could you please have a look why the license-header check does not terminate?

For some reason merging into master and develop was restricted by a strange license check that does not exist. We check the licenses in a self-defined Github action and don't require that check. 

TLDNR: everything should properly work now :).

Issue #1602, opened 2021-05-02
A bug in EntryPoints:Iterable<SootMethod> clinitsOf(SootClass cl).


DongjieHe:
Consider a case where the input class cl does not have the "clinit" method but its superclass has the "clinit" method. Then the correct output should give me the <clinit> in its superclass. However, the current implementation just returns an empty collection (line 189).

linghuiluo:
Are you analyzing java bytecode or dex file?
If it is the later case, it can be that some optimization was already done during the build of the APK.
A small example to reproduce this would be helpful.

DongjieHe:
Hi Linghui,
I am analyzing Java bytecode.
The attached file gives a small example that contains a total of 4 classes.
A1, A, and IA have the following relation:
**A1 extends A implements IA**
The main() method in Main presents a use case. 
When analyzing this example code, we invoke EntryPoints::clinitOf(A1) and expect it to return two clinits, i.e., "<soot.A: void clinit()>" and "<soot.IA: void clinit()>".
But the current implementation will give us an empty result due to its line 189. 

I hope I explain this issue clearly.

[Clinit-src.zip](https://github.com/soot-oss/soot/files/6438847/Clinit-src.zip)

linghuiluo:
thanks, I will take a look into this.

DongjieHe:
@linghuiluo Hi Linghui, suppose you are correct, and I need to get all static constructors that can be called by A1, if I should call EntryPoints::clinitOf on A1 and all its superclasses, then EntryPoints::clinitOf(SootClass cl) could be simplified by just returning the static constructor of cl.

linghuiluo:
@DongjieHe Returning the static constructor of cl is not equal to what it is now. This method doesn't do as the Javadoc says. You are right, it affects also the call graph. I made a pull request https://github.com/soot-oss/soot/pull/1610
Can you also review it and leave a comment?

PR #1610, opened 2021-05-13
Fix 1602

linghuiluo:
The bug in https://github.com/soot-oss/soot/issues/1602  affects the call graph, because the `EntryPoints.clinitOf` is used in the call graph construction. 
This patch fixes the missing edges to some static initializers.


DongjieHe:
@linghuiluo Hi, your fix seems correct now according to its specification (i.e., " Returns a list of all clinits of class cl and its superclasses."). But it still misses "<soot.IA: void clinit()>" which I mentioned in the example in #1602 and thus still affects the on-the-fly callgraph construction.

linghuiluo:
@DongjieHe That edge from main to AI.clinit is not there, because the AI.clinit is defined in A, not in AI. So there is no method with the signature AI.clinit in the IR. This can not be done unless a SootMethod AI.clinit is also added to the class AI, which is not the case now as the SootClasses are generated. 

The code of AI.clinit and A.clinit is the same, at least you wont miss analyzing the code now.

linghuiluo:
@DongjieHe Below is from Java spec, it basically means there is no static initializer in IA at all, because you only have a String field in IA: 

Note that interface fields that are constant variables (§4.12.4) are initialized before other interface fields. This also applies to static fields that are constant variables in classes (§8.3.2). Such fields will never be observed to have their default initial values (§4.12.5), even by devious programs.

A constant variable is a final variable of primitive type or type String that is initialized with a constant expression (§15.28). Whether a variable is a constant variable or not may have implications with respect to class initialization (§12.4.1), binary compatibility (§13.1, §13.4.9), and definite assignment (§16 (Definite Assignment)).

PR #1617, opened 2021-05-30
Fix dae

MarcMil:


linghuiluo:
Would you mind to add a test case to here testing your change? https://github.com/soot-oss/soot/blob/3966f565db6dc2882c3538ffc39e44f4c14b5bcf/src/test/java/soot/jimple/toolkits/scalar/DeadAssignmentEliminatorTest.java

MarcMil:
Done.

PR #1618, opened 2021-05-31
fix1308

linghuiluo:
Address an interesting abnormal case mentioned here https://github.com/soot-oss/soot/pull/1308
A concrete method with no instruction would cause AsmMethodSource throw null point exception.


linghuiluo:
@mbenz89 Maybe you can have a look at this PR, since you looked into #1308 last year.

linghuiluo:
@swissiety can you review this ?

src/main/java/soot/PackManager.java
```diff
@@ -958,13 +958,11 @@ private void runBodyPacks(SootClass c) {
         // whole shimple or not?
         {
           Body body = m.retrieveActiveBody();
-          if (body instanceof ShimpleBody) {
+          if (body != null && body instanceof ShimpleBody) {
```
swissiety, line 961: `!= null` and `null instanceof whatever` is redundant as `null instanceof` resolves to false

linghuiluo:
@linghuiluo  I am working on this

PR #1624, opened 2021-06-02
DexPrinter: improved handling of errors while writing a method

jpstotz:
Without this PR it was always necessary to use the debugger to identify the method which caused the problem.  
Therefore I added another RuntimeException which explicitly states which method was processed by DexPrinter when the error occurred.

+ other minor improvements

### DIFF
--- a/src/main/java/soot/toDex/DexPrinter.java
+++ b/src/main/java/soot/toDex/DexPrinter.java
@@ -1090,17 +1090,22 @@ public class DexPrinter {
         continue;
       }
 
-      MethodImplementation impl = toMethodImplementation(sm);
+      MethodImplementation impl;
+      try {
+        impl = toMethodImplementation(sm);
+      } catch (Exception e) {
+        throw new RuntimeException("Error while processing method " + sm, e);
+      }
 
       List<String> parameterNames = null;
-      if (sm.hasTag("ParamNamesTag")) {
-        parameterNames = ((ParamNamesTag) sm.getTag("ParamNamesTag")).getNames();
+      if (sm.hasTag(ParamNamesTag.NAME)) {
+        parameterNames = ((ParamNamesTag) sm.getTag(ParamNamesTag.NAME)).getNames();
       }
 
       int paramIdx = 0;
       List<MethodParameter> parameters = null;
       if (sm.getParameterCount() > 0) {
-        parameters = new ArrayList<MethodParameter>();
+        parameters = new ArrayList<>();
         for (Type tp : sm.getParameterTypes()) {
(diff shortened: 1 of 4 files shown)


src/main/java/soot/toDex/DexPrinter.java
```diff
@@ -1090,17 +1090,22 @@ private Annotation buildEnclosingMethodTag(EnclosingMethodTag t, Set<String> ski
         continue;
       }
 
-      MethodImplementation impl = toMethodImplementation(sm);
+      MethodImplementation impl;
+      try {
+        impl = toMethodImplementation(sm);
+      } catch (Exception e) {
+        throw new RuntimeException("Error while processing method " + sm, e);
```
StevenArzt, line 1097: Bitte hier eine spezifische Exception-Klasse anlegen und verwenden.

src/main/java/soot/toDex/DexPrinter.java
```diff
@@ -1148,12 +1153,16 @@ protected MethodImplementation toMethodImplementation(SootMethod m) {
...
     // check the method name to make sure that dexopt won't get into trouble
-    // when installing the app
-    if (m.getName().contains("<") || m.getName().equals(">")) {
-      if (!m.getName().equals("<init>") && !m.getName().equals("<clinit>")) {
-        throw new RuntimeException("Invalid method name: " + m.getName());
+    // when installing the app. See function IsValidMemberName
+    // https://android.googlesource.com/platform/art/+/refs/heads/master/libdexfile/dex/descriptors_names.cc#271
+    if (m.getName().contains("<") || m.getName().contains(">")) {
+      if (!m.isConstructor() && !m.isStaticInitializer()) {
+        throw new RuntimeException("Invalid method name: " + m.getSignature());
```
StevenArzt, line 1160: auch hier die spezifische Exception von oben verwenden

src/main/java/soot/toDex/DexPrinter.java
```diff
@@ -1148,12 +1153,16 @@ protected MethodImplementation toMethodImplementation(SootMethod m) {
...
       }
     }
+    if (m.getName().isEmpty()) {
+      throw new RuntimeException("Invalid empty method name: " + m.getSignature());
```
StevenArzt, line 1164: auch hier die spezifische Exception von oben verwenden

jpstotz:
Additionally the changes to `toMethodImplementation` were developed together with Marc. The original code was a bit strange and we updated it to reflect the changes done by the Andorid verifier.

StevenArzt (changes requested):

PR #1640, opened 2021-06-04
resubmit fix 1308

linghuiluo:
Address an interesting abnormal case mentioned here https://github.com/soot-oss/soot/pull/1308
A concrete method with no instruction would cause AsmMethodSource throw null point exception.


linghuiluo:
@swissiety I closed the other one and resubmit this. can you review this?

src/main/java/soot/PackManager.java
```diff
@@ -993,30 +996,30 @@ private void runBodyPacks(SootClass c) {
...
     if (tc != null) {
       processXMLForClass(c, tc);
-      // System.out.println("processed xml for class");
     }
 
     if (produceDava) {
       for (SootMethod m : c.getMethods()) {
-        if (!m.isConcrete()) {
+        if (!m.isConcrete() || !m.hasActiveBody()) {
+          //note: abnormal class can have a concrete method without body.
```
swissiety, line 1022: should we warn/inform the user about this abnormal case?

swissiety (approved):
:+1:

Issue #1641, opened 2021-06-04
jb.sils interferes with jb:use-original-names


michael-emmi:
**Describe the bug**

With SharedInitializationLocalSplitter enabled, local variables no longer seem to carry their original names. We had been using this option:
```
Options.v().setPhaseOption("jb", "use-original-names:true");
```
And seeing locals with names from the source program. However, new code #35c5bcb seems to interfere with this; now we just see many `$stack…` variables.

The original names appear again if we disable this pass as mentioned already:
```
Options.v().setPhaseOption("jb.sils", "enabled:false");
```

@linghuiluo has identified this as a defect: https://github.com/soot-oss/soot/issues/1614#issuecomment-853287543

**To reproduce**
Run soot with the `jb:use-original-names` option enabled:
```
Options.v().setPhaseOption("jb", "use-original-names:true");
```
Then compare the names of Jimple local variables with `jb.sils` disabled:
```
Options.v().setPhaseOption("jb.sils", "enabled:false");
```

**Expected behavior**
Jimple local variables are identified with their original source-code names.

**Stacktrace**
N/A

**Input file**
Any

**Additional context**
Preserving the names of local variables has been quite useful for debugging intermediate analysis results. Having to disable `jb.sils` to maintain this functionality is unfortunate.

linghuiluo:
I just tested the develop branch. Now we got sth really nasty. It seems that even disabling the transformer Options.v().setPhaseOption("jb.sils", "enabled:false"); will not help. The original names don't show up at all.

https://github.com/soot-oss/soot/commit/7abeb18d935c31b7bc0ea3524450c218b4e786cb

tim-hoffman:
@linghuiluo I just did a quick test of compiling soot both with and without that commit you mentioned and ran with "-p jb use-original-names:true". There was no change in the output between the 2 compiles which shows that this commit is not the cause of this issue. I did observe that the original names are not appearing in the output class file with either version but the original names do show up if you use Jimple output format. I also tried with Baf output and again the original names are missing (even without that commit you mentioned) so it seems like there was some change affecting the Jimple-to-Baf translation that causes the names to be lost.

linghuiluo:
@tim-hoffman you also merged many PRs in the last days. Not just that commit to that file.

tim-hoffman:
[...]

PR #1642, opened 2021-06-05
Trying to improve issue templates a bit

MarcMil:
Feel free to discuss about the changes. I think that the input files are probably rather overlooked if they are specified later on. When someone wants to submit a bug report, an error description is of course necessary and essentially everybody knows that. But having the input files is a necessary prerequisite in many cases and may not be as obvious as the other points.


linghuiluo:
@MarcMil looks good! However, I think we need to resolve this issue https://github.com/soot-oss/soot/issues/1641 at first before merging more PRs.  Maybe you have a clue what caused the bug https://github.com/soot-oss/soot/issues/1641?

PR #1646, opened 2021-06-09
add -process-jar-dir option

patricklam:
If one doesn't read too closely, one could read the description of -process-dir as already doing this, but it doesn't. When you have a directory full of .JAR files, this new -process-jar-dir option tells Soot to add all of the JAR files in that directory to the Soot classpath. (-process-dir only adds .class files etc from the directories you specify).


swissiety:
i guess we should make the `-process-dir` doc more clear, too - maybe adding a hint to the new option?

patricklam:
OK I've added more text to -process-dir suggesting the use of -process-jar-dir. I may or may not have permission to merge, but probably someone else should.

PR #1648, opened 2021-06-09
various code restructuring and cleanup/formatting

tim-hoffman:


src/main/java/soot/jimple/Stmt.java
```diff
@@ -27,7 +27,6 @@
 import soot.ValueBox;
 
 public interface Stmt extends Unit {
-  public void toString(UnitPrinter up);
```
tim-hoffman, line 30: This method is already inherited from the Unit interface

PR #1650, opened 2021-06-09
Add helpful methods to the Scene

tim-hoffman:
Extract the functionality of the exclude/include methods in the Scene to new methods with a String parameter.


anddann (approved):
Thanks for your work, LGTM!
In the future, it would be awesome if we can exchange such string operations with StringUtils from apache commons-lang

PR #1652, opened 2021-06-09
Remove redundant checks and reduce excess memory use in SootClass

tim-hoffman:
Also clarify some error messages.


swissiety:
- [ ] please remove the checkLevel NOTEs.
- [ ] please elaborate on the reduced excess memory usage if this is not the sideeffect of removing those redundant checks

tim-hoffman:
Thanks for reviewing @swissiety.
I do think those NOTE comments are helpful to have in the locations where I removed those redundant checks. The comments allow people to quickly see that the check should already be performed so they do not try to re-add those checks in the future.
Regarding the memory usage, see the comment at line 814 when the ```interfaces``` HashChain is created. The HashChain internally uses a ConcurrentHashMap which has a default table size of 16. This leads to a large amount of wasted memory allocation in practice because most classes implement very few (often 1) interfaces.

linghuiluo:
> Thanks for reviewing @swissiety.
> I do think those NOTE comments are helpful to have in the locations where I removed those redundant checks. The comments allow people to quickly see that the check should already be performed so they do not try to re-add those checks in the future.
> Regarding the memory usage, see the comment at line 814 when the `interfaces` HashChain is created. The HashChain internally uses a ConcurrentHashMap which has a default table size of 16. This leads to a large amount of wasted memory allocation in practice because most classes implement very few (often 1) interfaces.

I agree with Tim

linghuiluo:
thanks!

PR #1653, opened 2021-06-09
Clarify error messages in FieldRefValidator

tim-hoffman:


src/main/java/soot/jimple/validation/FieldRefValidator.java
```diff
@@ -66,26 +66,25 @@ public void validate(Body body, List<ValidationException> exceptions) {
...
           if (field == null) {
-            exceptions.add(new UnitValidationException(unit, body, "Resolved field is null: " + fr.toString()));
+            exceptions.add(new UnitValidationException(u, body, "Resolved field is null: " + fr.toString()));
           } else if (field.isStatic() && !field.isPhantom()) {
-            exceptions.add(new UnitValidationException(unit, body, "Trying to get an instance field which is static: " + v));
+            exceptions.add(new UnitValidationException(u, body, "Used InstanceFieldRef for a static field: " + v));
           }
         } catch (ResolutionFailedException e) {
-          exceptions.add(new UnitValidationException(unit, body, "Trying to get an instance field which is static: " + v));
+          exceptions.add(new UnitValidationException(u, body, "Unable to resolve FieldRef " + v + ": " + e.getMessage()));
```
linghuiluo, line 87: sometimes you append e.getMessage() to the message, sometimes you don't. The original messages contain what kind (static, or instance) of fields soot was trying to resolve. I don't see improvement in the messages. the variable unit was also better than u.

tim-hoffman: If you take a closer look, I append ```e.getMessage()``` in those cases where an exception ```e``` was caught. In the other cases, there is no exception to append.
Your second sentence I don't understand what you are trying to say. Are you talking about the 2 cases where I totally changed the text String because it was saying the incorrect thing (it seems to have been just copy pasted from an earlier line)? 
Regarding your final comment, I simply changed ```unit``` to ```u``` because it allowed the autoformatting rules (specifically the maximum line length rule) to not split the lines so much and the method was overall small enough that I figured the name ```u``` for the Unit would be obvious enough. But I can change that back if you like as it was simply a matter of preference and formatting.

src/main/java/soot/jimple/validation/FieldRefValidator.java
```diff
@@ -66,26 +66,25 @@ public void validate(Body body, List<ValidationException> exceptions) {
...
-            exceptions.add(new UnitValidationException(unit, body, "Resolved field is null: " + fr.toString()));
+            exceptions.add(new UnitValidationException(u, body, "Resolved field is null: " + fr.toString()));
           } else if (!field.isStatic() && !field.isPhantom()) {
-            exceptions
-                .add(new UnitValidationException(unit, body, "Trying to get a static field which is non-static: " + v));
+            exceptions.add(new UnitValidationException(u, body, "Used StaticFieldRef for a non-static field: " + v));
           }
         } catch (ResolutionFailedException e) {
-          exceptions.add(new UnitValidationException(unit, body, "Trying to get a static field which is non-static: " + v));
+          exceptions.add(new UnitValidationException(u, body, "Unable to resolve FieldRef " + v + ": " + e.getMessage()));
```
linghuiluo, line 74: Doesn't the exception happen when resolving a StaticFieldRef? Similarly, at line 87 it happens when trying to resolve an InstanceFieldRef.

tim-hoffman: Yes, that's obvious from the code. The problem is, a mismatch of static vs non-static is not the reason that a ```ResolutionFailedException``` can occur so the message is misleading. If you look at ```AbstractSootFieldRef.java``` you will also see that ```resolve()``` could also throw ```FieldResolutionFailedException``` (a subclass of ```ResolutionFailedException```) which has nothing to do with the static-ness of the field. That is the reason I replaced the majority of the String formed here with the message from the exception itself because that will provide more specific information.

tim-hoffman: If you think it's better, I could change two message to "Unable to resolve StaticFieldRef" and "Unable to resolve InstanceFieldRef" respectively. That can provide a little extra information (although it's something that looking at the stack trace of the exception would already tell you).

tim-hoffman:
@linghuiluo I've made some significant changes so that this could also handle BafBody instances and removed some of the duplication that was there (with multiple try-catch for instance). Perhaps we should have a unit test for this validator as well. I don't think one exists.

linghuiluo:
> @linghuiluo I've made some significant changes so that this could also handle BafBody instances and removed some of the duplication that was there (with multiple try-catch for instance).
> Perhaps we should have a unit test for this validator as well. I don't think one exists.

thanks, your implementation looks nice. yeah, I agree, some unit tests would be great.

MarcMil:
@StevenArzt What do you think? Would you like to update the branch and run the tests?

PR #1654, opened 2021-06-09
Minor SootResolver optimization

tim-hoffman:
use isBasicClass directly instead of retrieving the full list and then using contains


anddann (approved):
Thanks, LGTM!

PR #1655, opened 2021-06-10
fix UnitBox memory leak in UCE

tim-hoffman:


linghuiluo:
@swissiety can you review this?

src/systemTest/java/soot/jimple/toolkit/scalar/UnreachableCodeEliminatorTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    // Assert that the set of boxes obtained from all units equals the set
+    // obtained by checking Unit#getBoxesPointingToThis() on all units.
+    Assert.assertEquals(new HashSet<>(getAllUnitBoxes(body)), new HashSet<>(getAllBoxesPointingToUnits(body)));
+
+    UnreachableCodeEliminator.v().transform(body);
+
+    // Assert that the set of boxes obtained from all units equals the set
+    // obtained by checking Unit#getBoxesPointingToThis() on all units.
+    Assert.assertEquals(new HashSet<>(getAllUnitBoxes(body)), new HashSet<>(getAllBoxesPointingToUnits(body)));
```
swissiety, line 74: please assert that the transformer changed sth i.e. removed the trap

src/main/java/soot/jimple/toolkits/scalar/UnreachableCodeEliminator.java
```diff
@@ -99,9 +101,16 @@ protected void internalTransform(Body body, String phaseName, Map<String, String
     // occur in practice, and certainly no in code generated from Java source.
     final Chain<Trap> traps = body.getTraps();
     for (Iterator<Trap> it = traps.iterator(); it.hasNext();) {
-      Trap trap = it.next();
-      if ((trap.getBeginUnit() == trap.getEndUnit()) || !reachable.contains(trap.getHandlerUnit())) {
+      final Trap trap = it.next();
+      UnitBox beginBox = trap.getBeginUnitBox();
+      UnitBox endBox = trap.getEndUnitBox();
+      UnitBox handlerBox = trap.getHandlerUnitBox();
+      if ((beginBox.getUnit() == endBox.getUnit()) || !reachable.contains(handlerBox.getUnit())) {
```
swissiety, line 108: please restructure the UnitBox assignments inside the if block so the assignments are just executed when needed

tim-hoffman: I don't think any improvement can be made. The first two (beginBox and endBox) will always be used since they are in the first condition. The final variable (handlerBox) will always be used as well. If the first condition is true, it will be used inside the if statement body and if the first condition is false, it will be used in the second condition (and possibly again inside the body).

src/systemTest/java/soot/jimple/toolkit/scalar/UnreachableCodeEliminatorTest.java
tim-hoffman: done.

MarcMil:
This was fixed more generally in https://github.com/soot-oss/soot/pull/2249/changes/78e81fe0405dba89abeb749e54a8976f3ade24a1 and can thus be closed

PR #1658, opened 2021-06-10
minor SiteInliner improvements

tim-hoffman:


src/main/java/soot/jimple/toolkits/invoke/SiteInliner.java
```diff
@@ -73,7 +73,7 @@ public String getDefaultOptions() {
    * container).
    */
   public static void inlineSites(List<List<Host>> sites) {
-    inlineSites(sites, new HashMap<String, String>());
+    inlineSites(sites, Collections.emptyMap());
```
tim-hoffman, line 76: Avoid creating an unnecessary new HashMap object.

src/main/java/soot/jimple/toolkits/invoke/SiteInliner.java
```diff
@@ -92,8 +93,8 @@ public static void inlineSites(List<List<Host>> sites, Map<String, String> optio
   /**
    * Inlines the method <code>inlinee</code> into the <code>container</code> at the point <code>toInline</code>.
    */
-  public static void inlineSite(SootMethod inlinee, Stmt toInline, SootMethod container) {
-    inlineSite(inlinee, toInline, container, new HashMap<String, String>());
+  public static List<Unit> inlineSite(SootMethod inlinee, Stmt toInline, SootMethod container) {
```
tim-hoffman, line 96: Make return type consistent with the other 'inlineSite' method.

PR #1659, opened 2021-06-10
Cleanup, format, minor opts in Hierarchy classes

tim-hoffman:
- code cleanup and formatting
- reduce repeated method calls with the same result
- replace size()>0 with isEmpty()
- reduce ArrayList memory leaks by trimming or re-creating


src/main/java/soot/FastHierarchy.java
```diff
@@ -113,16 +113,8 @@ public Interval(int lower, int upper) {
...
-        return false;
-      }
-      if (lower > potentialSubrange.lower) {
-        return false;
-      }
-      if (upper < potentialSubrange.upper) {
-        return false;
-      }
-      return true;
+      return (potentialSubrange == this)
```
tim-hoffman, line 116: Here I converted the original checks into a single line and added the (potentialSubrange == this) check. This check implies the original (i.e. if potentialSubrange == this, then the lower and upper fields are equal) but is faster so I added it as an initial fast path.

PR #1660, opened 2021-06-10
Fix bug in AsmMethodSource when LocalVariableTable is inconsistent

tim-hoffman:
If there was a variable use/def not covered by a LocalVariableNode, then replacing the local variable could lead to a "local variable uninitialized" validation error (or worse, initialize to the wrong value!). Thus, before replacing any Locals, check for such inconsistencies and do nothing in that case.


src/systemTest/java/soot/asm/AsmMethodSourceOrigNamesTest.java
```diff
@@ -0,0 +1,88 @@
...
+    final Options opts = Options.v();
+    opts.set_validate(false);
+    opts.setPhaseOption("jb", "use-original-names:true");
+    opts.setPhaseOption("jb.sils", "enabled:false");
+  }
+
+  @Test
+  public void testWriterToUTF8Buffered1() {
+    final String clazz = "org.apache.xml.serializer.WriterToUTF8Buffered";
+    final String[] params = { "char[]", "int", "int" };
```
linghuiluo, line 55: Can you add the Java source of the classes?

tim-hoffman: These are pre-compiled classes that I got from within the DaCapo benchmark suite. The issue only presents itself because whatever compiler was used to produce these class files generated an inconsistent LocalVariableTable. I did manage to find the source for WriterToUTF8Buffered (from xalan-2.6.0) in an online archive and compiled it with javac 1.8.0_121 and the LocalVariableTable was consistent and so could not be used to really test this fix.

linghuiluo: How could the class with inconsistent LocalVariableTable be compiled? Can such classes be even verified by JVM?

StevenArzt: In general: If the code can be verified by the JVM, Soot should support it, even if it's strange / unusual. We have seen all sorts of funny code, especially with obfuscation. We can only safely discard such code if the JVM rejects it.

tim-hoffman: According to the JVM spec, the LocalVariableTable is optional and only used by debuggers to associate local variable names with the local variable indexing used by the bytecode. So I assume no verification of the LVT is performed since it's for debuggers only.

[https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.13](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.13)

linghuiluo: @tim-hoffman I agree that soot should support if jvm verifies the code. Can you use `javap -l` to the bytecode and paste the local variable table and relevant instructions in code comment with some explanation for a test case to help us understand this inconsistency?

src/systemTest/java/soot/asm/AsmMethodSourceOrigNamesTest.java
```diff
@@ -0,0 +1,1263 @@
...
+   *       107: aload_0
+   *       108: getfield      #5                  // Field count:I
+   *       111: istore        7
+   *       113: iload_2
+   *       114: istore        8
+   *       116: goto          133
+   *       119: aload         6
+   *       121: iload         7
+   *       123: iinc          7, 1
+   *    ?  126: iload         9
```
linghuiluo, line 135: @tim-hoffman  Thanks for adding this test! t is a really strange the case. Even for other parts of slot 8, the starting offset in the table is not exactly as where the variable is used.

src/main/java/soot/asm/AsmMethodSource.java
```diff
@@ -2239,72 +2240,107 @@ public Body getBody(SootMethod m, String phaseName) {
...
+            // 2. If any of the LocalVariableNode in 'vals' cover any of the same
+            // units, then they are ambiguous and cannot be used.
+            //
+            // To implement these checks, first collect all ValueBoxes in the body
+            // that reference the chosen Local. Then, as each LocalVariableNode is
+            // processed, map each ValueBox to the new Local that it should hold.
+            // If any ValueBox is found more than once or not found at all, then
+            // one of the inconsistency cases mentioned above exists and thus no
+            // changes should be made.
+            IdentityHashMap<ValueBox, Local> boxToNewLoc = new IdentityHashMap<>();
```
linghuiluo, line 2284: @tim-hoffman  Can you split he following piece of code into multiple methods? The getBody  method gets really long. For example, the code you check for consistency can be one method.

src/main/java/soot/asm/AsmMethodSource.java
```diff
@@ -2239,72 +2240,107 @@ public Body getBody(SootMethod m, String phaseName) {
...
+                      if (conflict != null) {
+                        isConsistent = false;
+                        break LV_LOOP;
+                      }
+                    }
                   }
                 }
               }
             }
+            // Finally, replace the locals only if both consistency conditions pass.
```
linghuiluo, line 2336: The code below could be another method with proper method name and doc.

src/main/java/soot/asm/AsmMethodSource.java
tim-hoffman: I split it into a couple of methods. I did consider making the consistency check inside the loop into its own method but it would end up needing a lot of parameters for all of the local variables at that point so I thought it would be best to leave it.

src/systemTest/java/soot/asm/AsmMethodSourceOrigNamesTest.java
tim-hoffman: Yes, from what I can tell, the starting offset is always the instruction following the store to the slot. I'm not sure why it is this way but it is at least consistent and I have addressed that: https://github.com/soot-oss/soot/blob/c295e358ec509d7ae93cc612a7de5e22c121e8ce/src/main/java/soot/asm/AsmMethodSource.java#L2325-L2327

PR #1667, opened 2021-06-17
minor improvements to ExceptionalUnitGraph and ThrowableSet

tim-hoffman:


src/main/java/soot/toolkits/graph/ExceptionalUnitGraph.java
```diff
@@ -322,8 +325,8 @@ protected void initialize(ThrowAnalysis throwAnalysis, boolean omitExceptingUnit
    * @return a <code>Map</code> which whose contents are equivalent to the input <code>map</code>, plus the information that
    *         <code>u</code> throws <code>caught</code> to <code>t</code>.
    */
-  private Map<Unit, Collection<ExceptionDest>> addDestToMap(Map<Unit, Collection<ExceptionDest>> map, Unit u, Trap t,
-      ThrowableSet caught) {
+  protected Map<Unit, Collection<ExceptionDest>> addDestToMap(Map<Unit, Collection<ExceptionDest>> map, Unit u,
```
tim-hoffman, line 328: Make it protected to allow a subclass to use and/or override it. There is no reason for it to remain private.

src/main/java/soot/toolkits/graph/ExceptionalUnitGraph.java
```diff
@@ -667,8 +670,6 @@ private void buildHeadsAndTails(Set<Unit> additionalHeads) {
     Collection<ExceptionDest> result = unitToExceptionDests.get(u);
     if (result == null) {
       ExceptionDest e = new ExceptionDest(null, null) {
-        private ThrowableSet throwables;
```
tim-hoffman, line 670: ExceptionDest already defines this field, it was just private so it was not accessible. So I removed the private from that field and remove this field so there would be no duplication.

src/main/java/soot/toolkits/graph/ExceptionalUnitGraph.java
```diff
@@ -255,47 +255,50 @@ protected void initialize(ThrowAnalysis throwAnalysis, boolean omitExceptingUnit
...
    *         caught within the method, it will be mapped to a <code>Collection</code> of <code>ExceptionDest</code>s
    *         describing the sets of exceptions that the <code>Unit</code> might throw to each {@link Trap}. But if all of a
-   *         <code>Unit</code>'s exceptions escape the method, it will be mapped to <code>null</code, rather than to a
+   *         <code>Unit</code>'s exceptions escape the method, it will be mapped to <code>null</code>, rather than to a
    *         <code>Collection</code> containing a single <code>ExceptionDest</code> with a <code>null</code> trap. (The
    *         special case for <code>Unit</code>s with no caught exceptions allows <code>buildExceptionDests()</code> to
    *         ignore completely <code>Unit</code>s which are outside the scope of all <code>Trap</code>s.)
    *         </p>
    */
   protected Map<Unit, Collection<ExceptionDest>> buildExceptionDests(ThrowAnalysis throwAnalysis) {
```
tim-hoffman, line 264: Restructured this method a bit to avoid unnecessary work when body.getTraps() is empty and to avoid obtaining ThrowableSet.Manager.v().EMPTY repeatedly.

src/main/java/soot/toolkits/exceptions/ThrowableSet.java
```diff
@@ -87,11 +87,15 @@
...
-
 public class ThrowableSet {
 
   private static final boolean INSTRUMENTING = false;
-  private final SootClass JAVA_LANG_OBJECT_CLASS = Scene.v().getObjectType().getSootClass();
+
+  // This pattern allows for the field to be static but to delay it's
+  // initialization until it's first use.
+  private static final class LazyInit {
+    static final SootClass JAVA_LANG_OBJECT_CLASS = Scene.v().getObjectType().getSootClass();
```
MarcMil, line 97: Nope, we cannot do that; that object would persist throughout a G.reset() and hold a wrong reference then.

MarcMil:
I've already done many of these optimizations in the meantime, thus this PR can be closed IMHO.

Issue #1132, opened 2019-03-17
StackOverflowError


roskad:
Hello,

Soot produces a StackOverflowError with [this apk](https://github.com/Sable/soot/files/2975683/109B67907AAEB3B3477B3191836E2DD810252B2F9B8DA125ABE475D26796D75C.zip).

This is the start of the printstack :
`at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:512) ~[?:?]
	at java.lang.StringBuilder.append(StringBuilder.java:141) ~[?:?]
	at soot.SootField.getSubSignature(SootField.java:102) ~[soot-3.3.0.jar:?]
	at soot.SootField.getSignature(SootField.java:76) ~[soot-3.3.0.jar:?]
	at soot.AbstractSootFieldRef.getSignature(AbstractSootFieldRef.java:84) ~[soot-3.3.0.jar:?]
	at soot.jimple.internal.AbstractInstanceFieldRef.toString(AbstractInstanceFieldRef.java:60) ~[soot-3.3.0.jar:?]
	at soot.jimple.internal.JAssignStmt.toString(JAssignStmt.java:218) ~[soot-3.3.0.jar:?]
	at java.lang.String.valueOf(String.java:2788) ~[?:?]
	at java.lang.StringBuilder.append(StringBuilder.java:135) ~[?:?]
	at soot.jimple.internal.JLookupSwitchStmt.toString(JLookupSwitchStmt.java:97) ~[soot-3.3.0.jar:?]
	at java.lang.String.valueOf(String.java:2788) ~[?:?]
	at java.lang.StringBuilder.append(StringBuilder.java:135) ~[?:?]
	at soot.jimple.internal.JLookupSwitchStmt.toString(JLookupSwitchStmt.java:97) ~[soot-3.3.0.jar:?]
	at java.lang.String.valueOf(String.java:2788) ~[?:?]
	at java.lang.StringBuilder.append(StringBuilder.java:135) ~[?:?]
	at soot.jimple.internal.JLookupSwitchStmt.toString(JLookupSwitchStmt.java:102) ~[soot-3.3.0.jar:?]
	at java.lang.String.valueOf(String.java:2788) ~[?:?]
	at java.lang.StringBuilder.append(StringBuilder.java:135) ~[?:?]
	at soot.jimple.internal.JLookupSwitchStmt.toString(JLookupSwitchStmt.java:97) ~[soot-3.3.0.jar:?]
[...]

PR #1675, opened 2021-06-23
Fix dead loop for nest JLookupSwitchStmt in toString.

hluwa:
fix bug for #1132.

### DIFF
--- a/src/main/java/soot/jimple/internal/JLookupSwitchStmt.java
+++ b/src/main/java/soot/jimple/internal/JLookupSwitchStmt.java
@@ -10,12 +10,12 @@ package soot.jimple.internal;
  * it under the terms of the GNU Lesser General Public License as
  * published by the Free Software Foundation, either version 2.1 of the
  * License, or (at your option) any later version.
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Lesser Public License for more details.
- * 
+ *
  * You should have received a copy of the GNU General Lesser Public
  * License along with this program.  If not, see
  * <http://www.gnu.org/licenses/lgpl-2.1.html>.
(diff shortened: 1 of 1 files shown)


linghuiluo:
@hluwa Can you add a test case for your code?

hluwa:
> @hluwa Can you add a test case for your code?

I'm added a test case, please check.

src/systemTest/java/soot/asm/NestLabelJumpTest.java
```diff
@@ -0,0 +1,127 @@
+package soot.asm;
+
+/*-
+ * #%L
+ * Soot - a J*va Optimization Framework
+ * %%
+ * Copyright (C) 2020 Manuel Benz
```
linghuiluo, line 7: you can change to your name

src/systemTest/targets/soot/asm/NestLabel.java
```diff
@@ -0,0 +1,41 @@
...
+/*-
+ * #%L
+ * Soot - a J*va Optimization Framework
+ * %%
+ * Copyright (C) 2012 Andreas Dann
+ *
+ * (c) 2012 University of Luxembourg - Interdisciplinary Centre for
+ * Security Reliability and Trust (SnT) - All rights reserved
+ * Alexandre Bartel
+ *
```
linghuiluo, line 12: you can change to your name

src/main/java/soot/jimple/internal/JLookupSwitchStmt.java
```diff
@@ -75,6 +76,21 @@ public Object clone() {
     return new JLookupSwitchStmt(getKey(), clonedLookupValues, getTargets(), getDefaultTarget());
   }
 
+  private String toSimpleString() {
+    final char endOfLine = ' ';
+    List<String> cases = lookupValues.stream()
+            .map((lookupValue) -> String.valueOf(lookupValue.value))
+            .distinct()
```
swissiety, line 83: please remove the distinct() as lookupvalues should already be distinct. Printing only non-duplicates could make it harder finding a possible bug somewhere that adds duplicate lookupvalues.

src/main/java/soot/jimple/internal/JLookupSwitchStmt.java
```diff
@@ -75,6 +76,21 @@ public Object clone() {
...
+            .collect(Collectors.toList());
+    if (getDefaultTarget() != null) {
+      cases.add("default");
+    }
+    return Jimple.LOOKUPSWITCH + "(" + keyBox.getValue().toString() + ')' + endOfLine
+              + "{" + endOfLine
+              + "cases:" + String.join(",", cases) + endOfLine
```
swissiety, line 90: please use `stream(). ... .collect(Collectors.joining(",")` directly, as the List of cases is not used elsewhere, removing the need to allocate a list

Issue #1678, opened 2021-06-23
Shimple uninitialized registers


tim-hoffman:
**Describe the bug**
Run Soot with ```-via-shimple``` option on ```PhiNodeTestInput.java``` and the generated class file causes ```java.lang.VerifyError: (class: soot/shimple/PhiNodeTestInput, method: sync signature: ()Z) Accessing value from uninitialized register 7``` when loaded into the JVM.

**Input file**
The input file and a Soot framework test case for can be found in this commit https://github.com/tim-hoffman/soot/commit/b31481667c0e87847d96c6c2cb8204dc6d9c4a1c

**To reproduce**
Option 1: Checkout the commit mentioned above and run the test case via the Soot test build and observe the test failures. 
Option 2: Just download the ```PhiNodeTestInput``` file and on command line run
```java -cp <soot-jar-and-dependencies> soot.Main -pp -cp systemTest-target-classes -via-shimple soot.shimple.PhiNodeTestInput``` and then run ```java -cp sootOutput soot.shimple.PhiNodeTestInput``` and observe the VerifyError.

**Expected behavior**
Running the Soot-modified version of PhiNodeTestInput should simply print ```false``` and terminate normally instead of terminating with a VerifyError.

**Additional context**
The ```PhiNodeTest#testComplexPhi``` case is set up to run Soot on the input file, run several of the Soot built-in validators, and then attempt to load the generated class file, causing the VerifyError to occur. You may have to comment out the ```Assert.assertTrue(exceptions.isEmpty())``` line to see the JVM try to load the class since the Soot Validators may detect the issue and trigger that assertion. The ```DEBUG_PRINT``` flag can be set as ```true``` to print the Soot Validator messages.

The ```PhiNodeTest#testStdLib``` case is set up to run Soot on a selection of standard JRE classes that are known to cause the same issue. It runs Soot on all of those classes without using ```-via-shimple``` to show that there are no issues in that case and then runs again with the ```-via-shimple``` option set to true to show that the generated bytecode is invalid in that case. There is also a ```USE_ASM_VERIFIER``` flag in that file that can be set as ```true``` to use ASM to verify the generated bytecode instead of loading the generated classes into the JRE. This allows the test to run on some additional classes in the ```java.*``` package which could not be tested via the other approach (due to the ClassLoader restriction on the java.* package).

tim-hoffman:
I added an even simpler example that has a similar problem (and made a few other changes).
[...]

PR #1680, opened 2021-06-25
fix Shimple issue with traps

tim-hoffman:


tim-hoffman:
This PR includes the new test cases mentioned in issue #1678 and is intended as a place to work on a solution to that issue.
Commit b1c1dab6219cd2797310bca3ca997c4d9108a0be provides a partial solution but there are still 4 classes in the test case that have the error but it must be due to a slightly different cause.

PR #1681, opened 2021-06-28
correct a simple syntax mistake and fix a bug

canliture:
- correct a simple syntax mistake
[Last](https://github.com/canliture/soot/commit/621a81dbc9179f9e54f825077a278a9846e5d746), I have corrected the mistake, but missed the same problem in the file 'soot_options.xml.orig'

- fix a bug
When I write the following test code:
```
public class SootResolverTest {

    @Test
    public void test1() {
        G.reset();

        // setting
        Options.v().set_whole_program(true);

        // No throw. ^_^
    }

    @Test
    public void test2() {
        G.reset();

        // setting
        Options.v().set_whole_shimple(true);

        // throw java.lang.AssertionError !!!
    }
}
```

'test1' will pass the test and 'test2' will fail to pass the test.

The reason that 'test2' fails to pass the test is written at the header comment of test class 'soot.SootResolverTest':

Before fixing the bug, we will get default java class path when <br/>
- we set <b>Options.v().set_whole_program(true);</b> : `path/to/rt.jar;path/to/jce.jar` <br/>
- we set <b>Options.v().set_whole_shimple(true);</b> : `path/to/rt.jar` <br/>

After fixing the bug, we will get default java class path when <br/>
- we set <b>Options.v().set_whole_program(true);</b> : `path/to/rt.jar;path/to/jce.jar` <br/>
- we set <b>Options.v().set_whole_shimple(true);</b> : `path/to/rt.jar;path/to/jce.jar` <br/>

### DIFF
--- a/src/main/java/soot/Scene.java
+++ b/src/main/java/soot/Scene.java
@@ -782,7 +782,8 @@ public class Scene {
       }
     }
 
-    if (!javaGEQ9 && (Options.v().whole_program() || Options.v().whole_shimple() || Options.v().output_format() == Options.output_format_dava)) {
+    if (!javaGEQ9
+        && (Options.v().whole_program() || Options.v().whole_shimple() || Options.v().output_format() == Options.output_format_dava)) {
       // add jce.jar, which is necessary for whole program mode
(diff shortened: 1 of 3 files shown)


swissiety (approved):
looks good :+1:

PR #1685, opened 2021-06-29
cleanup several validators and make them compatible with BafBody

tim-hoffman:


src/main/java/soot/jimple/validation/ReturnStatementsValidator.java
```diff
@@ -25,63 +25,66 @@
...
 
-  /**
-   * Checks the following invariants on this Jimple body:
-   * <ol>
-   * <li>this-references may only occur in instance methods
-   * <li>this-references may only occur as the first statement in a method, if they occur at all
-   * <li>param-references must precede all statements that are not themselves param-references or this-references, if they
-   * occur at all
-   * </ol>
-   */
```
tim-hoffman, line 53: This comment was incorrect, likely copy pasted from somewhere else.

src/main/java/soot/jimple/validation/ReturnStatementsValidator.java
```diff
@@ -25,63 +25,66 @@
...
   @Override
   public void validate(Body body, List<ValidationException> exceptions) {
-    // Checks that this Jimple body actually contains a return statement, and that the return statement
-    // is of the appropriate type (i.e. void/non-void)
+    final SootMethod method = body.getMethod();
+
+    // Checks that this Body actually contains a throw or return statement, and
+    // that the return statement is of the appropriate type (i.e. void/non-void)
     for (Unit u : body.getUnits()) {
-      if ((u instanceof RetStmt) || (u instanceof ThrowStmt)) {
```
tim-hoffman, line 59: ```RetStmt``` is not a method return, it is the deprecated ```ret``` opcode (that pairs with jsr).

linghuiluo: yeah, I agree. Why is this checked here

linghuiluo (approved):
Looks good to me

PR #1687, opened 2021-06-30
Use logger instead of Sysout

MarcMil:


StevenArzt:
Thne test failures have nothing to do with the changes here. I therefore merge the PR. Someone should definitely have a look at the test failure, though. The failures with the ``ConcurrentModificationException`` seem to be random, since the builds on our local SIT build node are fine.

PR #1688, opened 2021-07-01
fix a FlowAnalysis bug were non-returning loops were skipped

tim-hoffman:


MarcMil:
This is a tough one. I've had some discussions with @StevenArzt on this topic. Thanks for bringing it up!
First, as a recap, your test example as a basis to make clear that we're all on the same page:
We're analyzing the Jimple code shown here:
```
    void run(java.lang.Object)
    {
        java.lang.Throwable $e;
        SimpleLiveLocalsTestDummy this;
        java.lang.Object parameter0;
 
        this := @this: SimpleLiveLocalsTestDummy;
 
        parameter0 := @parameter0: java.lang.Object;
 
        if parameter0 != null goto label4;
 
     label1:
        virtualinvoke this.<SimpleLiveLocalsTestDummy: void run(java.lang.Object)>(parameter0);
 
     label2:
        goto label1;
 
     label3:
        $e := @caughtexception;
 
        goto label1;
 
     label4:
        return;
 
        catch java.lang.Throwable from label1 to label2 with label3;
    }
 ```
We're running the ```LiveLocal``` analysis, which is a backwards analysis, so it starts from the tails of the unit graph. Since the method can only be left via the ```return``` statement, this is the only place where the backwards analysis can start.
We then continue backwards; the problem in this case being that the loop between ```label1``` and ```label4``` is not considered at all by the backwards analysis, since it is not a transitive predecessor of the ```return``` statement.
Therefore, the live locals analysis deems the ```this``` local in the identity reference in the first line as not live, although this definition is used later in the code.

Edit: After some more critical thinking, I'm convinced that this is the proper fix and the changes should be merged.

~So, the problem is that the current behavior - albeit unusual - is actually what you would expect from a backwards flow analysis. Starting backwards there is no way to get to the loop.
What you essentially propose is using a representative statement from within the loop as another starting point. Presumably that would work for the LiveLocals analysis, but implementing that for all backwards analysis might lead to wrong results.~

MarcMil:
Was fixed with my PR, so this issue can be closed @StevenArzt

PR #1697, opened 2021-07-11
[Bug fix]return incorrect nullness in the method Scene#getTypeUnsafe

canliture:
In method soot.jimple.toolkits.callgraph.OnFlyCallGraphBuilder#addType, it calls Scene#getTypeUnsafe(String),
but the argument passed into the method Scene#getTypeUnsafe(String) may be <b>quoted</b>, just like: <br/>
<li>"sun.reflect.'annotation'.AnnotationType"</li> <br/>
<li>"java.lang.'annotation'.Annotation" </li> <br/>

Scene#getTypeUnsafe(String) will return null if the argument passed into is <b>quoted</b>, it will lead to Soot crashing with NullPointerException or IllegalArgumentException somewhere, just like the following 'test3' failing to pass the test with IllegalArgumentException

Here is a minimal test for evaluating the bug fixing.

<pre>
public class SootResolverTest {
    @Test
    public void test3() {
        G.reset();

        Options.v().set_whole_program(true);

        /* returnType maybe be null in SootMethodRefImpl's constructor, resulting in throwing IllegalArgumentException */
        PackManager.v().runPacks();
    }
}
</pre>

Before fixing the bug, soot will crash with the messages:
<pre>
java.lang.IllegalArgumentException: Attempt to create SootMethodRef with null returnType

	at soot.SootMethodRefImpl.<init>(SootMethodRefImpl.java:86)
	at soot.Scene.makeMethodRef(Scene.java:1962)
	at soot.jimple.toolkits.callgraph.OnFlyCallGraphBuilder.addType(OnFlyCallGraphBuilder.java:608)
	at soot.jimple.toolkits.callgraph.CallGraphBuilder.processReceivers(CallGraphBuilder.java:201)
	at soot.jimple.toolkits.callgraph.CallGraphBuilder.process(CallGraphBuilder.java:128)
	at soot.jimple.toolkits.callgraph.CallGraphBuilder.build(CallGraphBuilder.java:114)
	at soot.jimple.toolkits.callgraph.CHATransformer.internalTransform(CHATransformer.java:54)
	at soot.SceneTransformer.transform(SceneTransformer.java:36)
	at soot.Transform.apply(Transform.java:105)
	at soot.RadioScenePack.internalApply(RadioScenePack.java:64)
	at soot.jimple.toolkits.callgraph.CallGraphPack.internalApply(CallGraphPack.java:61)
	at soot.Pack.apply(Pack.java:118)
	at soot.PackManager.runWholeProgramPacks(PackManager.java:619)
	at soot.PackManager.runPacksNormally(PackManager.java:500)
	at soot.PackManager.runPacks(PackManager.java:425)
        ... ...
        ... ...
</pre>

### DIFF
--- a/src/main/java/soot/Scene.java
+++ b/src/main/java/soot/Scene.java
@@ -1031,12 +1031,35 @@ public class Scene {
    * representation of the type is an array, the returned type will be an ArrayType with the base type resolved as described
    * above.
    *
+   * Notice: Recommend to use {@link #getTypeUnsafeUnescape}
+   *
    * @param arg
    *          A string description of the type
    * @return The Type if it can be resolved and null otherwise
    */
   public Type getTypeUnsafe(String arg) {
     return getTypeUnsafe(arg, true);
+  }
+
+  /**
+   * The method do the same thing as the method {@link #getTypeUnsafe(String)} except for unescaping the argument.
+   *
+   * An argument passed into the method {@link #getTypeUnsafe} may be <b>quoted</b>, like them: <br/>
+   * <li>1. "sun.reflect.'annotation'.AnnotationType"<li/> <br/>
+   * <li>2. "java.lang.'annotation'.Annotation"<li/> <br/>
+   *
+   * If calling {@link #getTypeUnsafe(String)} with <b>quoted</b> argument, the method maybe return null. <br/>
+   *
+   * For example, after we call {@link #loadNecessaryClasses}, getTypeUnsafe("java.lang.'annotation'.Annotation") <br/>
+   * will return null but getTypeUnsafe("java.lang.annotation.Annotation") will not return null.
+   *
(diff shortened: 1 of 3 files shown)


MarcMil:
I'm afraid that merging this would cause even more chaos with the quotations. Actually, the quoting is only necessary for  the Jimple parser. In my opinion, instead of introducing unquoting at certain locations, it would be much better to _only_ quote when writing Jimple files. Conversely, we should unquote the names when parsing Jimple code.
Second opinion? @StevenArzt

StevenArzt:
Agreed. In the AST, we don't need quotes. That's only a problem when we have a textual representation of the code.

MarcMil:
@StevenArzt do you close the PR? I'll prepare a better fix.

Issue #1698, opened 2021-07-13
Better handling of AsyncTask edges


flankerhqd:
Actually we can have a better handling for AsyncTask, the functions for AsyncTask current now all have edge type of `AsyncTask`, as shown in the following xml snippet:

```xml
	<!-- Android AsyncTask -->
    <edge type="ASYNCTASK">
        <source invoketype="instance" subsignature="android.os.AsyncTask execute(java.lang.Object[])" />
        <targets>
            <direct subsignature="void onPreExecute()" target="base"/>
            <direct subsignature="java.lang.Object doInBackground(java.lang.Object[])" target="base"/>
            <direct subsignature="void onProgressUpdate(java.lang.Object[])" target="base"/>
            <direct subsignature="void onPostExecute(java.lang.Object)" target="base"/>
        </targets>
    </edge>
```

This lead to several problems, it may lead to confusion on return types in SPARK PAG, e.g. for the following code:
```java

public class AsyncTaskTestMainSample {

    AsyncTask task;
    public void target1() {
        task = new LongOperation().execute("stub");
    }
}

class LongOperation extends AsyncTask<String, Void, String> {
    @Override
    protected String doInBackground(String... params) {
        return null;
    }

    @Override
    protected void onPostExecute(String result) {

    }

    @Override
    protected void onPreExecute() {

    }

    @Override
    protected void onProgressUpdate(Void... values) {
    }
}
```

SPARK PAG will fails with a null pointer in PAG.addCallTarget -> PAG.makeLocalVarNode when constructing call graph, connecting this statement to AsyncTask.onPreExecute. The reason is `onPreExecute` returns VOID, so it cannot be assigned as a return value. Also, the return type of `doInBackground` will overwrite the SPARK node of `execute`, lead to possible confusion in type checking.

It might be better to assign `onPreExecute` and others normal `FAKE` attribute, split them from `doInBackground` like the following:
```
    <edge type="ASYNCTASK">
        <source invoketype="instance" subsignature="android.os.AsyncTask execute(java.lang.Object[])" />
        <targets>
            <direct subsignature="java.lang.Object doInBackground(java.lang.Object[])" target="base"/>
        </targets>
    </edge>
    <edge>
        <source invoketype="instance" subsignature="android.os.AsyncTask execute(java.lang.Object[])" />
        <targets>
            <direct subsignature="void onPreExecute()" target="base"/>
            <direct subsignature="void onProgressUpdate(java.lang.Object[])" target="base"/>
[...]

PR #1702, opened 2021-07-15
Update test case for AsyncTask Spark PAG handling

flankerhqd:
As referred in https://github.com/soot-oss/soot/issues/1698. 

It can be fixed by adding check in `addCallTarget`, but do we have more elegant options?


flankerhqd:
The test case now has passed after the fix.

https://github.com/soot-oss/soot/issues/1698

PR #1743, opened 2021-09-09
Handle nodes without predecessors in MHGDominators

benSepanski:
This would allow nodes with no predecessors to have empty dominator sets


linghuiluo:
It would be nice if you provide a test case

benSepanski:
Sorry about that. Will do. Is there any request for the format of the test e.g. junit? I don't see a designated format in the guidelines

linghuiluo:
jnit is good enough. Thanks

MarcMil:
This seems to fall apart when a node with no predecessor has a successor, because that successor will have the set initialized to the full set again.

MarcMil:
I've added a test case and fixed the bug (hopefully) properly:
https://github.com/soot-oss/soot/pull/2251

PR #1761, opened 2021-10-17
add step to upload javadoc and options to github pages

anddann:
I've created two additional GitHub actions steps to deploy Soot's JavaDoc and Options documentation in GH pages directly to get rid of the old soot-build server.

To do so, I've adapted the approach 
https://github.com/JamesIves/github-pages-deploy-action

However, I'm not an expert with GH Pages and GH actions, thus no guarantee. 😸

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,7 @@ jobs:
           cp ./doc/soot_options.css ${docdir}/options
 
       - name: Deploy JavaDoc to GH pages
-          uses: JamesIves/github-pages-deploy-action@4.1.5
-          with:
-            branch: gh-pages # The branch the action should deploy to.
-            folder: build/docs/ # The folder the action should deploy.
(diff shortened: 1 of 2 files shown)


anddann:
@mbenz89 I've added a script to add an autogenerated overview of all pages

Should I change the links in the readme now, or only after the first run successfully passed.
I would prefer the latter. Since the old links still proceed to work. That is why I did not delete the "old" deployment step immediatly.

mbenz89:
@anddann 
> I've added a script to add an autogenerated overview of all pages
Awesome!

> Should I change the links in the readme now, or only after the first run successfully passed.

After everything worked as intended I'd say.

Let's activate the doc/options deployment for this branch and see if it works as intended.

mbenz89:
@anddann I created a new PR for which I have access to the code so I can play around with it a bit and test the deployment: https://github.com/soot-oss/soot/pull/1801

anddann:
thx for your review

PR #1791, opened 2021-11-11
Extensibility and performance

MarcMil:


eeewxj:
the following change leads to an endless loop:
================================================================
 src/main/java/soot/jimple/toolkits/callgraph/ClinitElimAnalysis.java 
@@ -54,6 +54,11 @@ public void copy(FlowSet<SootMethod> src, FlowSet<SootMethod> dest) {
    src.copy(dest);
  }

57 +  @Override
58 +  protected void copyFreshToExisting(FlowSet<SootMethod> in, FlowSet<SootMethod> dest) {
59 +    dest.copyFreshToExisting(in);
60 +  }

  // out(s) = in(s) intersect { target methods of s where edge kind is clinit}
  @Override
  protected void flowThrough(FlowSet<SootMethod> inVal, Unit stmt, FlowSet<SootMethod> outVal) {
================================================================
in line 59, parameter in should be used as source, i.e. should be " 59 +    in.copyFreshToExisting(dest);"
@MarcMil @StevenArzt

MarcMil:
Created a follow-up PR https://github.com/soot-oss/soot/pull/1793
Goes to show that we apparently do not have any test cases for this analysis right now ;)

PR #1817, opened 2021-12-10
LocalPacker can generate code that doesn't pass the JVM verifier

bdemsky:
The java verifier considers that all operations in a try block could go the exception handler even if there is closer exception handler for the same type of exception.  The Soot analysis attempts to be more sophisticated and can determine that an exception will go the closest enclosing exception handler.


StevenArzt:
The local packer is also used for other targets such as Android / Dalvik bytecode. I wonder whether the same problem appears there as well. If not, I am fine to use (unnecessary) less precise local packing to reduce the complexity of our code. Still, we need to ensure that we don't break anything with regard to Dalvik code.

Also keep in mind that we may have inputs from multiple sources, e.g., an APK file and the corresponding Android platform JAR. We need to write out correct code depending on the *output format*. In the example, we would use Dalvik semantics when writing out an app, even though some inputs are class files from a JAR.

PR #1834, opened 2022-02-01
fix AsmMethodSource assignReadOps(...)

swissiety:
There is a if condition in `assignReadOps(Local l)` which is a contradiction and therefore always renders to false.
The intention seems to filter for Operands that resemble a (static) field read or an array load - if it is not the case the method is skipping the assignment of the `Operand` to a `Local`.
As the if statement currently always renders to false (i.e. no skips for that check) there have been (possibly) more `Local` assignments than necessary (and/or just a obsolete/useless check).

Additionally I restructured the if-stmts to avoid redundant condition evaluations.

### DIFF
--- a/src/main/java/soot/asm/AsmMethodSource.java
+++ b/src/main/java/soot/asm/AsmMethodSource.java
@@ -573,14 +573,15 @@ public class AsmMethodSource implements MethodSource {
       if (opr == DWORD_DUMMY || opr.stack != null) {
         continue;
       }
-      if (l == null && opr.value instanceof Local) {
-        continue;
-      }
-      int op = opr.insn.getOpcode();
-      if (l == null && op != GETFIELD && op != GETSTATIC && (op < IALOAD || op > SALOAD)) {
-        continue;
-      }
-      if (l != null && !opr.value.equivTo(l)) {
+      if (l == null) {
+        if (opr.value instanceof Local) {
+          continue;
+        }
+        int op = opr.insn.getOpcode();
+        if (op != GETFIELD && op != GETSTATIC && (op < IALOAD || op > SALOAD)) {
+          continue;
+        }
+      } else if (!opr.value.equivTo(l)) {
         List<ValueBox> uses = opr.value.getUseBoxes();
         boolean noref = true;
         for (ValueBox use : uses) {
--- a/src/main/java/soot/asm/AsmMethodSource.java
+++ b/src/main/java/soot/asm/AsmMethodSource.java
@@ -588,7 +588,7 @@ public class AsmMethodSource implements MethodSource {
         }
       }
       int op = opr.insn.getOpcode();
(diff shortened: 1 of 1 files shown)


src/main/java/soot/asm/AsmMethodSource.java
```diff
@@ -566,14 +566,19 @@ protected Local newStackLocal() {
...
         continue;
       }
-      if (l != null && !opr.value.equivTo(l)) {
+      if (l == null) {
+        if (opr.value instanceof Local) {
+          continue;
+        }
+        int op = opr.insn.getOpcode();
+        if (op != GETFIELD && op != GETSTATIC && (op < IALOAD || op > SALOAD)) {
+          continue;
```
linghuiluo, line 579: why are you changing `p < IALOAD && op > SALOAD` to  `op < IALOAD || op > SALOAD`?

linghuiluo:
Can you add a test case to this problem you are fixing?

src/main/java/soot/asm/AsmMethodSource.java
swissiety: `op < IALOAD && op > SALOAD` was the contradiction which made this expression always false (due to IALOAD < SALOAD) 
=> op can never be smaller than the smaller value (IALOAD) and greater than the bigger value (SALOAD) at the same time.

I think the original intention of this code was to continue (i.e *not* handling a read op) if the op code is *not* in:
a) one of the field read ops
b) the array read op range (from IALOAD to SALOAD, inclusive)
the type specific array op range translates to: (op >= IALOAD && op <= SALOAD)
Not wanting to skip in that case: !(op >= IALOAD && op <= SALOAD)
DeMorgan: (op < IALOAD || op > SALOAD)

linghuiluo: @swissiety I see, it makes sense. How did you find this bug? Can you provide a test case for it?  I would not merge code to this class without a test demonstrating the issue.

swissiety: IntelliJ highlighted it as always false and it bothered me almost since the beginning of FutureSoot - so i finally tried to make sense of it. A test case is hard to create (for me currently) as we need to manipulate the bytecode input into a specific way so that operands could be something else than those types of read ops - I'm not sure if this is even possible per per jvm specification sue to missing knowledge on my side. If its not possible to have other Operands the whole if condition could be removed from the code - just like its not doing anything at the moment.
If it is possible to have other types of operands (e.g. constants or anything else that is a value - which soot is able to assign - so there should not occur a problem) the code currently just reassigns those values to a Local which would result in increasing the jimple code unnecessarily.

swissiety: got an idea for a possible test - I was to fixated at the bytecode itself..

linghuiluo: @swissiety great! let me now when you have the test.

linghuiluo (changes requested):

PR #1857, opened 2022-04-21
Fix: null check

JordanSamhi:
Null check makes sense here since one could use `findEdge` to search an edge that does not exist --> therefore a `NullPointerException` occurs.


jpstotz:
Indentation of your code changes do not follow the code style rules for this project, please correct this. 

You can check code style by executing `mvn checkstyle:check` before committing.
https://github.com/soot-oss/soot/blob/develop/CONTRIBUTING.md#user-content-style-guidelines

JordanSamhi:
My bad, I just fixed the style..

PR #1861, opened 2022-04-24
potential bug fix for getSootClass

dah-fari7009:
To preserve intended functionality, getSootClass should not build phantom class 

- Might lead to unwanted Scene hierarchy modifications/pointer analysis removal

### DIFF
--- a/src/main/java/soot/Scene.java
+++ b/src/main/java/soot/Scene.java
@@ -1154,7 +1154,7 @@ public class Scene {
    * @return The class if it exists, otherwise null
    */
   public SootClass getSootClassUnsafe(String className) {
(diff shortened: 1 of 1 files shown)


StevenArzt:
This proposed change must be checked very carefully. If I remember correctly, Soot originally, i.e., before the overload with the Boolean parameter was introduced, always created a phantom class if a requested class did not exist and phantom classes were enabled in the Soot options. Therefore, the default flag was ``true`` when the overload came along. If I'm not mistaken, your proposed change would alter this semantic and can therefore break things.

dah-fari7009:
Thank you for the clarification! I'll close this pull request for now, in that case

PR #1870, opened 2022-06-01
Unpack units from unit containers

michael-emmi:
When running Soot with the following command line…
```
java -cp <myclasspath> soot.Main -v -w -keep-line-number -process-dir <my.jar> -cp <my.jar> -pp -allow-phantom-refs -p jb use-original-names:true
```

… we are getting the following exception.
```
Soot started on Wed Jun 01 16:06:44 UTC 2022
java.util.NoSuchElementException: HashChain.LinkIterator.next() reached end of chain without reaching specified tail unit
        at soot.util.HashChain$LinkIterator.next(HashChain.java:609)
        at soot.PatchingChain$PatchingIterator.next(PatchingChain.java:325)
        at soot.PatchingChain$PatchingIterator.next(PatchingChain.java:296)
        at soot.asm.AsmMethodSource.tryCorrectingLocalNames(AsmMethodSource.java:2391)
        at soot.asm.AsmMethodSource.getBody(AsmMethodSource.java:2259)
        at soot.jimple.toolkits.annotation.LineNumberAdder.internalTransform(LineNumberAdder.java:61)
        at soot.PackManager.runPacksNormally(PackManager.java:496)
        at soot.PackManager.runPacks(PackManager.java:425)
        at soot.Main.run(Main.java:280)
        at soot.Main.main(Main.java:142)
```

Unfortunately I am unable to share the source or byte code. But this PR fixes the issue.

### DIFF
--- a/src/main/java/soot/asm/AsmMethodSource.java
+++ b/src/main/java/soot/asm/AsmMethodSource.java
@@ -2373,10 +2373,16 @@ public class AsmMethodSource implements MethodSource {
               for (AbstractInsnNode i = lvn.start.getPrevious(); (uStart = units.get(i)) == null;) {
                 i = i.getNext();
               }
+              if (uStart instanceof UnitContainer) {
+                uStart = ((UnitContainer) uStart).getFirstUnit();
+              }
               // Get the previous real instruction before 'end'
               Unit uEnd;
               for (AbstractInsnNode i = lvn.end.getPrevious(); (uEnd = units.get(i)) == null;) {
                 i = i.getPrevious();
+              }
+              if (uEnd instanceof UnitContainer) {
+                uEnd = ((UnitContainer) uEnd).getFirstUnit();
               }
               if (newLocals == null) {
                 newLocals = HashBasedTable.create(this.maxLocals, 1);


linghuiluo:
@tim-hoffman  Hi Timothy, can you also review this fix as the relevant code was contributed in one of your PRs https://github.com/soot-oss/soot/pull/1660/files? Thanks

src/main/java/soot/asm/AsmMethodSource.java
```diff
@@ -2373,11 +2373,17 @@ protected void tryCorrectingLocalNames(final Jimple jimp, final JimpleBody jb) {
               for (AbstractInsnNode i = lvn.start.getPrevious(); (uStart = units.get(i)) == null;) {
                 i = i.getNext();
               }
+              if (uStart instanceof UnitContainer) {
```
tim-hoffman, line 2376: It seems reasonable to me that `uStart` could be a `UnitContainer` since `mergeUnits(..)` could add a `UnitContainer` to the `units` map. In the absence of a test case for this PR, getting the first Unit from the container seems reasonable for assigning `uStart` with the first Unit in the range defined by the `LocalVariableNode`.

src/main/java/soot/asm/AsmMethodSource.java
```diff
@@ -2373,11 +2373,17 @@ protected void tryCorrectingLocalNames(final Jimple jimp, final JimpleBody jb) {
...
+                uStart = ((UnitContainer) uStart).getFirstUnit();
+              }
               // Get the previous real instruction before 'end'
               Unit uEnd;
               for (AbstractInsnNode i = lvn.end.getPrevious(); (uEnd = units.get(i)) == null;) {
                 i = i.getPrevious();
               }
+              if (uEnd instanceof UnitContainer) {
```
tim-hoffman, line 2384: As above, it makes sense that there could be a `UnitContainer` here. However, in this case, `uEnd` is supposed to be assigned with the final Unit in the range defined by the `LocalVariableNode` so it seems unlikely to me that using `getFirstUnit()` is appropriate here. I would assume getting the last Unit in the `UnitContainer` would be the correct approach for this case. However, without a test case for this PR that utilizes this new code, I can't say for sure.

linghuiluo: The merge only happens here https://github.com/soot-oss/soot/blob/478ba74647dad3be72b696c999a7debeddedc754/src/main/java/soot/asm/AsmMethodSource.java#L1850 and it ensures there is only one local when converting instructions on different paths but actually the same local. The firstUnit is then used to decide the local: https://github.com/soot-oss/soot/blob/a2eb26fded4a4a7107aa3530635ab6a3ca0127ad/src/main/java/soot/asm/StackFrame.java#L174

So I think this should be correct.

PR #1871, opened 2022-06-03
Fix release hierarchy

dah-fari7009:
Potential patch for released hierarchy when retrieving abstract class (android.os.AsyncTask), addresses https://github.com/secure-software-engineering/FlowDroid/issues/465

### DIFF
--- a/src/main/java/soot/Scene.java
+++ b/src/main/java/soot/Scene.java
@@ -1176,11 +1176,11 @@ public class Scene {
     RefType type = nameToClass.get(className);
     if (type != null) {
       synchronized (type) {
-        if (type.hasSootClass() || !SootClass.INVOKEDYNAMIC_DUMMY_CLASS_NAME.equals(className)) {
-          SootClass tsc = type.getSootClass();
-          if (tsc != null) {
-            return tsc;
-          }
+        if (type.hasSootClass()) {
+          SootClass tsc = type.getSootClass(); //here this will add a new (non phantom class) and invalidate the hierarchy
+            if (tsc != null) {
+              return tsc;
+            }
         }
       }
     }
--- a/src/main/java/soot/Scene.java
+++ b/src/main/java/soot/Scene.java
@@ -1154,7 +1154,7 @@ public class Scene {
    * @return The class if it exists, otherwise null
    */
   public SootClass getSootClassUnsafe(String className) {
-    return getSootClassUnsafe(className, true);
+    return getSootClassUnsafe(className, false);
   }
 
   /**
--- a/src/main/java/soot/Scene.java
+++ b/src/main/java/soot/Scene.java
@@ -1178,9 +1178,9 @@ public class Scene {
       synchronized (type) {
         if (type.hasSootClass()) {
           SootClass tsc = type.getSootClass();
-            if (tsc != null) {
-              return tsc;
-            }
+          if (tsc != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/soot/Scene.java
```diff
@@ -1176,11 +1176,11 @@ public SootClass getSootClassUnsafe(String className, boolean phantomNonExist) {
     RefType type = nameToClass.get(className);
     if (type != null) {
       synchronized (type) {
-        if (type.hasSootClass() || !SootClass.INVOKEDYNAMIC_DUMMY_CLASS_NAME.equals(className)) {
```
StevenArzt, line 1179: Why did you remove this check?

dah-fari7009: Because invoking getSootClass (in the following statement) will always add a non phantom class (and invalidate the hierarchy). So here, I'm only invoking it if there was a SootClass to begin with.

An alternative solution could be for getSootClass to not add a class if there isn't already one, but I wasn't sure it wouldn't be a breaking change

StevenArzt: Sounds legit, but then I'd add the chreck for invokedynamic class afterwards to maintain the original semantics.

dah-fari7009: There was no other behavior guarded by the check for invokedynamic class, which is why I removed it altogether (but it's still there in the next if statement same as before)
image

StevenArzt:
@dah-fari7009 Have a look at the failed checks, please. If the checks complete, I can merge the MR.

StevenArzt:
Test failure seems unrelated to MR.

dah-fari7009:
Thank you! I was struggling to figure out the issue :)

linghuiluo:
@StevenArzt @dah-fari7009 There is a failing test since merge of this PR. Can you fix it, as it is blocking other PRs? https://github.com/soot-oss/soot/actions/runs/2435708643

StevenArzt:
I committed a fix. The problem was not this MR, there was a bug that was hidden by the old faulty behavior that this MR fixed.

Background: When loading a class with invoke-dynamic instructions, the classes of the arguments were not added to the dependencies of the class that contains the instruction. Therefore, this class was not put on the resolver worklist. When the body with the invoke-dynamic was later loaded, the method source called ``getSootClass()`` to retrieve the argument class. Before this MR, a new non-phantom class was silently created, and since the resolver was luckily still running, it got resolved. However, the same thing happened with just any class, so you got non-phantom classes for pretty much everything. With my fix, the dependency is added by the class source, and ``getMethod()`` sees the correct class. The MR still remains valid, i.e., if there is no class, we need to create a phantom. If this breaks stuff, that happens because someone forgot to properly declare their dependencies during resolving and was lucky enough to get away with it.

linghuiluo:
@StevenArzt Thanks a lot for the fix and explanation! I understand that the failing test was caused by another bug, but I think we need to have a policy that all tests must pass if a PR wants to get merged. What do you think?

StevenArzt:
@linghuiluo Sure. In general, that makes a lot of sense. It's quite an unfortunate case that a fix unearths a different bug that was previously hidden. It's ok to fail the MR test in such a case, but the contributer usually needs help to fix such problems that are not their fault (and that are usually burried somewhere deep in Soot).

PR #1905, opened 2022-09-29
updated to new axml release

DannyGooo:
make axml version consistent between flowdroid and soot.


DannyGooo:
Soot use the old version for axml but flowdroid use the latest version of axml. This inconsistency would lead to error when running running the flowdroid on latest version soot.

Issue #1811, opened 2021-12-08
[java.util.ConcurrentModificationException]Caused by Collections.synchronizedList in SootClass


Zongyin-Hao:
&ensp; &ensp; When I used soot to analyze andorid classes, I encountered a concurrency bug:
```
[Thread-13] ERROR heros.solver.CountingThreadPoolExecutor - Worker thread execution failed: Failed to apply jb to <com.ichi2.utils.JSONObject: java.lang.Object get(java.lang.String)>
java.lang.RuntimeException: Failed to apply jb to <com.ichi2.utils.JSONObject: java.lang.Object get(java.lang.String)>
	at soot.asm.AsmMethodSource.getBody(AsmMethodSource.java:2282)
	at soot.SootMethod.retrieveActiveBody(SootMethod.java:446)
	at soot.PackManager.lambda$retrieveAllBodies$2(PackManager.java:1252)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1043)
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:997)
	at soot.FastHierarchy.getSignaturePolymorphicMethod(FastHierarchy.java:932)
```
&ensp; &ensp; I tried debugging and found that it may be caused by `` List<SootMethod> methodList `` in ``soot.SootClass``.
&ensp; &ensp; According to ``soot.SootClass:679`` and ``soot.SootClass:700``(both are ``this.methodList = Collections.synchronizedList(new ArrayList<>());``), we can know that ``methodList`` is a `synchronizedList`.
&ensp; &ensp; However, ``synchronizedList`` can only guarantee the safety of ``reading`` and ``writing``, but not the safety of ``iteration``. 
&ensp; &ensp; This is the description of iterator in the ``synchronizedList``'s source code: 
```
public ListIterator<E> listIterator() {
    return list.listIterator(); // Must be manually synched by user
}

public ListIterator<E> listIterator(int index) {
    return list.listIterator(index); // Must be manually synched by user
}
```
[...]

Issue #1199, opened 2019-08-30
ConcurrentModificationException when using Options.v().set_ignore_resolution_errors(true);


muthu30011997:
I got below exception 

java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at soot.jimple.toolkits.annotation.LineNumberAdder.internalTransform(LineNumberAdder.java:58)
	at soot.PackManager.runPacksNormally(PackManager.java:491)
	at soot.PackManager.runPacks(PackManager.java:419)
	at soot.Main.run(Main.java:269)
	at soot.Main.main(Main.java:141)

when i set Options.v().set_ignore_resolution_errors(true); 

How can i solve this problem ?

muthu30011997:
Any updates on this ?

michael-emmi:
I have a very small reproducer for this
```
cat C.java I.java && javac C.java I.java && jar cf C.jar C.class && java -cp ... soot.Main -process-dir C.jar -keep-line-number -ignore-resolution-errors -allow-phantom-refs
abstract class C implements I {
    void bar() {
        try {
            foo();
        } catch (Exception e) {
            throw e;
        }
    }
}
interface I {
    void foo();
}
Soot started on Fri Jul 01 19:34:23 UTC 2022
java.util.ConcurrentModificationException
        at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1013)
        at java.base/java.util.ArrayList$Itr.next(ArrayList.java:967)
        at soot.jimple.toolkits.annotation.LineNumberAdder.internalTransform(LineNumberAdder.java:57)
        at soot.PackManager.runPacksNormally(PackManager.java:496)
        at soot.PackManager.runPacks(PackManager.java:425)
        at soot.Main.run(Main.java:280)
        at soot.Main.main(Main.java:142)

```

michael-emmi:
And here is my rough understanding of the root cause:
1. line number adder iterates over the methods of class C
2. runs jb pack on body of bar, including the unreachable code eliminator (uce)
3. uce creates an exceptional unit graph for bar, initializing exception destinations using unit throw analysis
4. unit throw analysis processes method references, trying to resolve foo
5. foo cannot be resolved, as it is defined in interface I which is not included in input jars
6. fallback logic in soot method ref impl creates an artificial method for foo in class C
[...]

PR #1953, opened 2023-02-27
Returning view of method list to avoid CME

CirQ:
The modified method `getMethodsByNameAndParamCount` is used (and only used) in the `FastHierarchy.getSignaturePolymorphicMethod` for resolving polymorphic methods. The for-loop will return the method list's iterator, but the list itself can be modified by another body pack transform thread, causing a `ConcurrentModificationException`. The patch fix the issue by returning a copy of this list.

This patch also fixes the issue #1811 , and a similar issue is #1199 (fixed by #1886 ).

### DIFF
--- a/src/main/java/soot/SootClass.java
+++ b/src/main/java/soot/SootClass.java
@@ -1316,7 +1316,7 @@ public class SootClass extends AbstractHost implements Numberable {
    */
   public Collection<SootMethod> getMethodsByNameAndParamCount(String name, int paramCount) {
     List<SootMethod> result = null;
-    for (SootMethod m : getMethods()) {
+    for (SootMethod m : new ArrayList<>(getMethods())) {
       if (m.getParameterCount() == paramCount && m.getName().equals(name)) {
         if (result == null) {
           result = new ArrayList<>();


StevenArzt:
The proposed fix for the multi-threading problem doesn't resolve the root cause. The copy constructor of the lists iterates over the base collection and adds each element. Your fix reduces the time during which a concurrent modification can occur to the time of copying the list. That's for sure muchsorter than the runtime of the transformer itself, but if you're unlucky, you can still hit a ``ConcurrentModificationException`` if the list is modified while the copy constructor is running.

I am slightly reluctant to accept an incomplete fix that - on the other hand - migth cost us some performance because the ``FastHierarchy`` is called many times.

CirQ:
> The proposed fix for the multi-threading problem doesn't resolve the root cause. The copy constructor of the lists iterates over the base collection and adds each element. Your fix reduces the time during which a concurrent modification can occur to the time of copying the list. That's for sure muchsorter than the runtime of the transformer itself, but if you're unlucky, you can still hit a `ConcurrentModificationException` if the list is modified while the copy constructor is running.
> 
> I am slightly reluctant to accept an incomplete fix that - on the other hand - migth cost us some performance because the `FastHierarchy` is called many times.

I see. I encountered this issue in an Android analysis project, while in the experiment, a large amount of apks fail due to CME at that site. This patch works fine for these apks.

Anyway, is there any better solution? I found the list itself is a wrapped synchronized collection (https://github.com/soot-oss/soot/blob/develop/src/main/java/soot/SootClass.java#L680), while the iterator returned by this wrapper does not guarantee thread-safety. JDK officially requires synchronization on the iterating site (https://stackoverflow.com/a/1775738). Will this solution be better? If so, I will create a new PR for that.

StevenArzt:
We should first understand why one thread is searching for a method in a class in which another thread is still creating methods. Let's assume the thread that queries the ``FastHierarchy`` runs first. In that case, we don't find the method, because the second threat hasn't created it yet. Let's assume the other way round: We get the method, because the other thread happened to be faster. Does this mean that our results are dependent on thread scheduling, i.e., luck?

My guess would be that we're dealing with phantom methods which are created when resolving a method reference to a callee that doesn't exist. If that is correct, the ``FastHierarchy`` would return cast-incompatibility anyway.

MarcMil:
It seems that the main culprit for adding methods while the FastHierarchy is being constructed is adding  lambda bootstrapping methods. The current implementation of concurrent retrieval of method name and parameter count avoids this problem because the real code does not reference these classes directly. This way a problem when soot used to iterate over all methods though.

PR #1962, opened 2023-03-24
Sync with original soot branch

LightningRS:
Sync with original soot branch


LightningRS:
sorry for the mistake

PR #1967, opened 2023-04-19
Fix (toDex): Soot generates virtual invoke in case of interface method

MarcMil:
Fix for toDex
Soot generates virtual invoke in case of interface method when the interface provides a default method

### DIFF
--- a/src/main/java/soot/toDex/ExprVisitor.java
+++ b/src/main/java/soot/toDex/ExprVisitor.java
@@ -172,7 +172,8 @@ public class ExprVisitor implements ExprSwitch {
             = Scene.v().getActiveHierarchy().getSuperinterfacesOfIncluding(sie.getMethodRef().getDeclaringClass());
         NumberedString subsig = sie.getMethodRef().getSubSignature();
         for (SootClass i : allInterfaces) {
-          if (i.getMethodUnsafe(subsig) != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/soot/toDex/ExprVisitor.java
```diff
@@ -165,14 +166,23 @@ public void caseSpecialInvokeExpr(SpecialInvokeExpr sie) {
...
-      // In Dex files version 037 or later, if the method_id refers to an interface method, invoke-super is used to invoke
-      // the most specific, non-overridden version of that method defined on that interface.
-      stmtV.addInsn(buildInvokeInsn("INVOKE_SUPER", method, arguments), origStmt);
     } else {
+      if (sie.getMethodRef().getDeclaringClass().isInterface()) {
+        List<SootClass> allInterfaces
+            = Scene.v().getActiveHierarchy().getSuperinterfacesOfIncluding(sie.getMethodRef().getDeclaringClass());
+        NumberedString subsig = sie.getMethodRef().getSubSignature();
+        for (SootClass i : allInterfaces) {
+          if (i.getMethodUnsafe(subsig) != null) {
```
StevenArzt: Who tells you that this method has an implementation in the interface? It might just be a normal interface method.

PR #1977, opened 2023-06-15
Improving performance of Clinit-Trim Optimizer in Spark's Callgraph 2

piskachev:
This PR adds a cache mechanism in the ClinitElim Analysis. 

The calculated flow sets do not need to be re-calculated each time. This change keeps a static **cachedFlowSet** variable, which improves the scalability of this analysis. 

It was tested on a maven package [(link)](https://repo1.maven.org/maven2/org/apache/servicemix/bundles/org.apache.servicemix.bundles.gae/1.9.18_1/org.apache.servicemix.bundles.gae-1.9.18_1.jar) which takes >30 min to complete the SPARKS' VTA callgraph. After this change the time decreases to 17 minutes in the callgraph phase. The time numbers also include the loading.

### DIFF
--- a/src/main/java/soot/jimple/toolkits/callgraph/ClinitElimAnalysis.java
+++ b/src/main/java/soot/jimple/toolkits/callgraph/ClinitElimAnalysis.java
@@ -22,7 +22,9 @@ package soot.jimple.toolkits.callgraph;
  * #L%
  */
 
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Map;
 
 import soot.Scene;
 import soot.SootMethod;
@@ -36,6 +38,8 @@ public class ClinitElimAnalysis extends ForwardFlowAnalysis<Unit, FlowSet<SootMe
 
   private final CallGraph cg = Scene.v().getCallGraph();
   private final UnitGraph g;
+
+  private static final Map<SootMethod, Iterator<Edge>> outEdgesCache = new HashMap<>();
 
   public ClinitElimAnalysis(UnitGraph g) {
     super(g);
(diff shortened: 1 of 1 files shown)


src/main/java/soot/jimple/toolkits/callgraph/ClinitElimAnalysis.java
```diff
@@ -79,13 +81,22 @@ protected FlowSet<SootMethod> entryInitialFlow() {
 
   @Override
   protected FlowSet<SootMethod> newInitialFlow() {
-    HashSparseSet<SootMethod> set = new HashSparseSet<SootMethod>();
+    HashSparseSet<SootMethod> returnedFlowSet = new HashSparseSet<>();
```
linghuiluo, line 84: will this method be called multiple times? I thought it is only called once at the beginning.

piskachev: It seems it is called multiple times for the same node. 
We see this by checking the CPU time spent in this method. For the test subject, we see that Soot spends almost 9 % of the CPU time in this method. 
After adding this cache mechanism, the CPU time of this method has almost no impact ... less than 1 % 

afterPR
beforePR

piskachev: It seems it is called multiple times for the same node.
We see this by checking the CPU time spent in this method. For the test subject, we see that Soot spends almost 9 % of the CPU time in this method.
After adding this cache mechanism, the CPU time of this method has almost no impact ... less than 1 %

piskachev: afterPR
beforePR

PR #1998, opened 2023-09-06
Avoid duplicating constant assignments to final instance fields

subarnob:
### Problem
The `ConstantValueToInitializerTransformer` ensures for instance fields initialized with a default constant value, that constructor implementation(s) have the constant assignment to such instance fields immediately after the call to `super` constructor. For `final` instance fields with constant initializers, however, the constant assignment statement is already present in the generated Jimple body of the constructor(s), and furthermore its ordinal position within the body should not matter since they cannot be overwritten. Since the current implementation does not check for this case, a duplicate copy of such constant assignment statements are introduced.
This behavior can affect downstream analyses- e.g., we observe that Tags computed from a [liveness analysis](https://github.com/soot-oss/soot/blob/develop/src/main/java/soot/toolkits/scalar/LiveLocals.java) are lost when such constant assignment statements are incorrectly duplicated.

### Solution
We check for the presence of an existing assignment statement that initializes the given instance field with the given constant value (`isInstanceFieldAssignedConstantInBody()`), and avoid duplication if present.
* Note: we observe that these constants are already initialized as expected in the bytecode generated by modern Java compilers and do not know under which circumstances this transformation would be necessary. So, we simply check for the above condition to avoid duplication.

### Testing
* `mvn clean verify`
* other internal tests

### DIFF
--- a/src/main/java/soot/toolkits/scalar/ConstantValueToInitializerTransformer.java
+++ b/src/main/java/soot/toolkits/scalar/ConstantValueToInitializerTransformer.java
@@ -41,10 +41,12 @@ import soot.UnitPatchingChain;
 import soot.Value;
 import soot.ValueBox;
 import soot.VoidType;
+import soot.jimple.AssignStmt;
 import soot.jimple.Constant;
 import soot.jimple.DoubleConstant;
 import soot.jimple.FieldRef;
 import soot.jimple.FloatConstant;
+import soot.jimple.InstanceFieldRef;
 import soot.jimple.IntConstant;
 import soot.jimple.InvokeExpr;
 import soot.jimple.Jimple;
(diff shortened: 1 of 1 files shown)


src/main/java/soot/toolkits/scalar/ConstantValueToInitializerTransformer.java
```diff
@@ -173,6 +180,22 @@ public void transformClass(SootClass sc) {
     }
   }
 
+  private boolean isInstanceFieldAssignedConstantInBody(SootField sf, Constant constant, Body body) {
+    for (Unit u : body.getUnits()) {
```
linghuiluo, line 184: wouldn't it be better to cache the results for the field and constant rather than go through every time this loop?

subarnob: If `SootClass.getFields()` and `SootClass.getMethods()` return distinct member fields and methods respectively, then this check is performed once per distinct pairs of instance field (with a constant initializer) and (constructor) method. 

Caching could then benefit only for multiple copies of the same `ConstantValueTag` in `getTags()`. Could that be the case?

subarnob: Further, a field may be initialized to (at most) one constant. Note that a constant initialization in a super-class that is overridden in a sub-class will be handled separately- once for constructor(s) of super-class, again for constructor(s) of sub-class. So, the only benefit of caching here could be due to duplicate tags for the same constant value?

If so, handling/fixing that may be considered outside the scope of this PR?

StevenArzt: Sounds reasonable to me.

PR #2015, opened 2023-11-06
added reset methods to Scene and SourceLocator to load multiple sources into one Scene and Concrete Native Methods

lukaswd:
These two methods allow for subsequent calls of `loadNecessaryClasses()`.

With that it is possible, if the two methods are called before the next `loadNecessaryClasses()` method call, to load multiple sources into the same Soot Scene. 

Example code to load two programs into the same Soot Scene:
```java
// Options setting for the first source here

// Options setting for the second source here

SourceLocator.v().resetCaches();
Scene.v().resetSootClassPathCache();
Scene.v().setResolving(false);

```


src/main/java/soot/SootMethod.java
```diff
@@ -230,7 +230,7 @@ public boolean isPhantom() {
    * Returns true if this method is not phantom, abstract or native, i.e. this method can have a body.
    */
   public boolean isConcrete() {
-    return !isPhantom() && !isAbstract() && !isNative();
+    return !isPhantom() && !isAbstract();
```
StevenArzt: We should only use this semantics if the native code module is enabled. Maybe we can add a flag to the Options to enable or disable native code and check this flag here. Keep in mind that the options are generated from XML files via XSLT, so don't modify the Java file directly.

@MarcMil What do you think?

StevenArzt (changes requested):

src/main/java/soot/SootMethod.java
MarcMil: Sounds like a good idea, this implementation right here would brick things.

I wonder whether it would be better to do something like ```hasActiveBody() || ms != null```, so if it is **possible** to retrieve a body at all, which is what every caller seems to actually want to know.

StevenArzt: That would be an option, but it would also force us to refactor the Jimple Native reader to make it a proper ``MethodSource``.

StevenArzt (approved):

StevenArzt:
@lukaswd The code style check failed. Please have a look and fix the code accordingly.

PR #2022, opened 2023-11-17
refactor: changing sootclass getname to getpathplusclassname

jkraft-pjr:
In the SootClass class, the function getName() is not specific enough regarding what it is returning. I refactored the function name to getPathPlusClassName() as it is a better description of what the function is returning. 

The motivation behind making this change is in certain areas Soot assumes getName() is returning the fully qualified name of the class, package name + the class name. However, in some applications that is not always the case. For instance, in a SpringBoot application, all the application classes are put into the directories `BOOT-INF/classes`. It is important to note, the `BOOT-INF/classes` directories are not part of the class's package name. Which causes the path + class name to not equal the class's fully qualified name.  

This boils down to the file path of the class does not always equal the class's package name. According to the JDK 8 documentation, it states, "Many implementations of the Java platform rely on hierarchical file systems to manage source and class files, although The Java Language Specification does not require this."  

This pull request is merely to refactor the SootClass function, getName(), to describe the information it is returning accurately.

### DIFF
--- a/src/main/generated/jastadd/soot/JastAddJ/ClassDecl.java
+++ b/src/main/generated/jastadd/soot/JastAddJ/ClassDecl.java
@@ -419,7 +419,7 @@ public class ClassDecl extends ReferenceType implements Cloneable {
     }
     for(Iterator iter = interfacesIterator(); iter.hasNext(); ) {
       TypeDecl typeDecl = (TypeDecl)iter.next();
-      if(!sc.implementsInterface(typeDecl.getSootClassDecl().getName()))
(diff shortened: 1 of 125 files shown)


MarcMil:
That is not a good idea for multiple reasons. 
First, changing the name of such a important method would break **all** exsting applications using Soot as a library. 
Second, the fact that Soot uses the a name ```BOOT-INF.classes.com.example.springboot.Application``` is incorrect in the first place. That's the underlying issue that needs to be fixed.

Issue #2024, opened 2023-11-20
Worker thread execution failed: Failed to apply jb to <com.google.firebase.snippets.FirebaseAuthSnippets: javax.ws.rs.core.Response clearSessionCookieAndRevoke(javax.ws.rs.core.Cookie)>


buriedpot:
Hi, when I use 4.2.1, this error never occurs or occurs few times. 

Recently, I updated soot from 4.2.1 to 4.3.0, and then many programs throw Exceptions as below:  

<pre><code>
[ERROR] Worker thread execution failed: Failed to apply jb to <com.google.firebase.snippets.FirebaseAuthSnippets: javax.ws.rs.core.Response clearSessionCookieAndRevoke(javax.ws.rs.core.Cookie)>
java.lang.RuntimeException: Failed to apply jb to <com.google.firebase.snippets.FirebaseAuthSnippets: javax.ws.rs.core.Response clearSessionCookieAndRevoke(javax.ws.rs.core.Cookie)>
    at soot.asm.AsmMethodSource.getBody (AsmMethodSource.java:2282)
    at soot.SootMethod.retrieveActiveBody (SootMethod.java:446)
    at soot.PackManager.lambda$retrieveAllBodies$2 (PackManager.java:1252)
    at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:624)
    at java.lang.Thread.run (Thread.java:748)
Caused by: java.util.ConcurrentModificationException
    at java.util.ArrayList$Itr.checkForComodification (ArrayList.java:909)
    at java.util.ArrayList$Itr.next (ArrayList.java:859)
    at soot.FastHierarchy.getSignaturePolymorphicMethod (FastHierarchy.java:932)
    at soot.FastHierarchy.resolveMethod (FastHierarchy.java:833)
    at soot.FastHierarchy.resolveMethod (FastHierarchy.java:785)
    at soot.SootMethodRefImpl.tryResolve (SootMethodRefImpl.java:227)
    at soot.SootMethodRefImpl.tryResolve (SootMethodRefImpl.java:213)
    at soot.toolkits.exceptions.UnitThrowAnalysis.mightThrow (UnitThrowAnalysis.java:304)
[...]

PR #2043, opened 2024-01-17
fix ConcurrentModificationException when calling SootClass.getMethods…

buriedpot:
Issue #2024


StevenArzt:
Your merge request seems inconsistent with the issue #2024. The synchronization will only help in a multithreaded context, and only if the other operations that modify the collection are also synchronized. I don't see that this is the case. I rather assume that the overhead introduced by the synchronization fixes some weird timing issue by chance. If that's correct, the problem remains. Further, everyone pays for the overhead of synchronization. In total, I'm not really happy with this fix suggestion.

buriedpot:
Can you review my newest commit [fb376b8](https://github.com/soot-oss/soot/pull/2043/commits/fb376b86e3556b2b6a848f8d73ce7983fdc36679), thank you!

buriedpot:
> Your merge request seems inconsistent with the issue #2024. The synchronization will only help in a multithreaded context, and only if the other operations that modify the collection are also synchronized. I don't see that this is the case. I rather assume that the overhead introduced by the synchronization fixes some weird timing issue by chance. If that's correct, the problem remains. Further, everyone pays for the overhead of synchronization. In total, I'm not really happy with this fix suggestion.



> Can you review my newest commit [fb376b8](https://github.com/soot-oss/soot/pull/2043/commits/fb376b86e3556b2b6a848f8d73ce7983fdc36679), thank you!

Can you review my newest commit [fb376b8](https://github.com/soot-oss/soot/pull/2043/commits/fb376b86e3556b2b6a848f8d73ce7983fdc36679), thank you!

StevenArzt:
The copy constructor internally just iterates over the incoming list. You can just have the ``ConcurrentModificationException`` inside the constructor if you're unlucky. A real fix would need to make the data structures concurrent, which is a major undertaking.

MarcMil:
I agree with @StevenArzt 
Furthermore, I have already replaced this with a concurrent structure in the meantime.

Issue #2045, opened 2024-02-04
Soot generates redundant statements


Momo-Not-Emo:
**Describe the bug**
Soot generates redundant Jimple statements while SootUp does not

**Input file**
https://github.com/soot-oss/SootUp/blob/develop/shared-test-resources/miniTestSuite/java6/binary/Reflection.class
https://github.com/soot-oss/SootUp/blob/develop/shared-test-resources/miniTestSuite/java6/source/Reflection.java

**To reproduce**
soot version (commit id): 0e7e27d457e036ade112d91e1d97cf6a764eb0a9

Copy the input file to test resource directory, and run the following code

```java
    public void loadClasses(String first, String... more) {
        G.reset();
        Path cp = Paths.get(first, more);
        G.reset();
        Options.v().set_prepend_classpath(true);
        Options.v().set_process_dir(Collections.singletonList(cp.toFile().getAbsolutePath()));
        Options.v().set_src_prec(Options.src_prec_class);
        Options.v().set_output_format(Options.output_format_jimple);
        Options.v().set_allow_phantom_refs(true);
        Options.v().set_ignore_resolving_levels(true);
        Scene.v().loadNecessaryClasses();
    }

    public static void main(String[] args) {
        loadClasses(RESOURCE_PATH);
        SootMethod method = Scene.v().getMethod("<Autoboxing: void autoboxing()>");
        Body body = method.retrieveActiveBody();
        List<String> actualStmts = bodyStmtsAsStrings(body);

        for (String item : actualStmts) {
            System.out.println(item);
        }
        System.out.println();
}
```

Its output is as below, where the assignments of `r1`, `r11`, `r10` are redundant ([only one assignment is generated in SootUp](https://github.com/soot-oss/SootUp/blob/develop/sootup.jimple.parser/src/test/java/sootup/jimple/parser/javatestsuite/java6/ReflectionTest.java#L30))

```
r9 := @this: Reflection
$r0 = new Reflection
specialinvoke $r0.<Reflection: void <init>()>()
r1 = class "LReflection;"
r11 = class "LReflection;"
r10 = class "LReflection;"
$r2 = <java.lang.System: java.io.PrintStream out>
virtualinvoke $r2.<java.io.PrintStream: void println(java.lang.Object)>(class "LReflection;")
$r3 = newarray (java.lang.Class)[0]
[...]

Issue #472, opened 2015-09-12
RuntimeException: Failed to apply jb when analyzing APK files


dengfengli:
I try to use FlowDroid to analyze Android malware in Bgserv family and flowDroid prints the following exception trace:

Exception in thread "main" java.lang.RuntimeException: Failed to apply jb to <com.android.internal.telephony.gsm.SmsMessage: com.android.internal.telephony.gsm.SmsMessage$SubmitPdu getSubmitPdu(java.lang.String,java.lang.String,java.lang.String,boolean,byte[])>
    at soot.asm.AsmMethodSource.getBody(AsmMethodSource.java:1804)
    at soot.SootMethod.getBodyFromMethodSource(SootMethod.java:91)
    at soot.SootMethod.retrieveActiveBody(SootMethod.java:322)
    at soot.jimple.toolkits.callgraph.OnFlyCallGraphBuilder.processNewMethod(OnFlyCallGraphBuilder.java:529)
    at soot.jimple.toolkits.callgraph.OnFlyCallGraphBuilder.processReachables(OnFlyCallGraphBuilder.java:420)
    at soot.jimple.spark.solver.OnFlyCallGraph.build(OnFlyCallGraph.java:55)
    at soot.jimple.spark.solver.PropWorklist.handleVarNode(PropWorklist.java:122)
    at soot.jimple.spark.solver.PropWorklist.propagate(PropWorklist.java:52)
    at soot.jimple.spark.SparkTransformer.internalTransform(SparkTransformer.java:152)
    at soot.SceneTransformer.transform(SceneTransformer.java:39)
    at soot.Transform.apply(Transform.java:90)
    at soot.RadioScenePack.internalApply(RadioScenePack.java:57)
    at soot.jimple.toolkits.callgraph.CallGraphPack.internalApply(CallGraphPack.java:49)
    at soot.Pack.apply(Pack.java:121)
    at soot.jimple.infoflow.android.SetupApplication.calculateCallbackMethods(SetupApplication.java:448)
    at soot.jimple.infoflow.android.SetupApplication.calculateSourcesSinksEntrypoints(SetupApplication.java:359)
    at soot.jimple.infoflow.android.SetupApplication.calculateSourcesSinksEntrypoints(SetupApplication.java:325)
    at soot.jimple.infoflow.android.TestApps.Test.runAnalysis(Test.java:521)
    at soot.jimple.infoflow.android.TestApps.Test.main(Test.java:206)
[...]

PR #2046, opened 2024-02-07
Fix always-false condition in <AsmMethodSource: void assignReadOps(Local l)>

Momo-Not-Emo:
### Problem Description:
The always-false condition was initially reported in [Soot PR #1834](https://github.com/soot-oss/soot/pull/1834), resulting in a branch that is never reached or covered. This issue may lead to redundant Jimple statements, as addressed by the fix in [SootUp PR #472](https://github.com/soot-oss/SootUp/pull/472) in SootUp. Unfortunately, Soot PR #1834 was closed without being merged due to the absence of test cases.

### Solution Overview:
I have ported the test cases from [SootUp PR #472](https://github.com/soot-oss/SootUp/pull/472) to Soot and observed that the current Soot implementation does not generate redundant statements despite retaining the always-false condition.

### Changes Made:
I corrected the always-false condition by referencing both Soot PR #1834 and [SootUp PR #472](https://github.com/soot-oss/SootUp/pull/472). Following these modifications, all tests, including the ported ones, pass successfully when executing `mvn -B clean test -PJava8`, `mvn -B clean test -PJava9`, and `mvn -B clean test -PJava11`. However, one ported test consistently produces redundant Jimple statements, as reported in [Issue #2045](https://github.com/soot-oss/soot/issues/2045).

### Further Insights:
The results suggest that the current Soot implementation may not require the if-structure, as it has no impact on the test cases, whether the condition is reachable or not. This conclusion is supported by tracking the line coverage in Jacoco reports before and after the modification.


StevenArzt:
Thank you for this effort, much appreciated. Could you take a look at the failed style check in our build pipeline and fix that?

PR #2048, opened 2024-02-08
New option: Ignore dex overflows

beerphilipp:
This pull request addresses the issue where Soot fails to build large Android applications with a `minSdkVersion` < 22 due to the absence of multi-dex support for API versions < 22, resulting in Soot throwing a `RuntimeException`. Given that Android API versions >= 22 come with built-in multi-dex support, these applications would theoretically work on devices running newer Android versions.

To allow Soot users to build such APKs, this pull request introduces an option to explicitly ignore dex overflows. The option can be enabled by calling `Options.v().set_ignore_dex_overflow(true)`.

An APK that results in such a `RuntimeException` is attached below. 
[Uploading com.mathgamesplayground.zip…]()

### DIFF
--- a/src/main/generated/options/soot/options/Options.java
+++ b/src/main/generated/options/soot/options/Options.java
@@ -1443,6 +1443,10 @@ public class Options extends OptionsBase {
                     || option.equals("no-writeout-body-releasing")
             )
                 no_writeout_body_releasing = true;
+            else if (false
+                    || option.equals("ignore-dex-overflow")
+            )
+                ignore_dex_overflow = true;
             else {
                 G.v().out.println("Invalid option -" + option);
                 return false;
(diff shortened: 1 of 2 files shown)


StevenArzt:
Soot already checks whether the target APK targets the ART engine instead of Dalvik. In other words, if your target SDK version is large enough, this exception should not be thrown. I'm not sure that producing an APK with a low target SDK version but a requirement on ART's multidex support is a good idea.

Additionally, the options Java file is generated from the options XML. Any code changes will be overwritten. If you want to introduce a new option value, you need to put it into the XML.

MarcMil:
Given that devices having the legacy Dalvik VM continue to loose relevance I wouldn't be so worried. We could log out a warning the the app might not work on older devices. The only downside is that we might get some bug reports from users enabling that option, ignoring the warning and running the (probably old) app on an old device.
So, @StevenArzt, what should we do here? I could reimplement the code to use the options xml file instead, or we close this PR without any changes to Soot.

StevenArzt:
@MarcMil I suggest that we log a warning instead of throwing the exception. There is no need for an option. The warning tells people that they are in trouble on prehistoric Android phones.

MarcMil:
Can be closed, we have a warning now @StevenArzt

PR #2060, opened 2024-03-26
Fix the time consume problem in method `instructionAtAddress`

wangnianwu:
If the address is not continuous, it will visit all integers from the current to target address(The length exceeds the map size), it causes taking too much time to call `retrieveActiveBody`  in SootMethod.


jpstotz:
As you have changed the Map type from HashMap to TreeMap wouldn't it be easier to use the TreeMap specific functions like `floorKey` or `ceilingKey` to get the nearest instruction to an address instead of iterating through the whole TreeMap to find the address?

wangnianwu:
@jpstotz updated it by your advise

Issue #1881, opened 2022-06-29
Upgrade soot code to 4.3.0 with JDK > 11


magg:
I use soot on my java project, but I currently need to upgrade my project from Java 8 to 11 or 17, so I would like to know if you can guide me on how to do it, I seem to be hitting lots of errors.

Here's the old code working with java 8 and soot 3.2.1, it was using java 7 in soot.

```
  G.reset();
            Scene.v().releaseSideEffectAnalysis();

            //define classpath with ConfigClass and java rt.jar
            final String classpath = SootGeneratorServiceImpl.class
                    .getProtectionDomain()
                    .getCodeSource()
                    .getLocation()
                    .getPath()
                    + File.pathSeparator
                    + Paths.get(System.getProperty("java.home")).resolve("lib").resolve("rt.jar");
            Options.v().set_soot_classpath(classpath);
            Options.v().set_java_version(Options.java_version_1_7);
            //define resolve level for some classes we are interested in
            Scene.v().addBasicClass(ILLEGAL_ARGUMENT_EXCEPTION, SootClass.SIGNATURES);
            Scene.v().addBasicClass(CONFIG_CLASS, SootClass.BODIES);
            Scene.v().addBasicClass(CLASS, SootClass.BODIES);
            Scene.v().addBasicClass(EXCEPTION, SootClass.SIGNATURES);
            Scene.v().addBasicClass(STRING, SootClass.SIGNATURES);
            Scene.v().addBasicClass(CONSTRUCTOR, SootClass.BODIES);
            Scene.v().loadNecessaryClasses();

            configClassTemplate = Scene.v().loadClassAndSupport(CONFIG_CLASS);

            //retrieve active bodies for ConfigClass
            for (SootMethod method : configClassTemplate.getMethods()) {
                if (!method.isConcrete()) {
                    continue;
                }
                if (!method.hasActiveBody()) {
                    method.retrieveActiveBody();
                }
            }
```

What can I do about rt.jar should I download it manually? for JDK 11 and 17?

Apparently there is this line for java 11 but not for java 17

` Options.v().set_java_version(Options.java_version_11);`

If i remove the resolve rt.jar call

I get 
```

            couldn't find class: java.lang.Object (is your soot-class-path set properly?) Try adding rt.jar to Soot's classpath, e.g.:
            java -cp sootclasses.jar soot.Main -cp .:/path/to/jdk/jre/lib/rt.jar <other options>

Caused by: java.lang.AssertionError
[...]

PR #2077, opened 2024-04-30
IOException happens during writing to jar

Ilya-Skiba:
Rolled back try-with resourses block, replaced with simple write https://github.com/soot-oss/soot/issues/1881

### DIFF
--- a/src/main/java/soot/toDex/DexPrinter.java
+++ b/src/main/java/soot/toDex/DexPrinter.java
@@ -10,12 +10,12 @@ package soot.toDex;
  * it under the terms of the GNU Lesser General Public License as
  * published by the Free Software Foundation, either version 2.1 of the
  * License, or (at your option) any later version.
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Lesser Public License for more details.
- * 
+ *
  * You should have received a copy of the GNU General Lesser Public
  * License along with this program.  If not, see
  * <http://www.gnu.org/licenses/lgpl-2.1.html>.
@@ -313,7 +313,7 @@ public class DexPrinter {
       }
 
       if (Options.v().output_jar()) {
-        // if we create JAR file, MANIFEST.MF is preferred
+        // if we create JAR file, MANIFEST.MF is preferred.
         addManifest(outputZip, files);
       }
 
@@ -404,8 +404,8 @@ public class DexPrinter {
     try (BufferedOutputStream bufOut = new BufferedOutputStream(destination)) {
       manifest.write(bufOut);
       bufOut.flush();
+      destination.closeEntry();
     }
-    destination.closeEntry();
   }
 
   /**
@@ -1116,7 +1116,7 @@ public class DexPrinter {
 
   /**
    * Checks whether the given method shall be ignored, i.e., not written out to dex
-   * 
+   *
    * @param sm
    *          The method to check
    * @return True to ignore the method while writing the dex file, false to write it out as normal
@@ -1127,7 +1127,7 @@ public class DexPrinter {
 
   /**
    * Checks whether the given field shall be ignored, i.e., not written out to dex
-   * 
+   *
    * @param sf
    *          The field to check
    * @return True to ignore the field while writing the dex file, false to write it out as normal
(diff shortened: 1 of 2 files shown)


src/main/java/soot/toDex/DexPrinter.java
```diff
@@ -401,10 +401,7 @@ private void addManifest(ZipOutputStream destination, Collection<File> dexFiles)
 
     final ZipEntry manifestEntry = new ZipEntry(JarFile.MANIFEST_NAME);
     destination.putNextEntry(manifestEntry);
-    try (BufferedOutputStream bufOut = new BufferedOutputStream(destination)) {
-      manifest.write(bufOut);
-      bufOut.flush();
-    }
+    manifest.write(new BufferedOutputStream(destination));
```
jpstotz: Are you sure that without try-with-resource and without flush the manifest data are always fully written to `destination`? In my understanding there is a chance that manifest data is only written to the buffer inside `BufferedOutputStream` but never (completely) flushed to `destination`.

Ilya-Skiba: Yes, I'm sure. This flush happens during `closeEntry` invocation. Current implementation causes BufferedOutputStream to close, which forces ZipOutputStream to close. As a result, an exception occurs

jpstotz: How should the `BufferedOutputStream` be flushed in `closeEntry`? The reference to `BufferedOutputStream` is not known to `closeEntry` also destination does not know anything about the `BufferedOutputStream` instance, so how should `BufferedOutputStream` be flushed?

Ilya-Skiba: I've added a test for this case. In general, closing manifest entry forces all input buffers to flush. This why it works correctly

StevenArzt: I have to side with @jpstotz here. Calling ``closeEntry`` on a ``ZipOutputStream`` handles the re-positioning and the CRC calculation, but doesn't close any streams. The JavaDoc does not say that it closes streams and the code doesn't do it.

In my opinion, the call to ``closeEntry`` needs to be moved inside the ``try`` block. You flush, you call ``closeEntry``, then the ``try`` block ends and the stream gets closed. With this approach, we don't get dangling streams and ``closeEntry`` should be happy.

Ilya-Skiba: According to [closable](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/FilterOutputStream.html#close()), try-with-resources will close underlying ZipOutputStream, that is not correct for this method- ZipOutputStream might be used by other methods also.  What do you think about removing BufferedOutputStream? I don't think there would be any issues with performance during small manifest file saving

StevenArzt: You're right, closing the ``BufferedOutputStream`` closes the inner stream as well.

The ``addManifest`` method should only be called by ``printZip`` and should only be called once there in line 317. In the stack trace in issue #1881, the line number is different, which confuses me slightly, but let's ignore that for now.

The ``printZip`` method owns the ``ZipOutputStream`` and the call to ``addManifest`` in line 317 is the last operation that does anything with the ZIP file. So no matter whether we close the inner stream in ``addManifest`` or not, noone should need that inner stream afterwards anyway. In fact, when ``printZip`` returns, it closes the ``ZipOutputStream`` anyway in its own ``try`` block (which I had overlooked in my previous post, so we don't get a dangling stream, even though ``closeEntry`` has nothing to do with it).

So whether we close the stream in ``adManifest`` or not, it shouldn't matter. However, this makes me wonder why changing ``addManifest`` has any effect on this exception, i.e., whether we're discussing something that is irrelevant to the problem at hand. If the entire ZIP file is closed after the call to ``addManifest``, how can we later get an exception in ``addManifest``? How can the control flow ever return there with the same ``ZipOutputStream``? The ``ZipOutputStream``  should never be recycled anyway?

Ilya-Skiba: My code was based on the principle, that a stream should not be closed in a method, that did not open it. I've updated the code and returned try-with-resources

StevenArzt: That totally makes sense. I am now just confused why the exception from the issue report happens at all.

Ilya-Skiba: `destination.closeEntry()` failed before, because was trying to write entry on closed stream. You can rollback my changes in DexPrinter and run `testWritingToJar` test to reproduce

PR #2081, opened 2024-05-15
Compute topological sort using post-order traversal

### LINKED ISSUE #2080: Incorrect topological sorting leads to incorrect SCC computation in Spark
Please examine each of the following points so that we can help you as soon and best as possible.

**Describe the bug**

Spark’s SCC-collapsing optimization merges PAG nodes which should be in separate components.

**Input file**

```java
package soot.jimple.spark.solver;

import java.util.Collections;

import org.junit.Test;

import static org.junit.Assert.assertEquals;

import soot.Scene;
import soot.Type;
import soot.jimple.spark.pag.PAG;
import soot.jimple.spark.pag.VarNode;
import soot.options.SparkOptions;

public class SCCCollapserTest {

    @Test
    public void testSeparateComponents() {
        Scene.v().loadBasicClasses();
        Type type = Scene.v().getObjectType();

        SparkOptions sparkOptions = new SparkOptions(Collections.emptyMap());
        PAG pag = new PAG(sparkOptions);

        VarNode a = pag.makeGlobalVarNode("a", type);
        VarNode b = pag.makeGlobalVarNode("b", type);
        VarNode c = pag.makeGlobalVarNode("c", type);
        pag.addEdge(a, b);
        pag.addEdge(a, c);
        pag.addEdge(b, c);

        SCCCollapser sccCollapser = new SCCCollapser(pag, false);
        sccCollapser.collapse();
        pag.cleanUpMerges();

        assertEquals(a, a.getReplacement());
        assertEquals(b, b.getReplacement());
        assertEquals(c, c.getReplacement());
    }
}
```

**To reproduce**

The second assertion fails in the test above.

**Expected behavior**

The test above should pass, since nodes a, b, and c should each be in their own component.

**Stacktrace**

N/A

**Additional context**

It appears that the SCC computation, following Kosaraju’s algorithm, is based on an invalid topological order. The topological sort should order node b before c. Instead, c gets ordered before b. This results in c being the SCC-root for b.

michael-emmi:
* fixes #2080

### DIFF
--- a/src/main/java/soot/jimple/spark/solver/TopoSorter.java
+++ b/src/main/java/soot/jimple/spark/solver/TopoSorter.java
@@ -22,9 +22,9 @@ package soot.jimple.spark.solver;
  * #L%
  */
 
-import java.util.ArrayList;
 import java.util.HashSet;
-import java.util.List;
+import java.util.Set;
+import java.util.Stack;
 
 import soot.jimple.spark.pag.Node;
 import soot.jimple.spark.pag.PAG;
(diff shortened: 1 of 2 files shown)


michael-emmi:
FYI @StevenArzt @MarcMil

PR #2084, opened 2024-05-27
Fix array type detection on fill-array-data DEX instructions - fixes #1806

### LINKED ISSUE #1806: java.lang.RuntimeException: Invalid Array Type occured in FillArrayDataInstruction: java.lang.Object
Used version: branch `develop` 34f871e4e24ca4dd18b4a1bff701c9eff56f2f00

**To reproduce**
Try to retrieve the body of `<X.D8D: X.1n9 AGs(X.1oJ,int,int)>`

**Stacktrace**
```
java.lang.RuntimeException: Invalid Array Type occured in FillArrayDataInstruction: com.facebook.flexlayout.styles.FlexItemCallback
	at soot.dexpler.instructions.FillArrayDataInstruction.getArrayElement(FillArrayDataInstruction.java:195)
	at soot.dexpler.instructions.FillArrayDataInstruction.jimplify(FillArrayDataInstruction.java:101)
	at soot.dexpler.DexBody.jimplify(DexBody.java:634)
	at soot.dexpler.DexMethod$1.getBody(DexMethod.java:120)

```

[com.facebook.orca_337.1.0.11.118 (297415372) - stripped.apk.zip](https://github.com/soot-oss/soot/files/7642259/com.facebook.orca_337.1.0.11.118.297415372.-.stripped.apk.zip)

jpstotz:
The previous `FillArrayDataInstruction` which converts DEX instructions to Jimpl instructions had two major problems:

1. The array type detection was only recognizing new-array instructions, not array that are created as a return value of a function
2. For detecting the array type it used a simple algorithm that inspected the DEX instructions before the fill-array-data instruction, not considering (conditional) branches of the control flow. This caused two problems: 
3. If the array type could not be detected a warning was logged and the fill-array-data instruction completely ignored
4. If multiple array were defined in a method with non-linear control flow the wrong array-new instruction could be assigned to the fill-array-data instruction causing various problems in the command itself or in a later phase.

The `FillArrayDataInstruction` implementation provided by this PR splits processing of fill-array-data instructions into two phases:

1. Command transforming to Jimple instructions, as the array type is not known the elements are stored as `UntypedConstant` 
2. In the second phase the new `DexFillArrayDataTransformer` checks AssignStatements that assign an `UntypedConstant` to an array element. It recovers the array data types and applies it to each value.

Note: The time to execute `DexFillArrayDataTransformer.v().transform(jBody);` was chosen as it is the latest possible point in time to execute it. The next transformer `TypeAssigner.v().transform(jBody);` can not handle `UntypedConstant` and thus throws an exception if it encounters one. 

In a test the new implementation was used to process ~350 recent Android apps.

### DIFF
--- a/src/main/java/soot/dexpler/DexFillArrayDataTransformer.java
+++ b/src/main/java/soot/dexpler/DexFillArrayDataTransformer.java
@@ -91,30 +91,8 @@ public class DexFillArrayDataTransformer extends BodyTransformer {
           ArrayRef leftArray = (ArrayRef) left;
 
           Local l = (Local) leftArray.getBase();
-          List<Unit> assDefs = defs.getDefsOfAt(l, ass);
           List<Type> arrayTypes = new LinkedList<>();
-          for (Unit d : assDefs) {
-            if (d instanceof AssignStmt) {
-              AssignStmt arrayAssign = (AssignStmt) d;
-              Value source = arrayAssign.getRightOp();
-              if (source instanceof NewArrayExpr) {
-                NewArrayExpr newArray = (NewArrayExpr) source;
-                arrayTypes.add(newArray.getBaseType());
-                continue;
-              }
-              if (source instanceof InvokeExpr) {
-                InvokeExpr invExpr = (InvokeExpr) source;
-                Type aType = invExpr.getMethodRef().getReturnType();
-                if (!(aType instanceof ArrayType)) {
-                  throw new InternalError("Failed to identify the array type. The identified method invocation "
-                      + "does not return an array type. Invocation: " + invExpr.getMethodRef());
-                }
-                arrayTypes.add(((ArrayType) aType).getArrayElementType());
-                continue;
-              }
-              throw new InternalError("Unsupported array definition statement: " + d);
-            }
-          }
+          checkArrayDefinitions(l, ass, defs, arrayTypes);
           if (arrayTypes.isEmpty()) {
             throw new InternalError("Failed to determine the array type ");
           }
@@ -135,4 +113,50 @@ public class DexFillArrayDataTransformer extends BodyTransformer {
     }
   }
 
+  /**
+   * Check the all available definitions of the current array to detect the array type and thus the type of the data loaded
+   * by the array-fill-data instruction.
+   * 
+   * @param l
+   *          local the array we are interested in is saved in
+   * @param u
+   *          unit we start our search
+   * @param defs
+   * @param arrayTypes
+   *          result list containing the discovered array type(s)
+   */
+  private void checkArrayDefinitions(Local l, Unit u, LocalDefs defs, List<Type> arrayTypes) {
+    List<Unit> assDefs = defs.getDefsOfAt(l, u);
+    for (Unit d : assDefs) {
+      if (d instanceof AssignStmt) {
+        AssignStmt arrayAssign = (AssignStmt) d;
+        Value source = arrayAssign.getRightOp();
+        if (source instanceof NewArrayExpr) {
+          // array is assigned from a newly created array
+          NewArrayExpr newArray = (NewArrayExpr) source;
+          arrayTypes.add(newArray.getBaseType());
+          continue;
+        }
+        if (source instanceof InvokeExpr) {
+          // array is assigned from the return value of a function
+          InvokeExpr invExpr = (InvokeExpr) source;
+          Type aType = invExpr.getMethodRef().getReturnType();
+          if (!(aType instanceof ArrayType)) {
+            throw new InternalError("Failed to identify the array type. The identified method invocation "
+                + "does not return an array type. Invocation: " + invExpr.getMethodRef());
+          }
+          arrayTypes.add(((ArrayType) aType).getArrayElementType());
+          continue;
+        }
+        if (source instanceof Local) {
+          // our array is defined by an assignment from another array => check the definition of that other array.
+          Local newLocal = (Local) source; // local of the "other array"
+          checkArrayDefinitions(newLocal, d, defs, arrayTypes);
+          continue;
+        }
+        throw new InternalError("Unsupported array definition statement: " + d);
+      }
+    }
+
+  }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/soot/dexpler/DexFillArrayDataTransformer.java
```diff
@@ -0,0 +1,138 @@
...
+      if (rightOp instanceof UntypedConstant) {
+        Value left = ass.getLeftOp();
+        if (left instanceof ArrayRef) {
+          ArrayRef leftArray = (ArrayRef) left;
+
+          Local l = (Local) leftArray.getBase();
+          List<Unit> assDefs = defs.getDefsOfAt(l, ass);
+          List<Type> arrayTypes = new LinkedList<>();
+          for (Unit d : assDefs) {
+            if (d instanceof AssignStmt) {
```
StevenArzt: This isn't fully generic either.You might have code like this:

```
b = new Foo[];
a = b;
```

In that case, this code wouldn't be able to find the type of variable ``a``.

src/main/java/soot/dexpler/DexFillArrayDataTransformer.java
```diff
@@ -0,0 +1,138 @@
...
+              throw new InternalError("Unsupported array definition statement: " + d);
+            }
+          }
+          if (arrayTypes.isEmpty()) {
+            throw new InternalError("Failed to determine the array type ");
+          }
+          if (arrayTypes.size() > 1) {
+            arrayTypes = arrayTypes.stream().distinct().collect(Collectors.toList());
+            if (arrayTypes.size() > 1) {
+              logger.warn("Found multiple possible array types, using first ignoreing the others: {}", arrayTypes);
```
StevenArzt, line 104: Wouldn't it make sense to find a least common ancestor (lcas) in hthe type hierarchy? Soot already uses this strategy in the type assigner and there is already code for it (look for ``lcas``).

StevenArzt (changes requested):

src/main/java/soot/dexpler/DexFillArrayDataTransformer.java
jpstotz: @StevenArzt Are you sure this is really necessary? As you noticed in #1806 fill-array-data only works for primitive types like byte, char, short, int, long, float, double. None of these types have a least common ancestor. Therefore if  `arrayTypes` would contain such a type no matter what we do we will end up in an exception.

StevenArzt: In #1806 I asked whether the instruction allows for objects. I have never seen that, but nobody has double-checked so far. If we can exclude this case, we're safe.

The only other concern would be widening if we have one use for ``int`` and one for ``long``, for example.

src/main/java/soot/dexpler/DexFillArrayDataTransformer.java
jpstotz: Indeed this case is possible. My updated PR is now capable of handling assignment from other Local. In such a case we are tracing back the array definition using recursion. See `checkArrayDefinitions` function.

src/main/java/soot/dexpler/DexFillArrayDataTransformer.java
jpstotz: When I was asking about objects I wasn't very familiar with the `fill-array-data` command. Theoretically it would be possible to use it to fill an object array, but the array can only held references to an object and not the object itself.
Thus it would be necessary to save the object references ("pointers") in the array data block. As this block is defined at compile-time you would have to know the object reference "pointers" at compile-time.

Also a quick practical test uncovered that the Android verifier rejects if you try to apply `fill-array-data` onto an `Object[]` array:

> Error from logcat: invalid fill-array-data with component type

StevenArzt: OK, great. That makes the Dexpler implementation a lot easier.

src/main/java/soot/dexpler/DexFillArrayDataTransformer.java
```diff
@@ -135,4 +113,50 @@ protected void internalTransform(final Body body, String phaseName, Map<String,
...
+      if (d instanceof AssignStmt) {
+        AssignStmt arrayAssign = (AssignStmt) d;
+        Value source = arrayAssign.getRightOp();
+        if (source instanceof NewArrayExpr) {
+          // array is assigned from a newly created array
+          NewArrayExpr newArray = (NewArrayExpr) source;
+          arrayTypes.add(newArray.getBaseType());
+          continue;
+        }
+        if (source instanceof InvokeExpr) {
```
StevenArzt: I'd use an ``else`` here to avoid duplicacte checks.

src/main/java/soot/dexpler/DexFillArrayDataTransformer.java
```diff
@@ -135,4 +113,50 @@ protected void internalTransform(final Body body, String phaseName, Map<String,
...
+            throw new InternalError("Failed to identify the array type. The identified method invocation "
+                + "does not return an array type. Invocation: " + invExpr.getMethodRef());
+          }
+          arrayTypes.add(((ArrayType) aType).getArrayElementType());
+          continue;
+        }
+        if (source instanceof Local) {
+          // our array is defined by an assignment from another array => check the definition of that other array.
+          Local newLocal = (Local) source; // local of the "other array"
+          checkArrayDefinitions(newLocal, d, defs, arrayTypes);
```
StevenArzt: We need a set of assignments that we have already seen to avoid infinite recursion. I am thinking about code like this:

```
if (...) {
a = b;
} else {
b = a;
}
```

StevenArzt (changes requested):

src/main/java/soot/dexpler/DexFillArrayDataTransformer.java
jpstotz: Done

src/main/java/soot/dexpler/DexFillArrayDataTransformer.java
jpstotz: recursion depth parameter introduced that decreases with every recursion and abort check added.

PR #2131, opened 2024-11-28
Fix bugs in dexpler and toDex

MarcMil:


src/main/java/soot/dexpler/DexBody.java
```diff
@@ -1307,7 +1307,15 @@ private static BooleanConstant fixBooleanConstant(IntConstant arg) {
    *          type constraints (these might be multiple valid possibilities)
    */
   private void handleKnownDexArrayTypes(Body b, Jimple jimple, MultiMap<Local, Type> typeConstraints) {
-
+    Set<Local> localsSingleDefinitions = new HashSet<>(b.getLocals());
+    for (Unit u : b.getUnits()) {
```
StevenArzt, line 1311: Why is this set not empty? This seems to look for uninitialized locals.

StevenArzt (changes requested):

StevenArzt (approved):

PR #2141, opened 2025-01-24
Fix the bug java.lang.IllegalArgumentException: Opcode: IPUT_OBJECT_VOLATILE

sfncat:
[PR] Fix the error java.lang.IllegalArgumentException: Opcode: IPUT_OBJECT_VOLATILE @ 0x0 that occurs when generating Jimple from an APK, which prevents successful Jimple generation.

[Solution] Added handling for the IPUT_OBJECT_VOLATILE opcode, treating it equivalently to the standard IPUT_OBJECT opcode by using the IputInstruction class for processing. This is because IPUT_OBJECT_VOLATILE is the volatile variant of IPUT_OBJECT, and their instruction-handling logic can be unified at the implementation level. Additionally, similar handling was added for the IGET_OBJECT_VOLATILE opcode.

[Test] After implementing these changes, the error no longer occurs, and Jimple code is generated successfully.

### DIFF
--- a/src/main/java/soot/SootClass.java
+++ b/src/main/java/soot/SootClass.java
@@ -1323,7 +1323,7 @@ public class SootClass extends AbstractHost {
    */
   public Collection<SootMethod> getMethodsByNameAndParamCount(String name, int paramCount) {
     List<SootMethod> result = new ArrayList<>();
-    // 创建副本以避免并发修改
+    // Create a copy to avoid concurrent modification
     List<SootMethod> methodsCopy = new ArrayList<>(this.getMethods());
     for (SootMethod m : methodsCopy) {
         if (m.getName().equals(name) && m.getParameterCount() == paramCount) {
--- a/src/main/java/soot/SootClass.java
+++ b/src/main/java/soot/SootClass.java
@@ -1322,18 +1322,13 @@ public class SootClass extends AbstractHost {
    * @return the methods
    */
   public Collection<SootMethod> getMethodsByNameAndParamCount(String name, int paramCount) {
-    List<SootMethod> result = null;
-    for (SootMethod m : getMethods()) {
-      if (m.getParameterCount() == paramCount && m.getName().equals(name)) {
-        if (result == null) {
-          result = new ArrayList<>();
+    List<SootMethod> result = new ArrayList<>();
+    // 创建副本以避免并发修改
+    List<SootMethod> methodsCopy = new ArrayList<>(this.getMethods());
+    for (SootMethod m : methodsCopy) {
+        if (m.getName().equals(name) && m.getParameterCount() == paramCount) {
+            result.add(m);
         }
-        result.add(m);
-      }
-    }
-
-    if (result == null) {
-      return Collections.emptyList();
     }
     return result;
   }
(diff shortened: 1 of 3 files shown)


sfncat:
Build and Test with java 11:OK
Build and Test with java 17:OK
Build and Test with java 21:OK
Check style´and license conformity:OK

src/main/java/soot/SootClass.java
```diff
@@ -1322,18 +1322,13 @@ public boolean isOpenedByModule() {
...
    */
   public Collection<SootMethod> getMethodsByNameAndParamCount(String name, int paramCount) {
-    List<SootMethod> result = null;
-    for (SootMethod m : getMethods()) {
-      if (m.getParameterCount() == paramCount && m.getName().equals(name)) {
-        if (result == null) {
-          result = new ArrayList<>();
+    List<SootMethod> result = new ArrayList<>();
+    // Create a copy to avoid concurrent modification
+    List<SootMethod> methodsCopy = new ArrayList<>(this.getMethods());
```
StevenArzt, line 1327: This isn't a proper fix. The copy constructor internally iterates over the list, which can trigger the same ``ConcurrentModificationException``. If you have a problem with concurrent modification, you need to fix the root cause.

src/main/java/soot/jimple/toolkits/scalar/CopyPropagator.java
```diff
@@ -218,7 +218,10 @@ protected void internalTransform(Body b, String phaseName, Map<String, String> o
                 if (l != m) {
                   Integer defCount = localToDefCount.get(m);
                   if (defCount == null || defCount == 0) {
-                    throw new RuntimeException("Variable " + m + " used without definition!");
+                    if (Options.v().verbose()) {
```
StevenArzt, line 221: Don't just ignore errors. If we generate invalid code, we need to know that. Keep in mind that Soot can also write out DEX code again. If we have wrong code, the output will be invalid as well. These sanity checks are highly important.

StevenArzt (changes requested):

src/main/java/soot/SootClass.java
sfncat: This problem occurs when calling soot to generate jimples from jars. The current scenario that will definitely occur is when processing the latest version of pixel9proxl's /system/framework/services.jar. This problem causes the jimple to fail to be generated. If this modification is not appropriate, I will raise an issue.

sfncat: This problem occurs when calling soot to generate jimples from jar. The current random scenario is when processing the latest version of pixel9proxl /system/framework/services.jar. This problem causes the jimple to fail to generate.

src/main/java/soot/jimple/toolkits/scalar/CopyPropagator.java
sfncat: OK.I will close the PR.

sfncat: I print the code of the error
Exception in thread "main" java.lang.RuntimeException: Undefined variable $u-1 in method: <com.android.server.policy.PhoneWindowManager$SwitchKeyboardLayoutMessageObject: int hashCode()>
Class: com.android.server.policy.PhoneWindowManager$SwitchKeyboardLayoutMessageObject

Method body:
   1: this := @this: com.android.server.policy.PhoneWindowManager$SwitchKeyboardLayoutMessageObject
   2: this.<android.apex.ApexInfo: boolean isFactory> = $u0
   3: $u0 = $u0
   4: $u0 = $u-1
   5: return $u0

PR #2214, opened 2025-11-24
Improve performance by introducing new caches for FastHierarchy

MarcMil:
Furthermore: Adds a field so that each Jimple statement knows which body the statement belongs to. This allows client analysis to use Statements everywhere instead of keeping track of the containing body.

### DIFF
--- a/src/main/java/soot/SootClass.java
+++ b/src/main/java/soot/SootClass.java
@@ -29,6 +29,8 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


src/main/java/soot/SootClass.java
```diff
@@ -682,10 +685,20 @@ public void addMethod(SootMethod m) {
...
     this.subSigToMethods.put(m.getNumberedSubSignature(), m);
     this.methodList.add(m);
+    addToNameAndParamMap(m);
     m.setDeclared(true);
     m.setDeclaringClass(this);
   }
 
+  protected void addToNameAndParamMap(SootMethod m) {
+    List<SootMethod> newList = new ArrayList<>();
+    List<SootMethod> use = methodParameterMap.putIfAbsent(new NameAndNumber(m.getName(), m.getParameterCount()), newList);
```
StevenArzt: Slight improvement: ``computeIfAbsent``

StevenArzt (changes requested):

PR #2220, opened 2025-12-01
SootMethodRefImpl: Fix static method resolving

MarcMil:


src/main/java/soot/FastHierarchy.java
```diff
@@ -969,7 +999,7 @@ protected SootMethod resolveMethod(final SootClass baseType, final SootClass dec
 
           SootMethod method = getSignaturePolymorphicMethod(iFace, name, parameterTypes, returnType);
           if (method != null && isVisible(declaringClass, iFace, method.getModifiers())) {
-            if ((!allowAbstract && method.isAbstract()) || method.isStatic()) {
+            if ((!allowAbstract && method.isAbstract()) || (allowStatic || method.isStatic())) {
```
StevenArzt: ``allowStatic`` rather seems to be ``forceStatic'' in the way that it is used here. Was that intended? If so, we should rename the parameter.

StevenArzt (changes requested):
See my comments on the code

src/main/java/soot/FastHierarchy.java
MarcMil: Correct, that was a bug.

PR #2240, opened 2026-01-16
Fix several Soot bugs related to importing apps

MarcMil:


src/main/java/soot/dexpler/DexNullArrayRefTransformer.java
```diff
@@ -142,12 +142,13 @@ protected void internalTransform(final Body body, String phaseName, Map<String,
    * @return True if the given local is guaranteed to always be null at the given statement, otherwise false
    */
   private boolean isAlwaysNullBefore(Stmt s, Local base, LocalDefs defs) {
-    List<Unit> baseDefs = defs.getDefsOfAt(base, s);
-    if (baseDefs.isEmpty()) {
+    Iterator<Unit> baseDefs = defs.getDefsOfAtIterator(base, s);
+    if (!baseDefs.hasNext()) {
```
StevenArzt, line 146: This check doesn't seem necessary here because the next line will check anyway and the loop won't run

src/main/java/soot/jimple/toolkits/base/ArrayWriteAggregator.java
```diff
@@ -125,6 +126,12 @@ public class ArrayWriteAggregator extends BodyTransformer {
    */
   private static final int THRESHOLD_MEMORY = 100000;
 
+  /**
+   * Maximum method size for which to attempt this analysis. The problem is that a large method size requires a lot of memory
+   * due to the MHGDominatorsFinder analysis.
+   */
+  private static final int MAX_METHOD_SIZE = 100000;
```
StevenArzt, line 133: Large methods are exactly what we want to optimize here to avoid them causing trouble in subsequent analyses. We want to get their size down.

src/main/java/soot/toolkits/scalar/SharedInitializationLocalSplitter.java
```diff
@@ -227,27 +242,26 @@ protected Set<Cluster> newSet() {
           Local luse = (Local) v;
           Iterator<Unit> allAffectingDefs = defs.getDefsOfAtIterator(luse, s);
 
-          MinMaxBitSet constantDefs = new MinMaxBitSet(idx);
-          MinMaxBitSet nonConstantDefs = null;
+          Set<AssignStmt> constantDefs = new HashSet<>();
```
StevenArzt, line 245: I'm not sure that a set is actually more efficient here than a (compacted) BitSet. I'd rather check if we can limit the number of bits we need by compressing the universe.

StevenArzt (changes requested):

PR #2255, opened 2026-02-09
Fix shimple issues with traps

### LINKED ISSUE #1678: Shimple uninitialized registers

MarcMil:
Contains and fixes all test cases in https://github.com/soot-oss/soot/pull/1680
Also incorporates the test cases in https://github.com/soot-oss/soot/pull/1662
Fixes https://github.com/soot-oss/soot/issues/1678


src/main/java/soot/shimple/internal/PhiNodeManager.java
```diff
@@ -361,83 +377,111 @@ public boolean dominates(Unit champ, Unit challenger) {
...
+          AssignStmt stmt = jimp.newAssignStmt(lhsLocal, p.getValue());
+          insertedStatements.add(stmt);
+          Value l = p.getValue();
+          if (l instanceof Local) {
+            List<Unit> ld = localDefs.getDefsOfAt((Local) l, pred);
+            assert ld.size() == 1;
+            Unit elem = ld.get(0);
+            //When the definition site happens to be in a loop, we cannot
+            //move the definition back, since otherwise we might destory
+            //the reference from thje previous iteration
```
StevenArzt: typo: ``thje``

src/main/java/soot/shimple/internal/PhiNodeManager.java
```diff
@@ -361,83 +377,111 @@ public boolean dominates(Unit champ, Unit challenger) {
...
+          }
+          AssignStmt stmt = jimp.newAssignStmt(lhsLocal, p.getValue());
+          insertedStatements.add(stmt);
+          Value l = p.getValue();
+          if (l instanceof Local) {
+            List<Unit> ld = localDefs.getDefsOfAt((Local) l, pred);
+            assert ld.size() == 1;
+            Unit elem = ld.get(0);
+            //When the definition site happens to be in a loop, we cannot
+            //move the definition back, since otherwise we might destory
```
StevenArzt: typo: ``destory``

src/main/java/soot/shimple/internal/PhiNodeManager.java
```diff
@@ -361,83 +377,111 @@ public boolean dominates(Unit champ, Unit challenger) {
...
-    if (equivStmts.size() != predBoxes.size()) {
-      throw new RuntimeException("Assertion failed.");
-    }
+          // if we need to insert the copy statement *before* an
+          // instruction that happens to be *using* the Local being
+          // defined, we need to do some extra work to make sure we
+          // don't overwrite the old value of the local
+          if (pred.branches()) {
+            boolean needPriming = false;
+            Local savedLocal = jimp.newLocal(lhsLocal.getName() + "_", lhsLocal.getType());
```
StevenArzt: It would be better to check for naming duplicates, e.g., using ``LocalGenerator``.

src/main/java/soot/shimple/internal/PhiNodeManager.java
```diff
@@ -361,83 +377,111 @@ public boolean dominates(Unit champ, Unit challenger) {
...
 
-        if (needPriming) {
-          body.getLocals().add(savedLocal);
-          units.insertBefore(jimp.newAssignStmt(savedLocal, lhsLocal), pred);
+            // this is all we really wanted to do!
+            units.insertBefore(stmt, pred);
+          } else {
+            Unit succ = pred;
+            while (true) {
+              Unit s = units.getSuccOf(succ);
```
StevenArzt, line 449: You should check whether there is no successor anymore to avoid infinite loops on malformed Jimple bodies.

StevenArzt (approved):

src/main/java/soot/shimple/internal/PhiNodeManager.java
MarcMil: How is this supposed to work? We are going along the unit chain, at the end there is a ```null``` and we break out.

StevenArzt: OK, due to the ``instanceof`` failing in that case.

MarcMil: Correct

PR #2262, opened 2026-02-11
Cache field ref results

MarcMil:
Using tim-hoffman's idea, partial code and test cases from https://github.com/soot-oss/soot/pull/1519

### DIFF
--- a/src/main/java/soot/AbstractSootFieldRef.java
+++ b/src/main/java/soot/AbstractSootFieldRef.java
@@ -38,11 +38,13 @@ import soot.options.Options;
 public class AbstractSootFieldRef implements SootFieldRef {
   private static final Logger logger = LoggerFactory.getLogger(AbstractSootFieldRef.class);
 
+  private final SootClass declaringClass;
+  private final String name;
+  private final Type type;
+  private final boolean isStatic;
+  private SootField resolveCache = null;
+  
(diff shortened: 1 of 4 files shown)


src/main/java/soot/AbstractSootFieldRef.java
```diff
@@ -106,7 +107,13 @@ public String toString() {
 
   @Override
   public SootField resolve() {
-    return resolve(null);
+    SootField cached = this.resolveCache;
+    // Use the cached SootField if available and still valid
+    if (cached == null || !cached.isValidResolve(this)) {
```
StevenArzt, line 112: Why should we try to re-resolve if the previous resolve failed?

StevenArzt (changes requested):

src/main/java/soot/AbstractSootFieldRef.java
MarcMil: Who says it failed? Possibly the code has been altered in the meantime and the field was renamed or removed. In this case, the resolve isn't valid anymore.
Even if it has failed, we should try again, because again, the code might have changed in the meantime and it might now be able to resolve.

StevenArzt: We can try, but we should not be stuck with cases where we try to resolve the same broken stuff again and again.

PR #2272, opened 2026-03-13
Utilizes the fixed version of AsmMethodSource version from SootUp

MarcMil:
Also fixes a concurrency bug in SootResolver and bugs in the SharedInitializationLocalSplitter


src/main/java/soot/validation/CheckInitValidator.java
```diff
@@ -48,8 +48,8 @@ public static CheckInitValidator v() {
 
   @Override
   public void validate(Body body, List<ValidationException> exception) {
-    ExceptionalUnitGraph g
-        = ExceptionalUnitGraphFactory.createExceptionalUnitGraph(body, ThrowAnalysisFactory.checkInitThrowAnalysis(), false);
+    ThrowAnalysis throwAnalysis = PedanticThrowAnalysis.v();
```
StevenArzt: We need a comment here why we can't use the ``ThrowAnalysis`` that was explicitly written  for this validator (``checkInitThrowAnalysis``).

StevenArzt (changes requested):

src/main/java/soot/validation/CheckInitValidator.java
MarcMil: I've reverted that change.

MarcMil:
Investigating a bug with the new implementation and FlowDroid

PR #2276, opened 2026-04-30
respect `jb.dae` flag in AsmMethodSource

ursachec:
currently the DeadAssignmentEliminator runs unconditionally in AsmMethodSource. This is undesirable in certain cases, for example when static analysis tools using soot do not want this optimization (as is the case for a tool I've contributed to recently). The option is already there, so might as well respect it.


ursachec:
P.S. test suite passes locally

MarcMil:
Looks good to me.
@StevenArzt Can we merge this?

PR #2280, opened 2026-05-21
Fix AbstractHost.remove

### LINKED ISSUE #2279: Broken code in AbstractHost (develop branch)
It seems that the AbstractHost class got broken in commit
6216f5c75201e8cc514d9de86f79766eb89cadf4 ("Allow concurrent handling of tags").
For instance, calling the public AbstractHost.removeTag(...) method raises the
following exception:

```
java.lang.RuntimeException: Unsupported; this is a bad idea concurrently
    at soot.util.ConcurrentList.remove(ConcurrentList.java:91)
    at soot.tagkit.AbstractHost.removeTag(AbstractHost.java:64)
    ...
```

We could probably "fix" element removal by using mTagList.iterator()
(instead of using loops that depend on mTagList.size()).

But to be honest, I do not really understand the commit at all: for instance,
if two different threads call AbstractHost.addTag(...) on the same instance,
both threads could execute the

```
    if (mTagList == null) {
      mTagList = new ConcurrentList<Tag>();
```

codepath. Is this the intended behavior?

MarcMil:
Fixes https://github.com/soot-oss/soot/issues/2279

### DIFF
--- a/src/main/java/soot/tagkit/AbstractHost.java
+++ b/src/main/java/soot/tagkit/AbstractHost.java
@@ -43,7 +43,7 @@ public class AbstractHost implements Host {
 
   // avoid creating an empty list for each element, when it is not used
   // use lazy instantiation (in addTag) instead
-  protected ConcurrentList<Tag> mTagList = null;
+  protected volatile ConcurrentList<Tag> mTagList = null;
 
   /**
    * Get the {@link List} of {@link Tag Tags} on {@code this} {@link Host}. This list should not be modified!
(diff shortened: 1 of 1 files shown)


src/main/java/soot/tagkit/AbstractHost.java
```diff
@@ -36,11 +37,13 @@
...
 public class AbstractHost implements Host {
 
+  private final Object lock = new Object();
+
   protected int line;
 
   // avoid creating an empty list for each element, when it is not used
   // use lazy instantiation (in addTag) instead
-  protected List<Tag> mTagList = null;
+  protected ConcurrentList<Tag> mTagList = null;
```
StevenArzt: This should be ``volatile``

StevenArzt (approved):

PR #2290, opened 2026-07-10
Replace hardcoded concurrency helpers with VirtualCallSiteFilter (#1604)

### LINKED ISSUE #1064: Add extension point to OnFlyCallGraphBuilder for handling special cases

Danish-git123:
Add extension point for virtual call site filters (concurrency helpers)

Introduces VirtualCallSiteFilter interface and AssignableTypeFilter default
implementation so that Thread/Executor/AsyncTask/Handler special-casing in
OnFlyCallGraphBuilder.skipSite() is no longer hardcoded. Callers can now
register custom filters via registerVirtualCallSiteFilter(Kind, filter)
without modifying OnFlyCallGraphBuilder itself.

Fixes #1064

### DIFF
--- a/src/main/java/soot/jimple/toolkits/callgraph/AssignableTypeFilter.java
+++ b/src/main/java/soot/jimple/toolkits/callgraph/AssignableTypeFilter.java
@@ -1,0 +1,31 @@
+package soot.jimple.toolkits.callgraph;
+
+import soot.FastHierarchy;
+import soot.RefType;
+import soot.Type;
+
+/**
+ * A {@link VirtualCallSiteFilter} that only accepts reaching types assignable
+ * to a given base type. Reused for every concurrency helper (Thread, Executor,
+ * AsyncTask, Handler) since they all share the same "must be assignable to X"
+ * rule -- only the base type X differs.
+ */
+public class AssignableTypeFilter implements VirtualCallSiteFilter {
+
+    private final RefType requiredType;
+
+    public AssignableTypeFilter(RefType requiredType) {
+        if (requiredType == null) {
+            throw new IllegalArgumentException("requiredType must not be null");
+        }
+        this.requiredType = requiredType;
+    }
+
+    @Override
+    public boolean skipSite(VirtualCallSite site, FastHierarchy fh, Type type) {
+        return !fh.canStoreType(type, requiredType);
+    }
+}
+
+//A reusable class .Here we can reuse this class for THREAD,EXECUTOR,ASYNCTASK,HANDLER
+//Following the DRY(Dont Repeat yourself) principle
--- a/src/main/java/soot/jimple/toolkits/callgraph/AssignableTypeFilter.java
+++ b/src/main/java/soot/jimple/toolkits/callgraph/AssignableTypeFilter.java
@@ -1,5 +1,27 @@
 package soot.jimple.toolkits.callgraph;
 
+/*-
+ * #%L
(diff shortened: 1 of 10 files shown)


Danish-git123:
Steven Arzt pointed out that virtualedges.xml already declares the Kind
(THREAD/EXECUTOR/HANDLER/ASYNCTASK) for each summary edge. This commit
extends the schema with an optional requiredType attribute so the type
constraint enforced in OnFlyCallGraphBuilder.skipSite() (previously
hardcoded as clRunnable/clAsyncTask/clHandler fields and an if/else chain
on Kind) is now fully data-driven, consistent with the rest of the
virtual edge summary mechanism.

Also fixes a latent inconsistency where the Thread.start() edge had no
type attribute and was silently falling back to Kind.GENERIC_FAKE instead
of Kind.THREAD.

StevenArzt:
Thank you for the contribution. Content-wise, it looks good. Can you please have a look at the style check issues?

Danish-git123:
> Thank you for the contribution. Content-wise, it looks good. Can you please have a look at the style check issues?

Yeah sure i will look at the issue and try to resolve it any further instructions from your side ..?

StevenArzt:
> > Thank you for the contribution. Content-wise, it looks good. Can you please have a look at the style check issues?
> 
> Yeah sure i will look at the issue and try to resolve it any further instructions from your side ..?

The rest is fine, we only need to get the style and license issues fixed. Please have a look at the respective parts of the test build output.

StevenArzt:
The checkstyle build still fails. If you're using Eclipse, you can configue the project with the style files in the repository that are consistent with what checkstyle expects.

I know that the style check can be a pain, but we need to keep the project somewhat consistent across contributors.

Danish-git123:
> The checkstyle build still fails. If you're using Eclipse, you can configue the project with the style files in the repository that are consistent with what checkstyle expects.
> 
> I know that the style check can be a pain, but we need to keep the project somewhat consistent across contributors.

i am using intellij and 
Screenshot 2026-07-20 165329
 even there are 2 indentation it still giving me 4 indentation problem can u help me out with this

Danish-git123:
the license checks are passed remaining is style checks also there any further issue for the same ??

Danish-git123:
All checks are passing now (style, license, and build on Java 11/17/21). 
Ready for another look whenever you get a chance!

StevenArzt:
Our test build is fine, so I can merge. Thank you again for your contribution!

PR #2293, opened 2026-07-22
ASM Backend: Fix too aggressive trap tightening

MarcMil:


src/main/java/soot/toolkits/exceptions/ThrowableSet.java
```diff
@@ -798,7 +798,12 @@ public Pair whichCatchableAs(RefType catcher) {
           // will already have been copied to caughtExcluded by
           // the preceding loop.
           caughtIncluded = addExceptionToSet(inclusion, caughtIncluded);
-        } else if (h.canStoreType(catcher, base)) {
+        } else if (h.canStoreType(catcher, base) ||
+
+        // when we don't have specific information we just assume that it is included.
+            (catcher.getSootClass().getSuperclassUnsafe() == null
+                || catcher.getSootClass().getSuperclass().getType().equals(scene.getObjectType()))
```
StevenArzt, line 805: What about ``Serializable`` and ``Cloneable``? They are all three handled specially in the Java bytecode specification.

src/main/java/soot/asm/AsmMethodSource.java
```diff
@@ -988,7 +989,6 @@ private void convertJumpInsn(JumpInsnNode insn) {
       if (!insnToStmt.containsKey(insn)) {
         UnitBox box = Jimple.v().newStmtBox(null);
         GotoStmt gotoStmt = Jimple.v().newGotoStmt(box);
-        stmtsThatBranchToLabel.put(gotoStmt, insn.label);
```
StevenArzt, line 991: What remains as the use case for ``stmtsThatBranchToLabel`` with all these additions removed?

StevenArzt (changes requested):

src/main/java/soot/toolkits/exceptions/ThrowableSet.java
MarcMil: This handles cases where the catcher is a phantom class *and* no hierarchy info is available; in these cases, java.lang.Object is the direct superclass (because it's the default used by Soot) and it's a phantom class.

PR #2295, opened 2026-08-17
Changed invoke type from instance to static

Danish-git123:


Danish-git123:
if there is any advancement let me know..

StevenArzt (approved):

PR #2296, opened 2026-08-31
Add lambda class naming strategies for the LambdaMetaFactory

MarcMil:
Furthermore, the option to keep the bytecode offset now works for ASM


StevenArzt:
@MarcMil  The build failed, please double-check.

StevenArzt (approved):

PR #2303, opened 2026-09-03
Fix incorrect outer class resolution for classes with dollar sign in name

### LINKED ISSUE #1956: Exception when processing classes with "$" that are not nested
Processing classes that contain a dollar sign in their name but are not nested throws an exception.

**To reproduce**
`driver/Driver.java`
```java
package driver;

import soot.Scene;
import soot.options.Options;
import java.util.Collections;

public class Main {
    public static void main(String[] args) {
        Options.v().set_process_dir(Collections.singletonList("target/classes/sample"));
    }
}
```

`sample/A$B.java`
```java
package sample;

public class A$B { }
```

**Expected behavior**
Class `A$B` should be successfully processed since `$` is a valid character in an identifier. This really happens in practice, e.g., Gson contains `/com/google/gson/internal/$Gson$Types.class`.

**Stacktrace**
```
Exception in thread "main" soot.SootResolver$SootClassNotFoundException: couldn't find class: A (is your soot-class-path set properly?)
	at soot.SootResolver.bringToHierarchyUnchecked(SootResolver.java:267)
	at soot.SootResolver.bringToHierarchy(SootResolver.java:243)
	at soot.SootResolver.bringToSignatures(SootResolver.java:313)
	at soot.SootResolver.processResolveWorklist(SootResolver.java:198)
	at soot.SootResolver.resolveClass(SootResolver.java:155)
	at soot.Scene.loadClass(Scene.java:1016)
	at soot.Scene.loadClassAndSupport(Scene.java:1003)
	at soot.Scene.loadNecessaryClasses(Scene.java:1968)
	at driver.Driver.main(Driver.java:10)
```

The dollar sign can be also at the beginning of a class name. If we rename the sample class to `$A`, the stack trace is:
```
Exception in thread "main" java.lang.RuntimeException: Classname must not be empty!
	at soot.SootResolver.makeClassRef(SootResolver.java:128)
	at soot.asm.AsmClassSource.resolve(AsmClassSource.java:86)
	at soot.SootResolver.bringToHierarchyUnchecked(SootResolver.java:274)
	at soot.SootResolver.bringToHierarchy(SootResolver.java:243)
	at soot.SootResolver.bringToSignatures(SootResolver.java:313)
	at soot.SootResolver.processResolveWorklist(SootResolver.java:198)
	at soot.SootResolver.resolveClass(SootResolver.java:155)
	at soot.Scene.loadClass(Scene.java:1016)
	at soot.Scene.loadClassAndSupport(Scene.java:1003)
	at soot.Scene.loadNecessaryClasses(Scene.java:1968)
	at driver.Driver.main(Driver.java:10)
```

**Additional context**
Guessing the outer class name using the nested class name is not necessary as the dollar sign is just a convention. The `InnerClasses` attribute in class files contains the `outer_class_info_index` item for this purpose, as described in the [JVM specification](https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-4.html#jvms-4.7.6).

mmustafasenoglu:
## Summary

Fixes #1956

When a class name contains `$`, Soot previously assumed the part before the last `$` was the outer class. This fails for non-nested classes that merely contain `$` in their names (e.g., `$Gson$Types`, `A$B`), causing a `SootClassNotFoundException` because the inferred outer class does not exist.

## Root Cause

In `AsmClassSource.resolve()` (line 79-92), when `!sc.hasOuterClass() && className.contains("$")`, Soot blindly infers the outer class by splitting on the last `$`:

```java
outerClassName = className.substring(0, className.lastIndexOf('$'));
```

This heuristic is incorrect for non-nested classes like `sample.$Gson$Types` — it would infer the outer class is `sample.$Gson`, which doesn't exist.

## Fix

The fix checks the `InnerClassTag` metadata (populated by `SootClassBuilder.visitInnerClass`) to determine whether the class is actually listed as an inner class with a known outer class in the `InnerClasses` attribute:

1. Iterate over the class's tags looking for an `InnerClassTag` where `innerClass` matches the current class name and `outerClass` is non-null
2. If found, use the `outerClass` from the tag (authoritative source)
3. Only fall back to the `$` heuristic if no matching `InnerClassTag` exists

**Key insight**: For a non-nested class with `$` in its name (e.g., `sample.A$B`), the class file's `InnerClasses` attribute will NOT contain an entry where `inner == sample/A$B` AND `outer != null`. So no matching `InnerClassTag` is found, and the class correctly gets no outer class set.

## Changes

- **`AsmClassSource.java`**: Added `InnerClassTag` import; replaced direct `$` heuristic with `InnerClassTag`-aware logic that first checks explicit metadata before falling back to the heuristic
- **`DollarSignClassNameTest.java`** (new): 4 tests using ASM-generated classes:
  - Genuine nested class (`Outer$Inner`) correctly resolves outer class
  - Non-nested `sample.A$B` does NOT get a false outer class
  - Non-nested `sample.$Gson$Types` does NOT get a false outer class  
  - Both non-nested classes load without `SootClassNotFoundException`

## Testing

All 4 new tests pass. Full test suite: 350 tests run, 0 new failures (1 pre-existing `PackManagerTest.testWritingToJar` DexPrinter error). Checkstyle passes.

### DIFF
--- a/src/main/java/soot/asm/AsmClassSource.java
+++ b/src/main/java/soot/asm/AsmClassSource.java
@@ -37,6 +37,7 @@ import soot.IFoundFile;
 import soot.SootClass;
 import soot.SootResolver;
 import soot.UserInputException;
+import soot.tagkit.InnerClassTag;
 
 /**
  * ASM class source implementation.
@@ -77,17 +78,35 @@ public class AsmClassSource extends ClassSource {
       // add the outer class information, could not be called in the builder, since sc needs to be
       // resolved - before calling setOuterClass()
       if (!sc.hasOuterClass() && className.contains("$")) {
-        String outerClassName;
-        if (className.contains("$-")) {
-          /*
-           * This is a special case for generated lambda classes of jack and jill compiler. Generated lambda classes may
-           * contain '$' which do not indicate an inner/outer class separator if the '$' occurs after a inner class with a
-           * name starting with '-'. Thus we search for '$-' and anything after it including '-' is the inner classes name
-           * and anything before it is the outer classes name.
-           */
-          outerClassName = className.substring(0, className.indexOf("$-"));
-        } else {
-          outerClassName = className.substring(0, className.lastIndexOf('$'));
+        String outerClassName = null;
+        // First, check if there is an InnerClassTag that explicitly declares this class as an
+        // inner class with a known outer class. This distinguishes genuine nested classes from
+        // non-nested classes that merely contain '$' in their name (e.g., "$Gson$Types").
+        for (soot.tagkit.Tag tag : sc.getTags()) {
+          if (tag instanceof InnerClassTag) {
+            InnerClassTag ict = (InnerClassTag) tag;
+            String innerName = ict.getInnerClass();
+            if (innerName != null && !innerName.contains("/") && AsmUtil.toQualifiedName(innerName).equals(className)
+                && ict.getOuterClass() != null) {
+              outerClassName = AsmUtil.toQualifiedName(ict.getOuterClass());
+              break;
+            }
+          }
+        }
+        // If no matching InnerClassTag with a non-null outer class was found, fall back to the
+        // dollar-sign heuristic for legacy support.
+        if (outerClassName == null) {
+          if (className.contains("$-")) {
+            /*
+             * This is a special case for generated lambda classes of jack and jill compiler. Generated lambda classes may
+             * contain '$' which do not indicate an inner/outer class separator if the '$' occurs after a inner class with a
+             * name starting with '-'. Thus we search for '$-' and anything after it including '-' is the inner classes name
+             * and anything before it is the outer classes name.
+             */
+            outerClassName = className.substring(0, className.indexOf("$-"));
+          } else {
+            outerClassName = className.substring(0, className.lastIndexOf('$'));
+          }
         }
         sc.setOuterClass(SootResolver.v().makeClassRef(outerClassName));
       }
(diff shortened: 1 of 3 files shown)


src/main/java/soot/asm/AsmClassSource.java
```diff
@@ -77,17 +78,35 @@ public Dependencies resolve(SootClass sc) {
...
-           * and anything before it is the outer classes name.
-           */
-          outerClassName = className.substring(0, className.indexOf("$-"));
-        } else {
-          outerClassName = className.substring(0, className.lastIndexOf('$'));
+        String outerClassName = null;
+        // First, check if there is an InnerClassTag that explicitly declares this class as an
+        // inner class with a known outer class. This distinguishes genuine nested classes from
+        // non-nested classes that merely contain '$' in their name (e.g., "$Gson$Types").
+        for (soot.tagkit.Tag tag : sc.getTags()) {
```
StevenArzt: This should probably be ``SootClass.isInnerClass()`` with a suitable implementation there. Otherwise, we fix the issue here, but keep the old flawed logic elsewhere.

StevenArzt (changes requested):

mmustafasenoglu:
Thanks for the review, @StevenArzt! I've moved the InnerClassTag check into `SootClass.isInnerClass()` as suggested.

**Changes made:**

1. **`SootClass.isInnerClass()`** — Now checks InnerClassTags in addition to `hasOuterClass()`. It iterates the class's tags, looks for an `InnerClassTag` where the inner class name matches this class and the outer class is non-null. This correctly identifies inner classes even when the outer class hasn't been resolved yet, and avoids the flawed dollar-sign heuristic.

2. **`AsmClassSource.resolve()`** — Now uses `sc.isInnerClass()` to gate the InnerClassTag lookup. Also fixed a bug where `!innerName.contains("/")` incorrectly filtered out inner classes in packages (since `InnerClassTag.getInnerClass()` returns internal names with `/` for classes in packages like `java/util/Map$Entry`).

This way, other parts of the codebase (e.g., `DexAnnotation`, `DexInnerClassParser`) can also use `SootClass.isInnerClass()` instead of their own dollar-sign heuristics.

StevenArzt (approved):

PR #2310, opened 2026-09-15
 Allow to save the output bytecode offset

MarcMil:


MarcMil:
Work in progress for now

src/main/java/soot/toDex/MultiDexBuilder.java
```diff
@@ -164,13 +175,57 @@ private void fixJumboStringInstructions() {
           continue;
         }
         List<BuilderInstruction> insns = t.getInstructions();
+        Map<Integer, Unit> previousBytecodeOffsetToStmt = null;
+        if (addBytecodeOffsets) {
+          SootClass sc
+              = Scene.v().getSootClassUnsafe(clz.getType().substring(1, clz.getType().length() - 1).replace('/', '.'));
```
StevenArzt, line 181: There is a helper function for that somewhere. It is Java to Jimple name translation.

StevenArzt (changes requested):

